PR #1, opened 2018-03-09
Redis sentinel role support

eguzki:
Enable sentinels role configuration for stats and resque connections.

Using 3scale_backend.conf config file, role configuration can be read from env vars:
 - ```CONFIG_QUEUES_SENTINEL_ROLE``` for queue connection
 - ```CONFIG_REDIS_SENTINEL_ROLE``` for stats storage connection

### DIFF
--- a/lib/3scale/backend/storage.rb
+++ b/lib/3scale/backend/storage.rb
@@ -46,7 +46,7 @@ module ThreeScale
           private_constant :CONN_WHITELIST
 
           # Parameters regarding target server we will take from a config object
-          URL_WHITELIST = [:url, :proxy, :server, :sentinels].freeze
+          URL_WHITELIST = [:url, :proxy, :server, :sentinels, :role].freeze
           private_constant :URL_WHITELIST
 
           # these are the parameters we will take from a config object
@@ -182,6 +182,9 @@ module ThreeScale
           #   { host: "hostN", port: "portN" }
           # ]
           def cfg_sentinels_handler(options)
+            # get role attr and remove from options
+            # will only be validated and included when sentinels are valid
+            role = options.delete :role
             sentinels = options.delete :sentinels
             # The Redis client can't accept empty string or array of :sentinels
             return options if sentinels.nil? || sentinels.empty?
@@ -205,6 +208,8 @@ module ThreeScale
               end
             end.compact
 
+            # Handle role option when sentinels are validated
+            options[:role] = role if role && !role.empty?
             options
           end
 
--- a/openshift/3scale_backend.conf
+++ b/openshift/3scale_backend.conf
@@ -19,11 +19,13 @@ ThreeScale::Backend.configure do |config|
   config.internal_api.password = "#{ENV['CONFIG_INTERNAL_API_PASSWORD']}"
   config.queues.master_name = "#{ENV['CONFIG_QUEUES_MASTER_NAME']}"
   config.queues.sentinels = "#{ENV['CONFIG_QUEUES_SENTINEL_HOSTS'] && !ENV['CONFIG_QUEUES_SENTINEL_HOSTS'].empty? ? ENV['CONFIG_QUEUES_SENTINEL_HOSTS'] : ENV['SENTINEL_HOSTS']}"
+  config.queues.role = "#{ENV['CONFIG_QUEUES_SENTINEL_ROLE']}".to_sym
   config.queues.connect_timeout = parse_int_env('CONFIG_QUEUES_CONNECT_TIMEOUT')
   config.queues.read_timeout = parse_int_env('CONFIG_QUEUES_READ_TIMEOUT')
   config.queues.write_timeout = parse_int_env('CONFIG_QUEUES_WRITE_TIMEOUT')
   config.redis.proxy = "#{ENV['CONFIG_REDIS_PROXY']}"
   config.redis.sentinels = "#{ENV['CONFIG_REDIS_SENTINEL_HOSTS']}"
+  config.redis.role = "#{ENV['CONFIG_REDIS_SENTINEL_ROLE']}".to_sym
   config.redis.connect_timeout = parse_int_env('CONFIG_REDIS_CONNECT_TIMEOUT')
   config.redis.read_timeout = parse_int_env('CONFIG_REDIS_READ_TIMEOUT')
   config.redis.write_timeout = parse_int_env('CONFIG_REDIS_WRITE_TIMEOUT')
--- a/test/unit/storage_test.rb
+++ b/test/unit/storage_test.rb
@@ -42,15 +42,111 @@ class StorageTest < Test::Unit::TestCase
     end
   end
 
+  def test_sentinels_connection_string
+    config_obj = {
+      url: 'redis://127.0.0.1:6379/0',
+      sentinels: ',redis://127.0.0.1:26379, ,    , 127.0.0.1:36379,'
+    }
+    conn = Storage.send :orig_new, Storage::Helpers.config_with(config_obj)
+    assert_sentinel_connector(conn.client)
+    assert_client_config(conn.client, url: config_obj[:url],
+                                      sentinels: [{ host: '127.0.0.1', port: 26_379 },
+                                                  { host: '127.0.0.1', port: 36_379 }])
+  end
+
+  def test_sentinels_connection_string_escaped
+    config_obj = {
+      url: 'redis://127.0.0.1:6379/0',
+      sentinels: 'redis://user:passw\,ord@127.0.0.1:26379 ,127.0.0.1:36379, ,'
+    }
+    conn = Storage.send :orig_new, Storage::Helpers.config_with(config_obj)
+    assert_sentinel_connector(conn.client)
+    assert_client_config(conn.client, url: config_obj[:url],
+                                      sentinels: [{ host: '127.0.0.1', port: 26_379 },
+                                                  { host: '127.0.0.1', port: 36_379 }])
+  end
+
+  def test_sentinels_connection_array_strings
+    config_obj = {
+      url: 'redis://127.0.0.1:6379/0',
+      sentinels: ['redis://127.0.0.1:26379 ', ' 127.0.0.1:36379', nil]
+    }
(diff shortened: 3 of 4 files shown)


lib/3scale/backend/storage.rb
```diff
@@ -205,6 +208,8 @@ def cfg_sentinels_handler(options)
               end
             end.compact
 
+            # Handle role option when sentinels are validated
+            options[:role] = role if role && !role.empty?
```
unleashed: I'd ensure here we do the `to_sym`. It also avoids adding the `to_sym` to the template.

unleashed: A question that I now think about is: what if `sentinels` is parsed and returns an empty array (ie. the input is a string made of spaces or commas), would adding the `role` parameter here break the client setup?

openshift/3scale_backend.conf
```diff
@@ -19,11 +19,13 @@ ThreeScale::Backend.configure do |config|
   config.internal_api.password = "#{ENV['CONFIG_INTERNAL_API_PASSWORD']}"
   config.queues.master_name = "#{ENV['CONFIG_QUEUES_MASTER_NAME']}"
   config.queues.sentinels = "#{ENV['CONFIG_QUEUES_SENTINEL_HOSTS'] && !ENV['CONFIG_QUEUES_SENTINEL_HOSTS'].empty? ? ENV['CONFIG_QUEUES_SENTINEL_HOSTS'] : ENV['SENTINEL_HOSTS']}"
+  config.queues.role = "#{ENV['CONFIG_QUEUES_SENTINEL_ROLE']}".to_sym
```
unleashed: maybe just assume we will get a `String` or `Symbol` here and apply `to_sym` in the code instead of here?

test/unit/storage_test.rb
```diff
@@ -114,14 +102,73 @@ def test_sentinels_malformed_url
...
+                                        sentinels: [{ host: '127.0.0.1', port: 26_379 }],
+                                        role: role)
+    end
+  end
+
+  def test_sentinels_role_empty
+    [''.to_sym, '', nil].each do |role|
+      config_obj = {
+        url: 'redis://127.0.0.1:6379/0',
+        sentinels: 'redis://127.0.0.1:26379',
```
unleashed: perhaps also add a test where `sentinels` is `"    "` and/or `",,,," and combined with `role` having a valid value, like the comment I do elsewhere wondering about what would happen? (ie. this client breaks in the weird case of specifying sentinels as empty, so I fear this could also be problematic).

eguzki: no problem with that. When ```sentinels``` is ```" "``` or ```,,,```, then redis connection will be called with empty array.
```
con = Redis.new(url: 'redis://mymaster', sentinels: [], role: :master)
```
No matter what is ```role``` value, error will be raised reporting error about sentinels configuration:
```
con = Redis.new(url: 'redis://mymaster', sentinels: [], role: :master)
=> #<Redis client v3.3.3 for redis://mymaster:6379/0>
irb(main):003:0> con.get('3')
Redis::CannotConnectError: No sentinels available.
```

If no sentinels is configured, ```role``` will not be read and will work as usual.

lib/3scale/backend/storage.rb
eguzki: Regarding removing ```to_sym``` from configuration file, I aggree. But to be consistent, all other conversions should be removed as well: *to_i* and other *to_sym*. I guess this task belongs to other PR, it is out of scope.

test/unit/storage_test.rb
```diff
@@ -42,6 +42,17 @@ def test_redis_url_without_scheme
     end
   end
 
+  def test_sentinels_simple_url
+    config_obj = {
+      url: 'master-name', # url of the sentinel master name conf
+      sentinels: 'redis://127.0.0.1:26379'
+    }
+    conn = Storage.send :orig_new, Storage::Helpers.config_with(config_obj)
+    test_sentinel_connector(conn.client)
```
davidor: Where is this method defined?

test/unit/storage_test.rb
```diff
@@ -149,12 +185,13 @@ def assert_sentinel_connector(client)
     assert_instance_of Redis::Client::Connector::Sentinel, connector
   end
 
-  def assert_client_config(client, host: nil, port: nil, url: nil, sentinels: nil)
+  def assert_client_config(client, host: nil, port: nil, url: nil, sentinels: nil, role: nil)
```
davidor: I think we can improve the function name: `assert_client_config_is_correct` or similar. Right now it reads like an incomplete sentence.

Also, have you considered passing `attrs_to_check = {}` ? All the params have `nil` as the default value, so it would be equivalent and cleaner, in my opinion.

lib/3scale/backend/storage.rb
```diff
@@ -182,6 +182,9 @@ def ensure_url_param(options)
           #   { host: "hostN", port: "portN" }
           # ]
           def cfg_sentinels_handler(options)
+            # get role attr and remove from options
+            # will only be validated and included when sentinels are valid
+            role = options.delete :role
```
davidor: I think it would be better to move this to the end of the method where `role` used. There are many lines between the assignment of `role` and its usage.

openshift/3scale_backend.conf
davidor: I agree, @unleashed. Otherwise, we'll be forcing all the users who create a config file like this to remember to call `to_sym`.

davidor:
@eguzki The PR looks good :+1:  
I added a few comments about minor things.

openshift/3scale_backend.conf
eguzki: https://github.com/3scale/apisonator/pull/1#discussion_r174224359

davidor: :+1:

test/unit/storage_test.rb
mikz: @eguzki yes, from 5 to 6 you go over rubocop default: http://rubocop.readthedocs.io/en/latest/cops_metrics/#metricsparameterlists

lib/3scale/backend/storage.rb
unleashed: well, it makes sense to not introduce _yet another_ setting doing the wrong thing - if you don't fix it here, can you create a PR to clean this up?

test/unit/storage_test.rb
unleashed: ok :+1:

test/unit/storage_test.rb
unleashed: we can have host and port separated from the rest (because they are used differently) and iterate the rest as a hash - wdyt @eguzki?

eguzki: ok, I am on it

test/unit/storage_test.rb
```diff
@@ -138,12 +185,15 @@ def assert_sentinel_connector(client)
...
-  def assert_client_config(client, host: nil, port: nil, url: nil, sentinels: nil)
+  def assert_client_config(client, host: nil, port: nil, url: nil, **opts)
     raise "bad usage of #{__method__}" unless host || port || url
     assert_equal client.port, port if port
     assert_equal client.host, host if host
     assert_equal client.options[:url], url if url
+    sentinels = opts[:sentinels]
     assert_equal client.options[:sentinels], sentinels if sentinels
+    role = opts[:role]
+    assert_equal client.options[:role], role if role
```
unleashed: how about:
```ruby
opts.each do |o, v|
  assert_equal client.options[o], v
end
```
This would also allow you to pass `url` as part of `**opts`: `def assert_client_config(client, host: nil, port: nil, **opts)`

eguzki: Host and port are no longer used in this test set. So, refactoring makes sense.

eguzki: done

test/unit/storage_test.rb
```diff
@@ -138,12 +185,11 @@ def assert_sentinel_connector(client)
...
-  def assert_client_config(client, host: nil, port: nil, url: nil, sentinels: nil)
-    raise "bad usage of #{__method__}" unless host || port || url
-    assert_equal client.port, port if port
-    assert_equal client.host, host if host
-    assert_equal client.options[:url], url if url
-    assert_equal client.options[:sentinels], sentinels if sentinels
+  def assert_client_config(client, opts)
+    raise "bad usage of #{__method__}" unless opts[:host] || opts[:port] || opts[:url]
+    opts.each do |k, v|
+      assert_equal client.options[k], v if v
```
unleashed: why the `if v`? this is under the caller's control, so they might want to assert a specific options is nil? the original `if v` comes from the keyword arguments being mandatory (so the key always exists), but it's not the case with a hash.

eguzki: agree.

Done

test/unit/storage_test.rb
```diff
@@ -138,12 +185,11 @@ def assert_sentinel_connector(client)
...
   end
 
-  def assert_client_config(client, host: nil, port: nil, url: nil, sentinels: nil)
-    raise "bad usage of #{__method__}" unless host || port || url
-    assert_equal client.port, port if port
-    assert_equal client.host, host if host
-    assert_equal client.options[:url], url if url
-    assert_equal client.options[:sentinels], sentinels if sentinels
+  def assert_client_config(client, opts)
+    raise "bad usage of #{__method__}" unless opts[:host] || opts[:port] || opts[:url]
```
mikz: Not really needed now no? Maybe just assert it is not empty?

eguzki: Definitely *url* has to be checked for not empty.
Others, not needed.

unleashed: so if there are mandatory fields, we should make them mandatory in the method signature:
```ruby
def assert_client_config(client, url, opts)
```

Also, probably opts is not the best name for the hash containing the rest of assertions. That said, are you sure the client connection configuration object always has a valid `url` field?

eguzki: we are testing exactly that and, moreover, you removed *host, port* mode of redis connection configuration.

updated to read (and making sure it exists) *url*

unleashed:
bors r=davidor,unleashed

PR #2, opened 2018-03-13
Environment for development and testing

unleashed:
This adds both `dev` and `test` targets to the root `Makefile` and removes old stuff we had for our previous private testing, simplifying the way we use containers for development and testing.

Passing in the `BUILD_CI` option set to `1` forces building images from scratch instead of pulling them when the required image is missing. Another interesting option is `DEV_TOOLS`, which is a set of packages that you can customize to have in your development image.

Please take a look at all the targets in `Makefile` in addition to the ones mentioned above to make sure everything makes sense.

### Development environment

`make dev` will create a persistent container with the current checked out code mapped inside. You can perform changes both inside and outside the container and they will always be in sync, and you can install extra packages and configure the development container to your tasted because it will always be resumed when you next do `make dev`, unless you explicitly want it removed. Testing can be launched here manually by running `script/test` and `script/test_external`.

### Test environment

Additionally there is a `make test` target which reuses the same `Dockerfile` as the development environment, but just copies the current checked out code, creates a container, runs the test suite and removes the container. This is intended for quick test runs without having a persistent container or any synchronization between host and container.


unleashed:
bors ping

unleashed:
bors try

davidor:
Works for me :+1: 

Does the `Maintain your dependencies up-to-date` section in the README still apply? If so, can we find a way so everytime the dev container is started it ensures that the dependencies are up to date?

davidor:
How about adding a `make help` target like the one in Apicast: https://github.com/3scale/apicast/blob/69ff05c03e934daf1fa3a9b2cfb4231773b5a22a/Makefile#L183 ?

I think it could be useful given the number of targets that we have.

README.md
```diff
@@ -75,7 +73,7 @@ the private container image.
 
 Bear in mind that we can only provide best effort support for this.
 
-1. Install Redis v2.8.19 (or install it elsewhere).
+1. Install Redis v2.8.19 (or whichever version is recommended for use).
```
davidor: So where should we document the recommended version?

unleashed:
bors r=@eguzki,@davidor

davidor:
bors r=@eguzki,@davidor

PR #9, opened 2018-04-04
Add dev-service and dev-service-clean Makefile rules

miguelsorianod:
This rules have been added to be able to start a container with the 3scale_backend service running.

The rule dev-service-clean removes the container.
Also, the dev-clean-image rule has been updated to remove the associated dev-service container too.
Due to now we may have the dev-service and dev container running I have improved the running container detection logic
in the dev-service and dev rules to only match the container name (whole name and not matching prefixes)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,14 @@ default: | clean test show_bench
 test: export BUILD_CI?=0
 test: export DEV_TOOLS?=""
 test: export IMAGE_NAME?=apisonator-test
+test: export CONTAINER_NAME?=apisonator-test
 test: DOCKER_OPTS?=
 test:
 	make -C $(PROJECT_PATH) -f $(MKFILE_PATH) dev-build
-	docker run -ti --rm -h apisonator-test -v \
+	docker run -ti --rm -h $(CONTAINER_NAME) -v \
 		$(PROJECT_PATH):$$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'cd && pwd')/apisonator:z \
 		-u $$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'id -u'):$$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'id -g') \
-	$(DOCKER_OPTS) --name apisonator-test $(IMAGE_NAME)
+	$(DOCKER_OPTS) --name $(CONTAINER_NAME) $(IMAGE_NAME)
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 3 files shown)


miguelsorianod:
Hi,

Try the work flow of this new functionality and let me know what you think about this please.

Makefile
```diff
@@ -29,29 +29,54 @@ show_bench:
...
+.PHONY: dev-service-clean
+dev-service-clean: CONTAINER_NAME=apisonator-dev-service
+dev-service-clean:
+	-docker kill $(CONTAINER_NAME)
+	-docker rm $(CONTAINER_NAME)
 
 .PHONY: dev-clean-image
-dev-clean-image: IMAGE_NAME?=apisonator-dev
-dev-clean-image: dev-clean
+dev-clean-image: IMAGE_NAME=apisonator-dev
```
maneta: are you sure you do want to use `?=` ?

miguelsorianod: Hi,

You are right, there is no need to remove the ? . I did that because in the case of the variable CONTAINER_NAME i wanted it to be reevaluated always due to different rules (dev-clean, dev-service-clean) use that variable with different values.

I added a new commit to update this.

Thank you.

README.md
```diff
@@ -47,6 +47,9 @@ you enter it.
 Getting rid of the persistent container is done with `make dev-clean`, whereas
 removing its image is done using `make dev-clean-image`.
 
+Alternatively you can start a container with the service running with `make dev-service`. This rule is intended for when you want to test
```
unleashed: You might want to start the new phrase in a new line (ie. it does not matter when visualizing the markdown, but when reading the source ASCII text file this line stands out).

Makefile
```diff
@@ -29,29 +29,54 @@ show_bench:
...
-	@if docker ps -a --filter name=apisonator-dev | grep -q $(IMAGE_NAME) 2> /dev/null >&2; then \
-		docker start -ai apisonator-dev ; \
+	@if docker ps -a --filter name=$(CONTAINER_NAME) --format "{{.Names}}" | grep -w -q '^$(CONTAINER_NAME)$$' 2> /dev/null >&2; then \
+		docker start -ai $(CONTAINER_NAME) ; \
 	else \
 		docker run -ti -h apisonator-dev -v \
 		$(PROJECT_PATH):$$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'cd && pwd')/apisonator:z \
 		-u $$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'id -u'):$$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'id -g') \
-		--name apisonator-dev $(IMAGE_NAME) /bin/bash ; \
+		--name $(CONTAINER_NAME) $(IMAGE_NAME) script/test_external ; \
```
unleashed: forgot to set the container's hostname to `$(CONTAINER_NAME)` a couple of lines above.

Makefile
```diff
@@ -29,29 +29,54 @@ show_bench:
...
+	@if docker ps -a --filter name=$(CONTAINER_NAME) --format "{{.Names}}" | grep -q '^$(CONTAINER_NAME)$$' 2> /dev/null >&2; then \
+		docker start -ai $(CONTAINER_NAME) ; \
+	else \
+		docker run -ti -h $(CONTAINER_NAME) -v \
+		$(PROJECT_PATH):$$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'cd && pwd')/apisonator:z \
+		-u $$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'id -u'):$$(docker run --rm $(IMAGE_NAME) /bin/bash -c 'id -g') \
+		--name $(CONTAINER_NAME) $(IMAGE_NAME) /bin/bash ; \
+	fi
+
+.PHONY: dev
```
unleashed: this should be `.PHONY: dev-service`.

Makefile
```diff
@@ -29,29 +29,54 @@ show_bench:
...
+dev-service: export IMAGE_NAME?=apisonator-dev
+dev-service: export CONTAINER_NAME=apisonator-dev-service
+dev-service:
+	@docker history -q $(IMAGE_NAME) 2> /dev/null >&2 || $(MAKE) -C $(PROJECT_PATH) -f $(MKFILE_PATH) dev-build
+	@if docker ps --filter name=$(CONTAINER_NAME) --format "{{.Names}}" | grep -q '^$(CONTAINER_NAME)$$' 2> /dev/null >&2; then \
+		echo "$(CONTAINER_NAME) container already started" >&2; false ; \
 	fi
-	@if docker ps -a --filter name=apisonator-dev | grep -q $(IMAGE_NAME) 2> /dev/null >&2; then \
-		docker start -ai apisonator-dev ; \
+	@if docker ps -a --filter name=$(CONTAINER_NAME) --format "{{.Names}}" | grep -w -q '^$(CONTAINER_NAME)$$' 2> /dev/null >&2; then \
```
unleashed: using `grep -w` here but not above nor in the other rule.

Makefile
miguelsorianod: I did leave the hostname to the hardcoded value because I was not sure if it was some requirement at application level. As we talked the application is not affected when the hostname changes so I will change it.

Makefile
miguelsorianod: It was incorrect yes... I'll fix it

Makefile
miguelsorianod: True. -w is not needed so I'll remove it from here

README.md
miguelsorianod: I'll edit the README.md file so it is more easily readable when not using a Markdown visualizer. I'll set the limit line of the lines I write to 80 characters as I see it seems is being done

unleashed:
bors try

PR #11, opened 2018-04-09
Individualize RSpec rake tasks

miguelsorianod:
I thought it would be a good idea to be able to execute RSpec tests inividually or by type of test (unit, integration, ...) via Rake from within the development environment. I have created specific Rake tasks to execute them.

Let me know if you see any problems. In each commit I have written the motivations of the changes associated to it.

Please tell me if this changes could cause problems in the current integration/testing systems  to see if I may be generating some incompatibility.

Thank you.


Rakefile
```diff
@@ -65,8 +65,48 @@ if saas?
...
+        task.pattern = 'spec/api/**/*_spec.rb'
+      end
+
+      desc 'Run RSpec use_cases tests'
+      RSpec::Core::RakeTask.new(:use_cases) do |task|
+        task.pattern = 'spec/use_cases/**/*_spec.rb'
+      end
+
+      desc 'Run RSpec test/s specified by input file pattern'
+      RSpec::Core::RakeTask.new(:specific, :test_name) do | t, task_args |
```
unleashed: style `|arg1, arg2|` - also, you could use `task` instead of `t` just like in the previous examples

Rakefile
```diff
@@ -102,8 +102,8 @@ if saas?
       end
 
       desc 'Run RSpec test/s specified by input file pattern'
-      RSpec::Core::RakeTask.new(:specific, :test_name) do | t, task_args |
-        t.pattern = "#{task_args[:test_name]}"
+      RSpec::Core::RakeTask.new(:specific, :test_name) do | task, task_args |
```
unleashed: ah, fixed the name here - the style issue remains, and you should squash this commit into the one that introduces these changes.

README.md
```diff
@@ -98,6 +98,37 @@ variables for the test scripts like so:
...
+needed to run them correctly must be started before with:
+
+> `$ script/services start`
+
+Then you can execute the following commands to execute all tests:
+
+> `$ bundle exec rake`
+
+Another alternative to execute all tests from within the development environment
+where it is not needed to manually start the services before is:
```
unleashed: this is not correct: `it is not needed` should be replaced with `you should not` (ie. it might work only out of chance and it is likely to produce problems) - and maybe mention how to stop the services in case you had them running.

README.md
```diff
@@ -98,6 +98,37 @@ variables for the test scripts like so:
...
+> `$ bundle exec rake spec:[unit|integration|acceptance|api|use_cases] SPEC=<test-file>`
+
+Alternatively there is another syntax to execute a RSpec type test:
+> `$ bundle exec rake spec:specific[<test-file-location>]`
+
+where in this last case the [ ] characters must be literally placed.
+
+For more information on accepted commands for testing with Rake you can execute:
+
+> `$bundle exec rake -T`
```
unleashed: typo: place a space in between the prompt and `bundle`.

README.md
```diff
@@ -98,6 +98,37 @@ variables for the test scripts like so:
...
+
+If the test is a RSpec type test (located in spec directory) you can execute:
+> `$ bundle exec rake spec:[unit|integration|acceptance|api|use_cases] SPEC=<test-file>`
+
+Alternatively there is another syntax to execute a RSpec type test:
+> `$ bundle exec rake spec:specific[<test-file-location>]`
+
+where in this last case the [ ] characters must be literally placed.
+
+For more information on accepted commands for testing with Rake you can execute:
```
unleashed: you could also document how to use rspec directly with `bundle exec rspec <parameters> <spec_file>` to run a specs file.

Rakefile
```diff
@@ -79,31 +79,37 @@ if saas?
...
       desc 'Run RSpec use_cases tests'
       RSpec::Core::RakeTask.new(:use_cases) do |task|
         task.pattern = 'spec/use_cases/**/*_spec.rb'
+        task.rspec_opts = '--require=spec_helper'
       end
 
       desc 'Run RSpec test/s specified by input file pattern'
       RSpec::Core::RakeTask.new(:specific, :test_name) do | task, task_args |
         task.pattern = "#{task_args[:test_name]}"
+        task.rspec_opts = '--require=spec_helper'
```
unleashed: these changes should also be squashed in the commit introducing them - perhaps add a comment in the first line doing this to explain why it is needed.

Rakefile
```diff
@@ -65,8 +65,48 @@ if saas?
...
+
+      desc 'Run RSpec api tests'
+      RSpec::Core::RakeTask.new(:api) do |task|
+        task.pattern = 'spec/api/**/*_spec.rb'
+      end
+
+      desc 'Run RSpec use_cases tests'
+      RSpec::Core::RakeTask.new(:use_cases) do |task|
+        task.pattern = 'spec/use_cases/**/*_spec.rb'
+      end
```
unleashed: Nice job - however, when I look at all these rules there is a DRY flag being raised: they all do the same except for a very small thing. This would be a bit of advanced stuff, but can we come up with a way to have Ruby write all these for us?

unleashed:
@miguelsorianod take a look at the "outdated" comments too, as I was writing some of them on invididual diffs that later got further modified.

README.md
```diff
@@ -102,6 +102,44 @@ variables for the test scripts like so:
...
+
+> `$ bundle exec rake`
+
+In case you need it/want it, it is possible to manually stop the services by executing:
+> `$ script/services stop`
+
+Another alternative to execute all tests from within the development environment
+where you should not need to manually start/stop the services before is:
+
+> `$ script/test`
```
davidor: I think this should be the first one to appear.

The documentation for running the tests has become quite complex with all the possible options and I think that the "normal way" (`script/test`) should be more visible.

miguelsorianod:
I have updated the README.md in order to have the basic steps needed to execute all the tests to be listed first in the testing section and then I added a subsection in the readme (running tests (advanced)) on how to execute the tests in more advanced ways

eguzki:

PR #12, opened 2018-04-09
 3scale/backend/manifest.rb: Config default cpus multiplier

miguelsorianod:
Refactor num workers multiplier into a constant.

When PUMA_WORKERS env var does not exist the following is done:

num workers  = multiplier * num_cpus

multiplier was a magic number in the code. Refactor this
magic number into the constant PUMA_WORKERS_CPUMULT.

### DIFF
--- a/spec/unit/manifest_spec.rb
+++ b/spec/unit/manifest_spec.rb
@@ -1,0 +1,47 @@
+module ThreeScale
+  module Backend
+    describe Manifest do
+      describe '.thread_safe?' do
+        it 'returns false' do
+          expect(described_class.thread_safe?).to eq(false)
+        end
+
+      end
+
+      describe '.compute_workers' do
+        let(:num_cpus) { ThreeScale::Backend::Util.number_of_cpus }
+
+        it 'returns PUMA_WORKERS_CPUMULT * <number of CPUs> if no PUMA_WORKERS environment variable is set' do
+          env_copy = ENV.to_h
+          env_copy.delete('PUMA_WORKERS')
+          stub_const('ENV', env_copy)
+          expect(described_class.compute_workers(num_cpus))
+            .to eq(num_cpus*described_class.singleton_class.const_get(:PUMA_WORKERS_CPUMULT))
+        end
+
+        it 'returns 0 if fork is not usable' do
+          stub_const('ENV', ENV.to_h.merge('PUMA_WORKERS' => '3'))
+          allow(Process).to receive(:respond_to?).with(:fork).and_return(false)
+          expect(described_class.compute_workers(num_cpus)).to eq(0)
+        end
+
+        it 'returns the specified number of workers in PUMA_WORKERS when set' do
+          stub_const('ENV', ENV.to_h.merge('PUMA_WORKERS' => '3'))
+          expect(described_class.compute_workers(ENV['PUMA_WORKERS'])).to eq(3)
+        end
+
+        it 'returns an exception when PUMA_WORKERS is not a number' do
+          stub_const('ENV', ENV.to_h.merge('PUMA_WORKERS' => 'test'))
+          expect { described_class.compute_workers(ENV['PUMA_WORKERS']) }.to raise_error(ArgumentError)
+        end
+
+        it 'returns PUMA_WORKERS_CPUMULT * <number of CPUs> when PUMA_WORKERS environment variable is empty' do
+          stub_const('ENV', ENV.to_h.merge('PUMA_WORKERS' => ''))
+          expect(described_class.compute_workers(num_cpus))
+            .to eq(num_cpus*described_class.singleton_class.const_get(:PUMA_WORKERS_CPUMULT))
+        end
+
+      end
+    end
+  end
+end
--- a/lib/3scale/backend/manifest.rb
+++ b/lib/3scale/backend/manifest.rb
@@ -8,6 +8,8 @@ module ThreeScale
       class << self
         PUMA_WORKERS = 'PUMA_WORKERS'.freeze
         private_constant :PUMA_WORKERS
+        PUMA_WORKERS_CPUMULT = 8
+        private_constant :PUMA_WORKERS_CPUMULT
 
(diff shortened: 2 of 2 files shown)


spec/unit/manifest_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        it 'returns false' do
+          expect(described_class.thread_safe?).to eq(false)
+        end
+
+      end
+
+      describe '.compute_workers' do
+
+        before :each do
+          ENV::delete('PUMA_WORKERS')
```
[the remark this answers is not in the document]
unleashed: Now about how to test this - it is not correct to modify the current process' environment as it modifies global state for all the tests. What you need is to mock the retrieval of `"PUMA_WORKERS"` from the `ENV` object. This should be done by mocking `ENV#[]` when the argument is `"PUMA_WORKERS"`, something similar to:
```ruby
allow(ENV).to receive(:[]).with("PUMA_WORKERS").and_return(nil)
```

spec/unit/manifest_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+
+        it 'returns 8(value of PUMA_WORKERS_CPUMULT) * <number of CPUs> if no PUMA_WORKERS environment variable is set' do
+          expect(described_class.compute_workers(Etc.nprocessors)).to eq(Etc.nprocessors*8)
```
unleashed: Here we also want to avoid coupling to the actual value of `PUMA_WORKERS_CPUMULT`, so we should avoid any hardcoding of values. Use `described_class.const_get :PUMA_WORKERS_CPUMULT` instead of `8` and change the spec description to also avoid the actual value.

spec/unit/manifest_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+
+        it 'returns 0 if fork is not usable' do
+          ENV['PUMA_WORKERS'] = '3'
```
unleashed: as above, this should be mocked - same thing for the examples below.

spec/unit/manifest_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+
+        it 'returns an exception when PUMA_WORKERS is not a number' do
+          ENV['PUMA_WORKERS'] = 'test'
+          expect { described_class.compute_workers(ENV['PUMA_WORKERS']) }.to raise_error(ArgumentError)
+        end
+
+        it 'returns 8 (default value) when PUMA_WORKERS environment variable is empty' do
+          ENV['PUMA_WORKERS'] = ''
+          expect(described_class.compute_workers(1)).to eq(8)
```
unleashed: same thing as above about hardcoding the actual value `8`.

unleashed:
also, another minor style nit: in commits, avoid the `.rb` extension, as it is implied and takes space from the title

spec/unit/manifest_spec.rb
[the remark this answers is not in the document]
mikz: @unleashed I'm not really convinced mocking is the right way. Because then you are really coupling it to the implementation - which could use any of the dozen ways to get the env `.fetch`, `.to_h`, `assoc`, `each`, `values_at`, ...

Standard way is to have global around filter that does `env = ENV.to_h; yield; ENV.replace(env)`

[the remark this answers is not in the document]
unleashed: You'd need to place the `replace` in an `ensure` statement and call it with a block - the whole thing looks rather unergonomic. You can also make a double out of a `ENV` and ensure the specific key is not there - in fact there is `stub_const("ENV", env_value)` which seems to be designed specifically for this https://relishapp.com/rspec/rspec-mocks/v/3-7/docs/mutating-constants/stub-defined-constant

spec/unit/manifest_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'etc'
+
+module ThreeScale
+  module Backend
+    describe Manifest do
+      describe '.thread_safe?' do
+
```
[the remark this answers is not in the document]
unleashed: yes, that would be a good idea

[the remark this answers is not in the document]
mikz: or hound which should be even simpler to configure

spec/unit/manifest_spec.rb
eguzki: we should not use ENV inside functions or methods. Unittesting them becomes difficult.

miguelsorianod:
I have updated the commits with the requested changes

spec/unit/manifest_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        it 'returns false' do
+          expect(described_class.thread_safe?).to eq(false)
+        end
+
+      end
+
+      describe '.compute_workers' do
+        it 'returns PUMA_WORKERS_CPUMULT * <number of CPUs> if no PUMA_WORKERS environment variable is set' do
+          stub_const('ENV', ENV.to_h.merge('PUMA_WORKERS' => nil))
+          expect(described_class.compute_workers(Etc.nprocessors))
```
unleashed: sorry I missed this the first time - `Etc.nprocessors` is an implementation detail. You have the number of CPUs actually offered by a method in the `Manifest` module, so use that (also use a `let` assignment and reuse in other examples).

unleashed: (the number of CPUs actually comes from a helper in `ThreeScale::Backend::Util`)

davidor:
It's not entirely related to this PR, but I see that you added tests for `compute_workers`. I wonder if this is the method that we should be testing. The reason is that `compute_workers` is only called from `Manifest.servel_model`, so I don't think it should be part of the public API of the class.

unleashed:
@davidor yes I think we are missing the test for that other method - perhaps add it in another PR @miguelsorianod ?

unleashed:
bors r=@unleashed,@eguzki,@davidor,@mikz

PR #13, opened 2018-04-09
Bugfix/Ensure an application without state can't be saved

miguelsorianod:
When saving an application, the existence of the application's state attribute was not checked.
Update Application logic to check for this and return an exception if not specified.

### DIFF
--- a/app/api/internal/applications.rb
+++ b/app/api/internal/applications.rb
@@ -18,8 +18,13 @@ module ThreeScale
           end
           app_attrs[:service_id] = service_id
           app_attrs[:id] = id
-          app = Application.save(app_attrs)
-          [201, headers, { status: :created, application: app.to_hash }.to_json]
+          begin
+            app = Application.save(app_attrs)
+          rescue ApplicationHasNoState => e
+            [400, headers, { status: :bad_request, error: e.message }.to_json]
+          else
+            [201, headers, { status: :created, application: app.to_hash }.to_json]
+          end
         end
 
         put '/:id' do |service_id, id|
@@ -27,8 +32,13 @@ module ThreeScale
           modified = Application.exists?(service_id, id)
           app_attrs[:service_id] = service_id
           app_attrs[:id] = id
-          app = Application.save(app_attrs)
-          { status: modified ? :modified : :created, application: app.to_hash }.to_json
+          begin
+            app = Application.save(app_attrs)
+          rescue ApplicationHasNoState => e
+            [400, headers, { status: :bad_request, error: e.message }.to_json]
+          else
+            { status: modified ? :modified : :created, application: app.to_hash }.to_json
+          end
         end
 
         delete '/:id' do |service_id, id|
--- a/lib/3scale/backend/application.rb
+++ b/lib/3scale/backend/application.rb
@@ -178,6 +178,8 @@ module ThreeScale
       end
 
       def save
+        raise ApplicationHasNoState.new(id) if !state
+
         storage.pipelined do
           persist_attributes
           persist_set
--- a/lib/3scale/backend/errors.rb
+++ b/lib/3scale/backend/errors.rb
@@ -373,5 +373,11 @@ module ThreeScale
       end
     end
 
+    class ApplicationHasNoState < BadRequest
+      def initialize(id)
+        super %(Application with id="#{id}" has no state )
+      end
+    end
+
   end
 end
--- a/spec/acceptance/api/internal/applications_api_spec.rb
+++ b/spec/acceptance/api/internal/applications_api_spec.rb
@@ -85,6 +85,17 @@ resource 'Applications (prefix: /services/:service_id/applications)' do
       # The returned data should not contain *some_param* attribute
       expect(app.to_hash).to eq application.merge(user_required: false)
     end
+
+    context 'with an application that has no state' do
+      let (:state) { nil }
+
+      example_request 'Trying to create the application' do
+        expect(status).to eq 400
+        expect(response_json['status']).to eq 'bad_request'
+        expect(response_json['error']).to match /has no state/i
+      end
+    end
+
   end
 
   put '/services/:service_id/applications/:id' do
@@ -155,6 +166,16 @@ resource 'Applications (prefix: /services/:service_id/applications)' do
         expect(status).to eq 400
         expect(response_json['status']).to eq 'error'
         expect(response_json['error']).to match /missing parameter/i
+      end
+    end
+
+    context 'with an application that has no state' do
+      let (:state) { nil }
+
+      example_request 'Trying to create/update the application' do
+        expect(status).to eq 400
+        expect(response_json['status']).to eq 'bad_request'
+        expect(response_json['error']).to match /has no state/i
       end
     end
   end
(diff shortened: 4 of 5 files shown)


lib/3scale/backend/application.rb
```diff
@@ -39,15 +39,20 @@ def attribute_names
...
-          return nil unless state
+          storage_results = storage.pipelined do
+            storage.mget(storage_key(service_id, id, :state),
+                         storage_key(service_id, id, :plan_id),
+                         storage_key(service_id, id, :plan_name),
+                         storage_key(service_id, id, :redirect_url),
+                         storage_key(service_id, id, :user_required))
+
+            #We do not call the exists? method due to it is memoized
+            storage.sismember(applications_set_key(service_id), id)
```
eguzki: - This operation is O(n), if there are lots of applications in the set, this could be expensive. Is it worth doing? just rising the question.

- Another option could be first check **storage.sismember**. Only when it exists, run **mget**. That's an idea.

davidor: It's O(1) @eguzki : https://redis.io/commands/sismember

eguzki: correct

test/unit/application_test.rb
```diff
@@ -66,6 +66,11 @@ def setup
     end
   end
 
+  test 'load returns nil if application does not exist' do
+    app = Application.load('1003', '2003')
```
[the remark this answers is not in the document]
miguelsorianod: In the file I see a method named Test::Unit::TestCase method defined (https://ruby-doc.org/stdlib-1.8.7/libdoc/test/unit/rdoc/Test/Unit/TestCase.html#method-i-setup)

That performs a flush of the database and resets the memoizer. I understand this would ensure that no Application/Service/any content exists at the beginning of each test.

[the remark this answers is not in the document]
unleashed: ah, that could work, yes

miguelsorianod:
I just updated new changes that revert the previous ones that I made and now I approach the problem by relying on checking the application existence from its state as originally, but now we try to make sure that if an application is created without a state then an exception is returned and this is also handled in the internal api.

I've kept the previous commits by commiting a revert of those commits just to have them available for the moment. If the new solution is good then I'll remove the previous original commits and its commit reversions.

miguelsorianod:
Finally I have removed the previous commits and its reverts so you can have a clean commit history for this PR and I keep those previous changes in a local branch of mine.

lib/3scale/backend/errors.rb
```diff
@@ -67,6 +67,12 @@ def initialize(id, user_key)
       end
     end
 
+    class ApplicationHasNoState < Error
```
eguzki: I would let exception middleware catche the error and set http status code on Error type. So, I would inherit from BadRequest error class.

app/api/internal/applications.rb
```diff
@@ -18,17 +18,25 @@ module API
           end
           app_attrs[:service_id] = service_id
           app_attrs[:id] = id
-          app = Application.save(app_attrs)
-          [201, headers, { status: :created, application: app.to_hash }.to_json]
+          begin
+            app = Application.save(app_attrs)
+            [201, headers, { status: :created, application: app.to_hash }.to_json]
+          rescue ApplicationHasNoState
```
eguzki: Why should not be catched by exception_catcher middleware?

app/api/internal/applications.rb
```diff
@@ -18,17 +18,25 @@ module API
...
           app_attrs = api_params Application
           modified = Application.exists?(service_id, id)
           app_attrs[:service_id] = service_id
           app_attrs[:id] = id
-          app = Application.save(app_attrs)
-          { status: modified ? :modified : :created, application: app.to_hash }.to_json
+          begin
+            app = Application.save(app_attrs)
+            { status: modified ? :modified : :created, application: app.to_hash }.to_json
+          rescue ApplicationHasNoState
```
eguzki: Why should not be catched by exception_catcher middleware?

unleashed: because this is the internal API - it does not run the exception catcher and if it did that would be broken because the exception catcher responds with XML stuff while this API should respond JSON

eguzki: Then, it would be nice to have its own exception catcher processing responds as json bodies. Makes application logic easier to develop and more readable.

app/api/internal/applications.rb
```diff
@@ -18,17 +18,25 @@ module API
           end
           app_attrs[:service_id] = service_id
           app_attrs[:id] = id
-          app = Application.save(app_attrs)
-          [201, headers, { status: :created, application: app.to_hash }.to_json]
+          begin
+            app = Application.save(app_attrs)
+            [201, headers, { status: :created, application: app.to_hash }.to_json]
```
unleashed: minor nit: this statement will not raise `ApplicationHasNoState` - it can either go outside or in the `else` clause of the rescue:
```ruby
begin
  app = Application.save(app_attrs)
rescue ApplicationHasNoState
  # ...
else
  [201, headers, { status: :created, application: app.to_hash }.to_json]
end
```

app/api/internal/applications.rb
```diff
@@ -18,17 +18,25 @@ module API
...
+          rescue ApplicationHasNoState
+            [400, headers, { status: :bad_request, error: 'application cannot be created, it has no state' }.to_json]
```
unleashed: this does not look bad, but the caught exception could already encode the `bad_request` status and the error message and use that here, which would help in catching several errors at once - perhaps it's already encoding that?

app/api/internal/applications.rb
miguelsorianod: I've moved the response encoding to an else of the rescue clause. I've not moved it outside because in the case of an exception it would execute both the content inside the rescue and the content outside of it.

app/api/internal/applications.rb
miguelsorianod: I have set ApplicationHasNoState as a BadRequest type of exception and then I obtain the http status code and the message from the exception object. I have pushed the changes

miguelsorianod: Finally, after discussing it we do not set the http status code from the received exception because the received exception is from the Model and we do not want to rely on the http status code returned by the model in case it changes

spec/acceptance/api/internal/applications_api_spec.rb
```diff
@@ -85,6 +85,17 @@
       # The returned data should not contain *some_param* attribute
       expect(app.to_hash).to eq application.merge(user_required: false)
     end
+
+    context 'With an application that has no state' do
```
unleashed: use downcase with specs unless referring to entities in code, concepts, etc

spec/acceptance/api/internal/applications_api_spec.rb
```diff
@@ -85,6 +85,17 @@
...
+      let (:state) { nil }
+
+      example_request 'Trying to create the application' do
```
unleashed: (this is for the documentation of the API, so no need to downcase :)

spec/acceptance/api/internal/applications_api_spec.rb
miguelsorianod: Done

miguelsorianod:
Updated

eguzki:
bors r=davidor,unleashed,eguzki

PR #14, opened 2018-04-10
exception_catcher: return XML response body when invalid %encoding is detected

miguelsorianod:
When performing a request with an invalid percentage-encoding
format we now return an XML from a Backend::Error type of error.
Previously when an invalid percentage-encoding was detected
the response was a sinatra error.

This has been done in the same way that was done with the response error management of INVALID_BYTE_SEQUENCE_ERR_MSG

### DIFF
--- a/test/integration/listener_test.rb
+++ b/test/integration/listener_test.rb
@@ -47,6 +47,21 @@ class ListenerTest < Test::Unit::TestCase
     assert_equal 'required_params_missing', node['code']
   end
 
+  def test_invalid_percent_encoding
+    # We pass a string and not a hash as a parameter due to
+    # the rack-test library performs modifications to the passed values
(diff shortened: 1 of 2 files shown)


test/integration/listener_test.rb
```diff
@@ -47,6 +47,21 @@ def test_missing_required_parameters
     assert_equal 'required_params_missing', node['code']
   end
 
+  def test_invalid_percent_encoding
+    #We pass a string and not a hash as a parameter due to
```
unleashed: minor nit: leave a space between the comment symbol and the start of the comment for readability

miguelsorianod: Done :+1:

davidor:
bors r=@unleashed,@davidor

PR #15, opened 2018-04-10
Feature/configurable current notify metrics

miguelsorianod:
We make current master service metrics configurable via the 3scale_backend.conf file by setting environment variables.

Current tests are passing but it is a change that is important to be sure it does not break anything. Probably some kind of unit test would be a good idea? I don't find/can't see any configuration unit test

### DIFF
--- a/lib/3scale/backend/configuration.rb
+++ b/lib/3scale/backend/configuration.rb
@@ -12,6 +12,11 @@ module ThreeScale
       end
     end
 
+    CONFIG_MASTER_METRICS_TRANSACTIONS_DEFAULT = "transactions".freeze
+    private_constant :CONFIG_MASTER_METRICS_TRANSACTIONS_DEFAULT
+    CONFIG_MASTER_METRICS_TRANSACTIONS_AUTHORIZE_DEFAULT = "transactions/authorize".freeze
+    private_constant :CONFIG_MASTER_METRICS_TRANSACTIONS_AUTHORIZE_DEFAULT
+
     @configuration = Configuration::Loader.new
 
     # assign @configuration first, since code can depend on the attr_reader
@@ -35,6 +40,11 @@ module ThreeScale
       config.add_section(:statsd, :host, :port)
       config.add_section(:internal_api, :user, :password)
       config.add_section(:oauth, :max_token_size)
+      config.add_section(:master, :metrics)
+
+      # Configure nested fields
+      master_metrics = [:transactions, :transactions_authorize]
+      config.master.metrics = Struct.new(*master_metrics).new
 
       # Default config
       config.master_service_id  = 1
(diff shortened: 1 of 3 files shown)


lib/3scale/backend/configuration.rb
```diff
@@ -60,6 +72,16 @@ def configure
         ENV['CONFIG_FILE']
       ].compact)
 
+      # Assign default values to some configuration values
+      # that might been set in the config file but their
+      # environment variables not.
+      if config.master.metrics.transactions.to_s.empty?
+        config.master.metrics.transactions = CONFIG_MASTER_METRICS_TRANSACTIONS_DEFAULT
+      end
```
unleashed: In the else case you still want to make sure that the contents of `config.master.metrics.transactions` is a `String` - either call `to_s` and re-assign it before or do it in the else branch. Same thing for the setting below.

miguelsorianod: I understand this would be needed due to currently the user can configure a Ruby datatype directly so we try to convert it to String just in case has introduced a different data type.
I have updated the code to do this.

lib/3scale/backend/configuration.rb
```diff
@@ -60,6 +72,22 @@ def configure
...
 
+      # Assign default values to some configuration values
+      # that might been set in the config file but their
+      # environment variables not. Also make sure
+      # that what we have is a String, just in case
+      # from the config file a type of data different than a
+      # String has been assigned to this configuration
+      # parameter
+      config.master.metrics.transactions = config.master.metrics.transactions.to_s
+      if config.master.metrics.transactions.empty?
```
davidor: A user can assign `nil` in his config file. Is `nil` a value we should accept? If it's not, we should also check for `nil` here.

davidor: Also, if we check for nil here, we could delete the 2 assignments above.

lib/3scale/backend/transactor/notify_batcher.rb
```diff
@@ -12,21 +12,17 @@ module NotifyBatcher
         include Resque::Helpers
         include Configurable
 
-        METRIC_AUTHORIZE = 'transactions/authorize'.freeze
-        METRIC_TRANSACTIONS = 'transactions'.freeze
-        private_constant :METRIC_AUTHORIZE, :METRIC_TRANSACTIONS
-
         def notify_authorize(provider_key)
-          notify(provider_key, METRIC_AUTHORIZE => 1)
+          notify(provider_key, configuration.master.metrics.transactions_authorize => 1)
```
davidor: Have you considered keeping the constants?
```ruby
METRIC_AUTHORIZE = configuration.master.metrics.transactions_authorize
METRIC_TRANSACTIONS = configuration.master.metrics.transactions
private_constant :METRIC_AUTHORIZE, :METRIC_TRANSACTIONS
```

lib/3scale/backend/configuration.rb
miguelsorianod: Hi,

This was done intentionally in this way. When calling the .to_s method to an object that is nil the result is converted into the empty string:
```
irb(main):004:0> mystring = nil
=> nil
irb(main):005:0> mystring.to_s
=> ""
```

miguelsorianod: Regarding the redundancy of the 2 assignments above, I agree with you and I saw this too but it seemed to me that the "defaults assignment" logic was located there among the other currently set default configuration parameters so I kept that assignments there just to be clear (even if they are repeated code).
I can remove it without any problem if we prefer to avoid code duplication in this case.

lib/3scale/backend/transactor/notify_batcher.rb
eguzki: I am not sure it is a good pattern to mix private constants and configuration parameters

unleashed: but those are not constants, those are method calls that could return anything at invocation time - setting those as constants makes two assumptions:
- at load time of this file the methods will return something sane and validated.
- the values will never change after they are obtained

Both are very risky assumptions.

miguelsorianod: I agree with you that are risky assumptions and I thought the same regarding the second point you mentioned but after commeting it with davidor it's improbable for this values to change because if we changed this in the middle of the execution how would the other components respond?

If we cannot afford this assumptions then I agree we should not use the constants, because other factor that I thought that could be risky is the order on which the configuration is loaded.

davidor:

PR #16, opened 2018-04-12
refactor/provider_key_change_use_case

miguelsorianod:
The ProviderKeyChangeUseCase has been refactored to use the
Service class methods to clear the cache, removing duplication

### DIFF
--- a/spec/use_cases/provider_key_change_use_case_spec.rb
+++ b/spec/use_cases/provider_key_change_use_case_spec.rb
@@ -68,14 +68,35 @@ module ThreeScale
         let(:service_ids_with_old_key) { [7001, 7002] }
         let(:use_case) { ProviderKeyChangeUseCase.new(old_key, new_key) }
 
+        def exercise_provider_service_memoizer(provider_key, service_id)
+          # Call the Service class methods just to force the memoization
+          # of them with specific values
+          Service.default_id(provider_key)
+          Service.authenticate_service_id(service_id, provider_key)
+          Service.load(provider_key)
+          Service.load_by_id(service_id)
+          Service.list(provider_key)
+          Service.provider_key_for(service_id)
+        end
+
+        def build_provider_service_memoizer_keys(provider_key, service_id)
+          Memoizer.build_keys_for_class(Service,
+            authenticate_service_id: [service_id, provider_key],
+            default_id: [provider_key],
+            load: [provider_key],
+            load_by_id: [service_id],
+            list: [provider_key],
+            provider_key_for: [service_id])
+        end
+
         before do
           service_ids_with_old_key.each do |service_id|
             Service.save! id: service_id, provider_key: old_key
           end
-          use_case.process
         end
 
         it 'changes the provider key of existing services' do
+          use_case.process
           expect(Service.list(new_key))
               .to eq service_ids_with_old_key.map(&:to_s)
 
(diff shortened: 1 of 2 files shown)


spec/use_cases/provider_key_change_use_case_spec.rb
```diff
@@ -68,14 +68,36 @@ module Backend
         let(:service_ids_with_old_key) { [7001, 7002] }
         let(:use_case) { ProviderKeyChangeUseCase.new(old_key, new_key) }
 
+        def exercise_provider_service_memoizer(provider_key, service_id)
+          # Call the Service class methods just
+          # to force the memoization of them with
+          # specific values
```
unleashed: comment lines can be longer here

spec/use_cases/provider_key_change_use_case_spec.rb
```diff
@@ -85,20 +107,53 @@ module Backend
...
 
         it 'changes the provider key associated with the affected services' do
+          use_case.process
           service_ids_with_old_key.each do |service_id|
             expect(Service.provider_key_for(service_id)).to eq new_key
           end
         end
+
+        it 'clears the Service cache keys of the old provider key' do
+          service_ids_with_old_key.each do |service_id|
```
unleashed: this can be described in a new `context` or `describe` block if you split these examples in one part for the "general" specs and another one for the memoization stuff, and this way you would not need to insert `use_case.process` in all the other examples as a consequence of removing it from the `before` block (you would have two `describe`/`context` blocks, and each with independent `before` blocks if you do not nest them).

spec/use_cases/provider_key_change_use_case_spec.rb
```diff
@@ -68,14 +68,36 @@ module Backend
...
+          Service.default_id(provider_key)
+          Service.authenticate_service_id(service_id, provider_key)
+          Service.load(provider_key)
+          Service.load_by_id(service_id)
+          Service.list(provider_key)
+          Service.provider_key_for(service_id)
+        end
+
+        def build_provider_service_memoizer_keys(provider_key, service_id)
+          memoizer_keys = Memoizer.build_keys_for_class(Service,
```
davidor: no need to define `memoizer_keys`. It's not used anywhere.

miguelsorianod: That's true. I'll remove it

spec/use_cases/provider_key_change_use_case_spec.rb
```diff
@@ -85,20 +107,53 @@ module Backend
...
+            end
+          end
+        end
+
+        it 'clears the Service cache keys of the new provider key' do
+          service_ids_with_old_key.each do |service_id|
+            exercise_provider_service_memoizer(new_key, service_id)
+            memoizer_keys = build_provider_service_memoizer_keys(new_key, service_id)
+
+            memoizer_keys.each do |memoized_key|
```
davidor: To test a change in the state of an object you can use the Rspec `change` matcher: https://relishapp.com/rspec/rspec-expectations/v/3-7/docs/built-in-matchers/change-matcher

In this case, we want to check how all the elements of an array changed, so using the `change` matcher becomes a bit more complex:
```
            expect { use_case.process }
                .to change { memoizer_keys.count { |k| Memoizer.memoized?(k) } }
                .from(memoizer_keys.size)
                .to(0)
```

I'm not proposing to make the change, just wanted to let you know that the matcher exists.

unleashed: Why not proposing it? Seems sensible.

unleashed: It's not only a matter of preference: it vastly simplifies the code, leaving less room for error, and it reads way better.

miguelsorianod: I have changed it and now uses the change matcher

unleashed:
bors r=@davidor,@unleashed

PR #17, opened 2018-04-16
transactor/status_test: Add application/user_id existence testing

miguelsorianod:
Add unit tests that cover application_id and user_id existence in responses

### DIFF
--- a/test/unit/transactor/status_test.rb
+++ b/test/unit/transactor/status_test.rb
@@ -237,5 +237,33 @@ module Transactor
       assert_nil           month['exceeded']
       assert_equal 'true', day['exceeded']
     end
+
+    test '#to_xml shows the application id when OAuth is used' do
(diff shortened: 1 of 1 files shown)


test/unit/transactor/status_test.rb
```diff
@@ -237,5 +237,33 @@ def setup
...
+                       username: "testusername")
+      status = Transactor::Status.new(service_id: @service_id,
+                                      application: @application,
+                                      values: usage,
+                                      oauth: true,
+                                      user: user
+                                      )
+      doc = Nokogiri::XML(status.to_xml)
+
+      assert_equal "testusername", doc.at('status user id').content
```
davidor: You could define `test_user_name` at the beginning of the test and then use it to when instantiating the user and also here. It would avoid duplication.

unleashed: +1

eguzki:

PR #19, opened 2018-04-17
Add more error messages for the transactions.xml endpoint 

miguelsorianod:
We now return a proper 400 error code with some information in XML
format when the following situations are detected:

· The transactions parameter is blank
· The transactions parameter is not a hash
· Some transaction of the transactions parameter is nil

Also, we have added temporary logging to confirm that the params hash can never be nil and we'll remove the checking of the nil value and the logging if we do not detect related logs for a while.

### DIFF
--- a/lib/3scale/backend/listener.rb
+++ b/lib/3scale/backend/listener.rb
@@ -9,6 +9,8 @@ module ThreeScale
       enable :raise_errors
       disable :show_exceptions
 
+      include Logging
+
       ## ------------ DOCS --------------
       ##~ namespace = ENV['SAAS_SWAGGER'] == "1" ? "Service Management API" : "Service Management API (on-premises)"
       ##~ sapi = source2swagger.namespace(namespace)
@@ -416,7 +418,10 @@ module ThreeScale
         check_post_content_type!
 
         # 403 Forbidden for consistency (but we should return 400 Bad Request)
-        halt 403 if params.nil?
+        if params.nil?
+          logger.notify("listener: params hash is nil in method '/transactions.xml'")
+          halt 403
+        end
 
         # returns 403 when no provider key is given, even if other params have an invalid encoding
         provider_key = params[:provider_key] ||
--- a/lib/3scale/backend/listener.rb
+++ b/lib/3scale/backend/listener.rb
@@ -433,12 +433,7 @@ module ThreeScale
         check_params_value_encoding!(params, REPORT_EXPECTED_PARAMS)
 
         transactions = params[:transactions]
-
-        if blank?(transactions) ||
-            !transactions.is_a?(Hash) ||
-            transactions.any? { |_id, data| data.nil? }
-          halt 400
-        end
+        check_transactions_validity(transactions)
 
         Transactor.report(provider_key, params[:service_id], transactions, response_code: 202, request: request_info)
         202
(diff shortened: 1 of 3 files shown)


lib/3scale/backend/listener.rb
```diff
@@ -413,10 +415,14 @@ def do_api_method(method_name)
       ##
       ##
       post '/transactions.xml' do
+
```
unleashed: extra blank line here

lib/3scale/backend/listener.rb
```diff
@@ -582,6 +577,20 @@ def check_post_content_type!
...
+        if blank?(transactions)
+          raise TransactionsIsBlank
+        end
+
+        if !transactions.is_a?(Hash)
+          raise TransactionsFormatInvalid
+        end
+
+        if transactions.any? { |_id, data| data.nil? }
+          raise TransactionsHasNilTransaction
```
unleashed: can this actually happen? all the cases I imagine where a hash has a key the "data"/value part of it should not be nil but at most an empty string.

test/integration/report_test.rb
```diff
@@ -831,7 +831,15 @@ def setup
                             1 => nil }
 
     assert_equal 400, last_response.status
-    assert_equal '', last_response.body
+    assert_not_equal '', last_response.body
+
+
```
unleashed: extra blank line

lib/3scale/backend/listener.rb
miguelsorianod: This can happen.

For example, I reproduced the problem by performing the following request (see the last transactions specified without values):

```
[<username>@localhost apisonator]$ curl -X POST -v '127.0.0.1:3000/transactions.xml' -d 'provider_key=provider_keytest&transactions[0][app_id]=5207&transactions[0][usage][hits]=1&transactions[0][timestamp]=2018-04-17 10:13:11 UTC&transactions[1]'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 3000 (#0)
> POST /transactions.xml HTTP/1.1
> Host: 127.0.0.1:3000
> User-Agent: curl/7.55.1
> Accept: */*
> Content-Length: 156
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 156 out of 156 bytes
< HTTP/1.1 400 Bad Request
< Content-Type: application/vnd.3scale-v2.0+xml
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: ETag, Link, 3scale-rejection-reason
< X-Content-Type-Options: nosniff
< Transfer-Encoding: chunked
< 
* Connection #0 to host 127.0.0.1 left intact
<?xml version="1.0" encoding="UTF-8"?><error code="transactions_has_nil_transaction">transactions has a nil transaction</error>[<username>@localhost apisonator]$ 
```

The test 'returns error if any of the reported transactions are nil' in test/integration/report_test.rb also reproduces the nil error message.

unleashed: ok :+1:

lib/3scale/backend/listener.rb
miguelsorianod: I have fixed the formatting issues with blank the blank lines in a new forced push

lib/3scale/backend/listener.rb
```diff
@@ -415,7 +415,6 @@ def do_api_method(method_name)
       ##
       ##
       post '/transactions.xml' do
-
```
unleashed: this diff belongs to the previous commit

unleashed:
bors r=@eguzki,@unleashed

PR #20, opened 2018-04-17
user: remove version attribute

eguzki:
unused *version* attribute removed from User class


unleashed:
bors r=@davidor,@unleashed

PR #21, opened 2018-04-23
usage_limit: remove seconds from periods set

eguzki:
"seconds" level of granularity does not exist for usage limits. It was tried to load and never found.

### DIFF
--- a/lib/3scale/backend/usage_limit.rb
+++ b/lib/3scale/backend/usage_limit.rb
@@ -3,7 +3,7 @@ module ThreeScale
     class UsageLimit
       include Storable
 
-      PERIODS = Period::ALL_DESC
+      PERIODS = (Period::ALL_DESC - [Period::Second]).freeze
 
       attr_accessor :service_id, :plan_id, :metric_id, :period, :value
 
--- a/test/unit/usage_limit_test.rb
+++ b/test/unit/usage_limit_test.rb
@@ -75,6 +75,19 @@ class UsageLimitTest < Test::Unit::TestCase
     assert usage_limits.empty?, 'Expected usage_limits to be empty'
   end
 
+  def test_usage_limit_periods_do_not_include_second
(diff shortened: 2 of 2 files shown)


lib/3scale/backend/usage_limit.rb
```diff
@@ -3,7 +3,7 @@ module Backend
     class UsageLimit
       include Storable
 
-      PERIODS = Period::ALL_DESC
+      PERIODS = Period::ALL_DESC.reject { |period| period == :second }
```
unleashed: `.freeze`?

davidor: Another option:
```ruby
Period::ALL_DESC - [:second]
```

davidor:
I think we should add tests for this.

unleashed:
@eguzki at minimum we need the constant frozen to merge this

lib/3scale/backend/usage_limit.rb
eguzki: This
```
Period::ALL_DESC - [:second]
```
does not work. It is an array of granularity class objects created [here](https://github.com/3scale/apisonator/blob/0c290d4eb00584f1c51acee274ffa4373d080443/lib/3scale/backend/period/period.rb#L224)

But somehow (not investigated yet why)
```
Period::ALL_DESC.reject { |period| period == :second }
```
works

unleashed: Because the equality is overridden - set operations are probably looking for an implementation of `hash` which differs between different objects, so they are not seen as equal. Using `Period::Second` or `Period.from :second` would likely work.

eguzki:
unittests added

unleashed:
@davidor perhaps the edit was not caught by bors?

bors r=@unleashed,@davidor

unleashed:
heisentest again :/

bors r=@unleashed,@davidor

PR #22, opened 2018-04-24
Fix default notification batch when left unset

unleashed:
The `notification_batch` setting would be set to 0 when unset in the configuration, which would force lots of `NotifyJob`s to be created. Make it default to a reasonable value when not specified.


eguzki:
good job ;)

PR #23, opened 2018-04-25
Makefile: prefix default APISONATOR_REL with "v" and export in ci-build

unleashed:
This switches to have the tag prefixed by default with "v" like the
apisonator images (non CI), and makes sure that this setting is
passed down to the ci-flatten target when building.


unleashed:
bors r=@davidor

PR #24, opened 2018-04-25
Explain how to release a new version in a section of the README.md

unleashed:
This explains a set of steps to be performed when releasing a new version, and includes a few helpers in Rakefile to generate a shortlog and to produce proper tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,3 +205,25 @@ This is our basic integration flow:
 4. Merge to master.
 
 We keep stable branches receiving bug and security fixes for long term releases.
+
+### Releasing a new version
+
+Currently we need to follow this process:
+
+1. Change the contents of version.rb with the new version.
+2. Run `bundle install` with all Gemfiles to update all lockfiles.
+3. Run `rake license_finder:report:xml > licenses.xml`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -205,3 +205,25 @@ This is our basic integration flow:
 4. Merge to master.
 
 We keep stable branches receiving bug and security fixes for long term releases.
+
+### Releasing a new version
```
davidor: Should this be `##` ? So it's the same as `Integration flow`.
Also, have you considered creating a separated file for this and link it here?

unleashed: yes, I have considered both - the integration flow is what we have now that most closely resembles a "developer" section, and for creating a new file we still don't have enough content, but certainly looking forward to adding it once we get more material.

davidor: I prefer the option of hiding content that it's only useful for a few people and leave a simpler README for everyone, but this is just a matter of preference.

unleashed:
bors r=@davidor

README.md
unleashed: yes, and I expect we will do that once we add a bit more content.

PR #25, opened 2018-04-25
Release 2.86.0

unleashed:
Generated using the steps defined in README.md.


unleashed:
Ok, waiting for #21 before releasing.

CHANGELOG.md
```diff
@@ -2,6 +2,81 @@
...
+
+- Running make in systems with a non-GNU version of `time` should stop failing
+  because of using unsupported options.
+- Fixed a bug that made extra requests for stats with the inexistent `seconds`
+  period.
+- Fixed a bug in dealing with the `notification_batch` configuration that made
+  Apisonator enqueue NotifyJobs continuously when the setting was left
+  unspecified.
+
+### Shortlog v2.85.0..v2.86.0
```
davidor: I'm not sure it's worth it to have all the commits here. Maybe it would be more useful to have links to PRs in the section above.

unleashed: I am not very convinced of this either, and linking to PRs is a good idea (and we would need to change the text in the README then), just a bit more work.

This in particular is inspired on similar practices in other projects such as GNU's, which are even more detailed with changes per file. There's another option I think might make sense to consider as well, which is creating a directory for release notes and then adding this commit summary and other information there (ie. back compat stuff, known issues, etc) like the git project does.

It might be too heavy for frequent releases though. Will wait for more feedback from the team.

eguzki: I would attach PR links to previous sections points something like 
```
* Added extra measures to avoid creating applications through the internal API without mandatory attributes. PR #17
```

unleashed:
wait, @davidor, I will be updating stuff for having the PRs linked and a rake task to autolink them.

Rakefile
```diff
@@ -2,14 +2,9 @@
...
+require '3scale/tasks/helpers/environment'
 
-def testable_environment?
-  !%w(preview production).include?(ENV['RACK_ENV'])
-end
-
-def saas?
-  ThreeScale::Backend.configuration.saas
-end
+include ThreeScale::Backend::Tasks::Environment
```
davidor: It should be `ThreeScale::Backend:Tasks::Helpers::Environment` , right?

unleashed:
@davidor ready to check again

lib/3scale/tasks/helpers/command.rb
```diff
@@ -0,0 +1,65 @@
+module ThreeScale
+  module Tasks
+    module Helpers
+      module Command
```
davidor: If this module is not used, I'd rather not include it.

CHANGELOG.md
```diff
@@ -2,6 +2,36 @@
...
+### Added
+
+- The metrics used for reporting transactions and authorizations to the master
+  account are now configurable under `master.metrics.transactions` and
+  `master.metrics.transactions_authorize`. ([#15](https://github.com/3scale/apisonator/pull/15))
+- There is now a set of rake targets for selectively run tests and specs by type
+  (ie. integration, unit, etc) and by specific file. ([#11](https://github.com/3scale/apisonator/issues/11))
+- The `make dev` command supports specifying an env variable `PORT` to expose
+  the listener in that port on the local machine. ([#8](https://github.com/3scale/apisonator/pull/8))
+- Minor miscellaneous improvements. ([#7](https://github.com/3scale/apisonator/pull/7), [#9](https://github.com/3scale/apisonator/pull/9), [#12](https://github.com/3scale/apisonator/issues/12), [#16](https://github.com/3scale/apisonator/pull/16), [#17](https://github.com/3scale/apisonator/pull/17), pr [#20](https://github.com/3scale/apisonator/pull/20), [#23](https://github.com/3scale/apisonator/pull/23), [#24](https://github.com/3scale/apisonator/pull/24))
```
davidor: There's a `pr` there that should be deleted.

lib/3scale/tasks/helpers/command.rb
unleashed: This is a byproduct of this work - the issue is how likely are you to have a use for it in the future?

In fact, there is a similar `run_command` method elsewhere in the tasks, but that case was solved with just `system`, because there was no need to capture output. This is likely to be useful in the future - not including it means losing it while having good chances someone will have to solve the same issue again.

CHANGELOG.md
unleashed: good catch, that is a leftover

lib/3scale/tasks/helpers/command.rb
eguzki: IMO code that is not used should not be included. 

If it is very likely to be used in near future, unmerged PR with the method would be appropriate.

lib/3scale/tasks/helpers/command.rb
```diff
@@ -0,0 +1,65 @@
...
+        module_function
```
eguzki: this shouldn't be like this for module functions?

```ruby
module_function :run
```

davidor: Both are correct @eguzki : https://ruby-doc.org/core-2.4.1/Module.html#method-i-module_function
```
If used with no arguments, subsequently defined methods become module functions.
```

lib/3scale/tasks/helpers/command.rb
unleashed: Ok, I will remove this from the PR.

unleashed:
bors try

davidor:
bors r=@davidor,@eguzki

PR #26, opened 2018-05-02
Refactor/transactor status

miguelsorianod:
I have refactored some parts of the transactor/status class.

Also I have removed some unreferenced methods/code from the class. I removed the options on the to_xml method named 'exclude_application' and 'exclude_user' because are not referenced/set anywhere on the code and seem to be legacy options (it seems were added in commit 3424475acea289f715e7a876a6a880e3f799f464). Maybe @unleashed / @mikz that have been longer with the codebase can confirm this.

Thank you.

### DIFF
--- a/lib/3scale/backend/transactor/status.rb
+++ b/lib/3scale/backend/transactor/status.rb
@@ -124,12 +124,12 @@ module ThreeScale
           end
 
           hierarchy_reports = [] if @hierarchy_ext
-          if !@application.nil? && !options[:exclude_application]
+          if !@application.nil?
             add_plan_section(xml, 'plan'.freeze, plan_name)
             add_reports_section(xml, application_usage_reports)
             hierarchy_reports.concat application_usage_reports if hierarchy_reports
           end
-          if !@user.nil? && !options[:exclude_user]
+          if !@user.nil?
             add_plan_section(xml, 'user_plan'.freeze, user_plan_name)
             add_reports_section(xml, user_usage_reports, true)
             hierarchy_reports.concat user_usage_reports if hierarchy_reports
@@ -200,16 +200,12 @@ module ThreeScale
         end
 
         def add_authorize_section(xml)
-          xml << '<authorized>'
           if authorized?
-            xml << 'true' \
-                   '</authorized>'
+            xml << '<authorized>true</authorized>'.freeze
           else
-            xml << 'false' \
-                   '</authorized>' \
-                   '<reason>' \
-                   "#{rejection_reason_text}" \
-                   '</reason>'
+            xml << '<authorized>false</authorized><reason>'.freeze
+            xml << rejection_reason_text
+            xml << '</reason>'.freeze
           end
         end
 
@@ -224,7 +220,7 @@ module ThreeScale
 
         def add_user_section(xml)
           if !@user.nil?
-            xml << "<user><id>#{@user.username}</id></user>"
+            xml << "<user><id>".freeze << @user.username << "</id></user>".freeze
           end
         end
 
--- a/lib/3scale/backend/transactor/status.rb
+++ b/lib/3scale/backend/transactor/status.rb
@@ -118,42 +118,29 @@ module ThreeScale
         def to_xml(options = {})
           xml = ''
           xml << '<?xml version="1.0" encoding="UTF-8"?>'.freeze unless options[:skip_instruct]
-          xml << '<status><authorized>'.freeze
+          xml << '<status>'.freeze
 
-          if authorized?
-            xml << 'true</authorized>'.freeze
-          else
-            xml << 'false</authorized><reason>'.freeze
-            xml << rejection_reason_text
-            xml << '</reason>'.freeze
-          end
+          add_authorize_section(xml)
 
           if oauth
-            redirect_uri = application.redirect_url
-            xml << '<application>' \
-                   "<id>#{application.id}</id>" \
-                   "<key>#{application.keys.first}</key>" \
-                   "<#{@redirect_uri_field}>#{redirect_uri}</#{@redirect_uri_field}>" \
-                   '</application>'
-            if !@user.nil?
-              xml << "<user><id>#{@user.username}</id></user>"
-            end
+            add_application_section(xml)
+            add_user_section(xml)
           end
 
           hierarchy_reports = [] if @hierarchy_ext
           if !@application.nil? && !options[:exclude_application]
-            add_plan(xml, 'plan'.freeze, plan_name)
-            xml << aux_reports_to_xml(application_usage_reports)
+            add_plan_section(xml, 'plan'.freeze, plan_name)
+            add_reports_section(xml, application_usage_reports)
             hierarchy_reports.concat application_usage_reports if hierarchy_reports
           end
           if !@user.nil? && !options[:exclude_user]
-            add_plan(xml, 'user_plan'.freeze, user_plan_name)
-            xml << aux_reports_to_xml(user_usage_reports, true)
+            add_plan_section(xml, 'user_plan'.freeze, user_plan_name)
+            add_reports_section(xml, user_usage_reports, true)
             hierarchy_reports.concat user_usage_reports if hierarchy_reports
           end
 
           if hierarchy_reports
-            add_hierarchy(xml, hierarchy_reports)
+            add_hierarchy_section(xml, hierarchy_reports)
           end
 
           xml << '</status>'.freeze
@@ -194,7 +181,7 @@ module ThreeScale
           end
         end
 
-        def add_hierarchy(xml, reports)
+        def add_hierarchy_section(xml, reports)
           xml << '<hierarchy>'.freeze
           with_report_and_hierarchy(reports) do |ur, children|
             xml << '<metric name="'.freeze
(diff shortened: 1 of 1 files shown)


lib/3scale/backend/transactor/status.rb
```diff
@@ -211,11 +194,40 @@ def with_report_and_hierarchy(reports)
...
         end
 
-        def add_plan(xml, tag, plan_name)
+        def add_plan_section(xml, tag, plan_name)
           xml << "<#{tag}>"
           xml << plan_name.to_s.encode(xml: :text) << "</#{tag}>"
         end
 
+        def add_authorize_section(xml)
+          xml << '<authorized>'
```
mikz: This is not frozen and will allocate plenty of strings.

miguelsorianod:
Regarding the use of  Nokogiri::XML::Builder, I agree with you @eguzki 

The idea is to move this code out of transactor/Status.rb in another PR, in order to decouple the data representation from the model, along with trying to generalize the data representation to be compatible with other formats (like json).

lib/3scale/backend/transactor/status.rb
```diff
@@ -211,11 +194,40 @@ def with_report_and_hierarchy(reports)
...
-        def add_plan(xml, tag, plan_name)
+        def add_plan_section(xml, tag, plan_name)
           xml << "<#{tag}>"
           xml << plan_name.to_s.encode(xml: :text) << "</#{tag}>"
         end
 
+        def add_authorize_section(xml)
+          xml << '<authorized>'.freeze
+          if authorized?
+            xml << 'true'.freeze << '</authorized>'.freeze
```
unleashed: you can make this perform better with a single `xml << '<authorized>true</authorized>'.freeze`

lib/3scale/backend/transactor/status.rb
```diff
@@ -211,11 +194,40 @@ def with_report_and_hierarchy(reports)
...
+          else
+            xml << 'false'.freeze << '</authorized>'.freeze
+            xml << '<reason>'.freeze
```
unleashed: here you can do `xml << '<authorized>false</authorized><reason>'.freeze`

unleashed: the same principle applies to other places in the added methods

lib/3scale/backend/transactor/status.rb
miguelsorianod: :+1:

lib/3scale/backend/transactor/status.rb
miguelsorianod: :+1: . Done. In some other places I left it the way it is for readability

lib/3scale/backend/transactor/status.rb
```diff
@@ -211,11 +194,35 @@ def with_report_and_hierarchy(reports)
...
+            xml << rejection_reason_text
+            xml << '</reason>'.freeze
+          end
+        end
+
+        def add_application_section(xml)
+          redirect_uri = application.redirect_url
+          xml << "<application><id>".freeze << "#{application.id}" << "</id>".freeze
+          xml << "<key>".freeze << "#{application.keys.first}" << "</key>".freeze
+          xml << "<#{@redirect_uri_field}>#{redirect_uri}</#{@redirect_uri_field}>"
```
unleashed: you can join here literals in a single string: `"</id><key>"`, `"</key><" as it will avoid a couple of string concatenation ops.

lib/3scale/backend/transactor/status.rb
```diff
@@ -211,11 +194,35 @@ def with_report_and_hierarchy(reports)
...
+          xml << '</application>'.freeze
+        end
+
+        def add_user_section(xml)
+          if !@user.nil?
```
unleashed: have you considered moving the check to the caller?

lib/3scale/backend/transactor/status.rb
```diff
@@ -211,11 +194,35 @@ def with_report_and_hierarchy(reports)
...
+            xml << '</reason>'.freeze
+          end
+        end
+
+        def add_application_section(xml)
+          redirect_uri = application.redirect_url
+          xml << "<application><id>".freeze << "#{application.id}" << "</id><key>".freeze
+          xml << "#{application.keys.first}" << "</key><".freeze
+          xml << "#{@redirect_uri_field}>#{redirect_uri}</#{@redirect_uri_field}"
+          xml << '></application>'.freeze
```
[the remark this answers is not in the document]
miguelsorianod: If I do that then this literals will not be frozen even if I put a .freeze, which we were trying to do for the literal parts as mikz commented.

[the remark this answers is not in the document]
mikz: @miguelsorianod but there is no good reason to have extra frozen string when you are allocating it moments before anyway

```ruby
RubyVM::InstructionSequence.compile(<<'XML').disasm
xml << "#{@redirect_uri_field}>#{redirect_uri}</#{@redirect_uri_field}";
xml << '></application>'.freeze
XML
```

```
== disasm: #<ISeq:<compiled>@<compiled>>================================
0000 trace            1                                               (   1)
0002 putself
0003 opt_send_without_block <callinfo!mid:xml, argc:0, FCALL|VCALL|ARGS_SIMPLE>, <callcache>
0006 getinstancevariable :@redirect_uri_field, <is:0>
0009 tostring
0010 putobject        ">"
0012 putself
0013 opt_send_without_block <callinfo!mid:redirect_uri, argc:0, FCALL|VCALL|ARGS_SIMPLE>, <callcache>
0016 tostring
0017 putobject        "</"
0019 getinstancevariable :@redirect_uri_field, <is:0>
0022 tostring
0023 concatstrings    5
0025 opt_ltlt         <callinfo!mid:<<, argc:1, ARGS_SIMPLE>, <callcache>
0028 pop
0029 trace            1                                               (   2)
0031 putself
0032 opt_send_without_block <callinfo!mid:xml, argc:0, FCALL|VCALL|ARGS_SIMPLE>, <callcache>
0035 opt_str_freeze   "></application>"
0037 opt_ltlt         <callinfo!mid:<<, argc:1, ARGS_SIMPLE>, <callcache>
0040 leave
```

versus 

```ruby
RubyVM::InstructionSequence.compile(<<'XML').disasm
xml << "#{@redirect_uri_field}>#{redirect_uri}</#{@redirect_uri_field}></application>"
XML
```

```
== disasm: #<ISeq:<compiled>@<compiled>>================================
0000 trace            1                                               (   1)
0002 putself
0003 opt_send_without_block <callinfo!mid:xml, argc:0, FCALL|VCALL|ARGS_SIMPLE>, <callcache>
0006 getinstancevariable :@redirect_uri_field, <is:0>
0009 tostring
0010 putobject        ">"
0012 putself
0013 opt_send_without_block <callinfo!mid:redirect_uri, argc:0, FCALL|VCALL|ARGS_SIMPLE>, <callcache>
0016 tostring
0017 putobject        "</"
0019 getinstancevariable :@redirect_uri_field, <is:0>
0022 tostring
0023 putobject        "></application>"
0025 concatstrings    6
0027 opt_ltlt         <callinfo!mid:<<, argc:1, ARGS_SIMPLE>, <callcache>
0030 leave
```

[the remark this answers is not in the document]
unleashed: in other words: there is one allocation already going on there with the interpolation, and freezing and adding the frozen strings adds two extra operations, whereas those strings could already be part of the allocated string in the first place.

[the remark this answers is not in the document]
miguelsorianod: I understand that an allocation is less costly that an allocation + a concatenation in that specific case. My doubt was about what happened if that same literal would appear in other places of the code.
The solution then would be to freeze the '></application>' part in the other part of the code. It would not make sense to freeze it in this specific section of the code because anyway you will be doing an allocation for the redirect part. I'll change it now, thank you!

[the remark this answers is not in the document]
unleashed: The thing is what are you willing to optimize for. If you have the same literals sprinkled elsewhere you can make all of them frozen, only some, or none. With strings as interpreted in MRI, using `.freeze` makes the effect of interning them, which is to say they become similar to static, immutable constants. In that case, when the interpreter finds references to the same string (and also with `.freeze`*), it will refer to the interned one to save on unneeded allocations and extra copies.

But of course, having a few copies of the same string (or more accurately, substring) to save in execution speed is usually a good tradeoff. So, in some cases you can intern them, but it only makes sense if you actually save the allocation, and in that case, it's worse if it also adds an extra cost in execution time.

`*`: Ruby 2.3+ have flags for making the `String#freeze` behaviour the default with literals.

lib/3scale/backend/transactor/status.rb
```diff
@@ -200,31 +200,27 @@ def add_plan_section(xml, tag, plan_name)
...
+            xml << '<authorized>false</authorized><reason>'.freeze
+            xml << rejection_reason_text
+            xml << '</reason>'.freeze
           end
         end
 
         def add_application_section(xml)
           redirect_uri = application.redirect_url
-          xml << '<application>' \
+          xml << "<application>" \
```
mikz: I'd say this was correctly using single quotes. There is no interpolation within them so it signals to the reader: don't worry, this is the actual string, no interpolation there.

miguelsorianod: :+1: changed.

lib/3scale/backend/transactor/status.rb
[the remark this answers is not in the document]
unleashed: This is 93% the same comment as above! I want my 10s back!!

unleashed:
bors r=@eguzki,@davidor,@mikz,@unleashed

unleashed:
bors r=@eguzki,@davidor,@mikz,@unleashed

unleashed:
last try, otherwise it is clear that some powerful, higher order entity does not want this merged, and if so better not to annoy them (a la tercera va la vencida)

bors r=@eguzki,@davidor,@mikz,@unleashed

PR #29, opened 2018-05-30
failed_jobs_scheduler: Use Redis Resque instead of Storage for distributed lock

miguelsorianod:
Currently the Redis Distributed Lock is set on the Redis Storage
when the failed jobs scheduler is executed. However, this not allows
to have multiple failed jobs schedulers in execution at the
same time pointing to different Resque servers. Moving the distributed
lock to the Resque server itself solves this problem.

### DIFF
--- a/lib/3scale/backend/failed_jobs_scheduler.rb
+++ b/lib/3scale/backend/failed_jobs_scheduler.rb
@@ -66,7 +66,7 @@ module ThreeScale
 
         def dist_lock
           @dist_lock ||= DistributedLock.new(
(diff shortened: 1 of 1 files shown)


unleashed:
@miguelsorianod but you also need to send jobs to the other Resque server - how do you plan to do that?

miguelsorianod:
We will have a set of listener/workers pointing to one Resque server, and another set pointing to the other Resque server. In the first set we will have the failed jobs scheduler enabled in two workers (for redundancy), that will process the failed queues of the first Resque, and in the second set we will have the failed jobs scheduler enabled too, that will process the failed queues of the second Resque.

PR #32, opened 2018-06-11
Fix race condition in running the race condition test in EventStorage

### LINKED ISSUE #31: Rewrite or remove event_storage spec that keeps marking builds as failed
This spec failure:
```
Failures:

  1) ThreeScale::Backend::EventStorage.ping_if_not_empty with events in set with two calls in same moment (race condition) returns false the second time
     Failure/Error: values = threads.each(&:wakeup).map { |thread| thread.join.value }
     
     fatal:
       No live threads left. Deadlock?
     # ./spec/unit/event_storage_spec.rb:224:in `join'
     # ./spec/unit/event_storage_spec.rb:224:in `block (6 levels) in <module:Backend>'
     # ./spec/unit/event_storage_spec.rb:224:in `map'
     # ./spec/unit/event_storage_spec.rb:224:in `block (5 levels) in <module:Backend>'

Finished in 2.77 seconds (files took 2.09 seconds to load)
938 examples, 1 failure
```
is much more likely to be triggered by the Circle CI infrastructure. In fact, is the most likely reason tests fail, and it is becoming so annoying I am seriously considering removing the whole thing.

unleashed:
There was a race condition in a spec trying to test a race condition. So meta. :)

Closes #31. Hopefully deals the fatal blow to the failing builds, which has been nagging us for too long.

Issues fixed:

  - "wakeup" on a thread needs it to be alive.
  - "sleep" status can also mean I/O blocked, not just stopped.
  - Thread.stop was being called to sync at the end, which is
    useless. It is now used to give a better chance at parallelism
    by sync'ing before starting the real workload of each thread.
  - The Rspec stubbing is unnecessary.

### DIFF
--- a/spec/unit/event_storage_spec.rb
+++ b/spec/unit/event_storage_spec.rb
@@ -206,24 +206,21 @@ module ThreeScale
             end
           end
 
-          context 'with two calls in same moment (race condition)' do
-            it 'returns false the second time' do
-              original_method = EventStorage.method(:pending_ping?)
-              allow(EventStorage).to receive(:pending_ping?) do
-                original_method.call.tap { Thread.stop }
+          context 'with multiple calls at the same moment (race condition)' do
+            it 'returns falsey except for one successful case' do
+              num_threads = 4
+
+              threads = num_threads.times.map do
+                Thread.new { Thread.stop; EventStorage.ping_if_not_empty }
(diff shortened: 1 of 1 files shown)


unleashed:
bors r=@davidor,@eguzki

PR #38, opened 2018-06-22
Implement Service's Users removal

miguelsorianod:
Provide a method to delete all Users of a given Service.

Also provide the HTTP method 'DELETE /services/:service_id/users/all' via Internal API in order to provide this to apisonator administrators.

The decision to NOT delete the Users of the Service when removing a Service is intentional. This has been done to follow coherence like with other entities that are accessible and manageable via Internal API, like with Application, Metrics, Application Keys, etc...

The removal of the Users of the Service has been done taking into account that the number of users of a Service can be big and a batching strategy has been followed.

### DIFF
--- a/app/api/internal/users.rb
+++ b/app/api/internal/users.rb
@@ -44,6 +44,16 @@ module ThreeScale
             [400, headers, { status: :error, error: e.message }.to_json]
           end
         end
+
+        delete '' do |service_id|
+          begin
+            User.delete_all(service_id)
+            { status: :deleted}.to_json
+          rescue => e
+            [400, headers, { status: :error, error: e.message }.to_json]
+          end
+        end
+
       end
     end
   end
--- a/lib/3scale/backend/service.rb
+++ b/lib/3scale/backend/service.rb
@@ -257,7 +257,6 @@ module ThreeScale
 
       def delete_attributes
         keys = ATTRIBUTES.map { |attr| storage_key(attr) }
-        keys << storage_key(:user_set)
         keys << storage_key_by_provider(:id) if default_service?
         storage.del keys
       end
--- a/lib/3scale/backend/storage.rb
+++ b/lib/3scale/backend/storage.rb
@@ -12,6 +12,12 @@ module ThreeScale
       UnspecifiedURIScheme = Class.new Error
       private_constant :UnspecifiedURIScheme
 
+      # Constant used for when batching of operations
+      # is desired/needed. Batching is performed when a lot
+      # of storage operations are need to be performed and we
+      # want to minimize database blocking of other clients
+      BATCH_SIZE = 400
+
       class UnspecifiedURI < Error
         def initialize
           super "Redis URL not specified with url, " \
--- a/lib/3scale/backend/user.rb
+++ b/lib/3scale/backend/user.rb
@@ -77,9 +77,7 @@ module ThreeScale
       def self.delete!(service_id, username)
         service = Service.load_by_id(service_id)
         raise UserRequiresValidService if service.nil?
-        ServiceUserManagementUseCase.new(service, username).delete
-        clear_cache(service_id, username)
-        storage.del(self.key(service_id, username))
+        delete_users_in_batch(service_id, [username])
       end
 
       def self.delete_all(service_id)
@@ -163,15 +161,14 @@ module ThreeScale
 
       def self.delete_users_in_batch(service_id, usernames)
         if usernames.size != 0
-          usernames.each do |username|
-            clear_cache(service_id, username)
-          end
           service_usernames_keys = usernames.map { |username| self.key(service_id, username) }
           storage.pipelined do
             storage.del(service_usernames_keys)
             storage.srem(self.service_users_set_key(service_id), usernames)
           end
-
+          usernames.each do |username|
+            clear_cache(service_id, username)
+          end
         end
       end
 
--- a/lib/3scale/backend/user.rb
+++ b/lib/3scale/backend/user.rb
@@ -82,8 +82,27 @@ module ThreeScale
         storage.del(self.key(service_id, username))
       end
 
+      def self.delete_all(service_id)
+        service = Service.load_by_id(service_id)
+        raise UserRequiresValidService if service.nil?
+        current_scan_cursor_position = "0"
+        finished = false
+        users_set_key = service_users_set_key(service_id)
+        while !finished
+          (current_scan_cursor_position, current_usernames) =
+            storage.sscan(users_set_key, current_scan_cursor_position,
+                          { count: Storage::BATCH_SIZE })
+          delete_users_in_batch(service_id,  current_usernames)
+          finished = current_scan_cursor_position == "0"
+        end
+      end
+
       def self.key(service_id, username)
         "service:#{service_id}/user:#{username}"
+      end
+
+      def self.service_users_set_key(service_id)
+        "service/id:#{service_id}/user_set"
       end
 
       def to_hash
@@ -140,6 +159,20 @@ module ThreeScale
       def self.clear_cache(service_id, user_id)
         key = Memoizer.build_key(self, :load_or_create!, service_id, user_id)
         Memoizer.clear key
+      end
+
+      def self.delete_users_in_batch(service_id, usernames)
+        if usernames.size != 0
+          usernames.each do |username|
+            clear_cache(service_id, username)
+          end
+          service_usernames_keys = usernames.map { |username| self.key(service_id, username) }
+          storage.pipelined do
+            storage.del(service_usernames_keys)
+            storage.srem(self.service_users_set_key(service_id), usernames)
+          end
+
+        end
       end
 
       def self.validate_attributes(attributes)
--- a/spec/acceptance/api/internal/users_api_spec.rb
+++ b/spec/acceptance/api/internal/users_api_spec.rb
@@ -144,4 +144,13 @@ resource 'Users (prefix: /services/:service_id/users)' do
     end
   end
 
+  delete '/services/:service_id/users' do
+    parameter :service_id, 'Service ID', required: true
+
+    example_request 'Deleting all Users of a Service' do
+      expect(status).to eq 200
+      expect(response_json['status']).to eq 'deleted'
+      expect(ThreeScale::Backend::User.exists?(service_id, username)).to be false
+    end
+  end
 end
--- a/test/unit/user_test.rb
+++ b/test/unit/user_test.rb
@@ -119,4 +119,32 @@ class UserTest < Test::Unit::TestCase
 
     assert_equal metrics, user.load_metric_names
   end
+
+  test '.delete_all removes all the Users of a Service' do
+    Service.save!(id: 7003, provider_key: 'test_provkey', default_service: false)
+    Service.save!(id: 7004, provider_key: 'test_provkey', default_service: false)
+    num_users = Storage::BATCH_SIZE + 1
+    num_users.times do |i|
+      User.save! username: "username#{i+1}", service_id: 7003, plan_id: '1001',
+                 plan_name: 'planname'
+    end
+    User.save! username: "username_differentservice", service_id: 7004,
+               plan_id: '1001',plan_name: 'planname'
+
+    User.delete_all(7003)
+
+    num_users.times do |i|
+      assert_equal User.exists?(7003, "username#{i+1}"), false
+    end
+    assert_equal User.exists?(7004, "username_differentservice"), true
+  end
+
+  test '.delete_all does not crash when a Service does not have any user' do
+    Service.save!(id: 7003, provider_key: 'test_provkey', default_service: false)
+
+    assert_nothing_raised do
+      User.delete_all(7003)
+    end
+
+  end
 end


lib/3scale/backend/service.rb
```diff
@@ -241,7 +241,6 @@ def state=(value)
 
       def delete_attributes
         keys = ATTRIBUTES.map { |attr| storage_key(attr) }
-        keys << storage_key(:user_set)
```
miguelsorianod: The responsability of removing the members of the Service's Users set has been intentionally moved to User

lib/3scale/backend/user.rb
```diff
@@ -82,10 +82,29 @@ def self.delete!(service_id, username)
...
+      def self.delete_all(service_id)
+        service = Service.load_by_id(service_id)
+        raise UserRequiresValidService if service.nil?
+        current_scan_cursor_position = "0"
+        finished = false
+        users_set_key = service_users_set_key(service_id)
+        while !finished
+          (current_scan_cursor_position, current_usernames) =
+            storage.sscan(users_set_key, current_scan_cursor_position,
+                          { count: PIPELINED_SLICE_SIZE })
```
miguelsorianod: Here I use the global constant named PIPELINED_SLICE_SIZE, that was already defined, but the name might be a little misleading because we are not really doing pipelining. Do you think we should create a different constant in the user.rb file to not cause confusion?

app/api/internal/users.rb
```diff
@@ -44,6 +44,16 @@ def self.save(service_id, username, usr_attrs, method, headers)
             [400, headers, { status: :error, error: e.message }.to_json]
           end
         end
+
+        delete '/all' do |service_id|
+          begin
+            User.delete_all(service_id)
+            { status: :deleted}
```
eguzki: .to_json ??

lib/3scale/backend/user.rb
eguzki: I do

lib/3scale/backend/user.rb
```diff
@@ -82,10 +82,29 @@ def self.delete!(service_id, username)
...
+          delete_users_in_batch(service_id,  current_usernames)
+          finished = current_scan_cursor_position == "0"
```
eguzki: can you assure that finished will be eventually ```true``` , no matter what? on redis connection error, exception will rise, right?

eguzki:
missing removal of user hash with attributes

lib/3scale/backend/user.rb
unleashed: do not create it in user.rb, since it has nothing to do with it - create a new one with comments on why is it needed in the storage module

app/api/internal/users.rb
miguelsorianod: :+1:

lib/3scale/backend/user.rb
miguelsorianod: :+1:

lib/3scale/backend/user.rb
```diff
@@ -142,6 +161,17 @@ def self.clear_cache(service_id, user_id)
         Memoizer.clear key
       end
 
+      def self.delete_users_in_batch(service_id, usernames)
```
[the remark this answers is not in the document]
miguelsorianod: Seems a good idea :+1: I'm going to implement it

miguelsorianod:
I've checked all your commits and answered them, implementing some of the requested changes. Could you review them?

Thank you.

test/unit/user_test.rb
```diff
@@ -119,4 +119,27 @@ def test_create_user_successful_service_not_require_registered_users
...
+    User.save! username: "username_differentservice", service_id: 7004,
+               plan_id: '1001',plan_name: 'planname'
+
+    User.delete_all(7003)
+
+    for i in 1..num_users
+      assert_equal User.exists?(7003, "username#{i}"), false
+    end
+    assert_equal User.exists?(7004, "username_differentservice"), true
+    User.delete!(7004, "username_differentservice")
```
eguzki: clean up should not be part of the test

miguelsorianod: The cleanup is there because I did not want to leave the user in the DB when the test ends

eguzki:
I wanted to rise a discussion about the design decision made here that might not be right. 

This implementation is deleting potentially big sets of data synchronously from the API endpoint call? 

Delete job could be better option?

Martouta:
Please also remember that when this is done, the Pisoni Gem should have this endpoint too 😄

lib/3scale/backend/user.rb
```diff
@@ -77,15 +77,32 @@ def self.save!(attributes)
...
+
+      def self.delete_all(service_id)
+        service = Service.load_by_id(service_id)
+        raise UserRequiresValidService if service.nil?
+        current_scan_cursor_position = "0"
+        finished = false
+        users_set_key = service_users_set_key(service_id)
+        while !finished
+          (current_scan_cursor_position, current_usernames) =
+            storage.sscan(users_set_key, current_scan_cursor_position,
```
eguzki: Have you considered using `scard` and `spop` operations to delete all elements of the set? At least seems less risky than using cursors on potentially big data sets.

eguzki:
This feature should be merge. 

Anything else to be added/updated? 

Looks quite good to me

lib/3scale/backend/service.rb
eguzki: Is this key ever removed when deleting all users from a service?

miguelsorianod: Yes. When all elements of a set are removed then the set is automatically removed in Redis.

lib/3scale/backend/user.rb
miguelsorianod: When this was implemented, we were using Redis server 2.8 in SaaS. The spop command does not support specifying the number of elements to delete until Redis 3.2 (https://redis.io/commands/spop) so we could not use this command and that was the reason we opted to use cursor-based deletion.
However, recently we have upgraded our Redis server version so we could replace the cursor to use scard and spop.
If we do this we would need to update the apisonator development Docker images to use a newer version of Redis too because right now it is still using 2.8 and would stop working.
Right now we have the following Redis versions:
SaaS: Redis 4.X
On-Prem: Redis 3.2

If we upgrade Redis server version in the development images I'd say to use Redis 3.2 (the lower version). This would mean that we would not be using in dev the same version that in SaaS but we would be testing a version where all commands should be compatible in both versions. What do you think?

lib/3scale/backend/user.rb
davidor: Methods defined under `private` with `self.f` are not private in Ruby. Just one of those things that might be confusing at first :)

Private class methods need to be defined under `private` inside `class << self` or using https://ruby-doc.org/core-2.5.3/Module.html#method-i-private_class_method

No need to change this now, as I see that there are other methods in this class that are supposed to be private but are not.

lib/3scale/backend/user.rb
davidor: @miguelsorianod I think we should update the development images as you suggested so we can change this to use spop :+1:

test/unit/user_test.rb
```diff
@@ -119,4 +119,27 @@ def test_create_user_successful_service_not_require_registered_users
 
     assert_equal metrics, user.load_metric_names
   end
+
+  test '.delete_all removes all the Users of a Service' do
+    Service.save!(id: 7003, provider_key: 'test_provkey', default_service: false)
+    Service.save!(id: 7004, provider_key: 'test_provkey', default_service: false)
+    num_users = 600
```
davidor: 600 looks like an arbitrary number. I guess you decided this to be larger than the number of users to be deleted on each batch? If so, I think it would be better for readability to relate this variable with that constant, something like `num_users = BATCH_SIZE + 1`.
What do you think?

miguelsorianod: :+1: . It was indeed to be larger than the number of users to be deleted on each batch

test/unit/user_test.rb
```diff
@@ -119,4 +119,27 @@ def test_create_user_successful_service_not_require_registered_users
 
     assert_equal metrics, user.load_metric_names
   end
+
+  test '.delete_all removes all the Users of a Service' do
```
davidor: I think we should also test the case of deleting the users of a service that does not have any. Just to make sure that nothing crashes.

test/unit/user_test.rb
```diff
@@ -119,4 +119,27 @@ def test_create_user_successful_service_not_require_registered_users
...
+    for i in 1..num_users
+      User.save! username: "username#{i}", service_id: 7003, plan_id: '1001',
+                 plan_name: 'planname'
+    end
+    User.save! username: "username_differentservice", service_id: 7004,
+               plan_id: '1001',plan_name: 'planname'
+
+    User.delete_all(7003)
+
+    for i in 1..num_users
```
davidor: Nitpick: In Ruby `(1..num_users).each { |n| ... }` is more idiomatic.

test/unit/user_test.rb
davidor: There's a `setup` method above that cleans the DB.
That method runs before each test, so there's no need to clean up on every individual one.

spec/acceptance/api/internal/users_api_spec.rb
```diff
@@ -144,4 +144,14 @@
     end
   end
 
+  delete '/services/:service_id/users/all' do
```
davidor: I think there's no need for the `/all` part.

lib/3scale/backend/storage.rb
```diff
@@ -12,6 +12,12 @@ class Storage
       UnspecifiedURIScheme = Class.new Error
       private_constant :UnspecifiedURIScheme
 
+      # Constant used for when batching of operations
+      # is desired/needed. Batching is performed when a lot
+      # of storage operations are need to be performed and we
+      # want to minimize database blocking of other clients
+      BATCH_SIZE = 400
```
davidor: We already have a constant related to this one in a different module: https://github.com/3scale/apisonator/blob/baa78aec5bb1c27f0e145ea20de0fa7893c5a5b5/lib/3scale/backend/constants.rb#L6

I think `BATCH_SIZE` should be where you placed it because it's clearly related to `Storage`. I think that's better than having a global constants module. But I also think that it would be clearer if we moved the other constant here. That could be done on a different PR.

spec/acceptance/api/internal/users_api_spec.rb
```diff
@@ -144,4 +144,14 @@
...
+    parameter :service_id, 'Service ID', required: true
+
+    example_request 'Deleting all Users of a Service' do
+      expect(status).to eq 200
+      expect(response_json['status']).to eq 'deleted'
+      expect(ThreeScale::Backend::User.delete_all(service_id)).to be nil
```
davidor: I think that this `expect` assumes things that are not documented nor tested.

In the unit tests of `User` you did not check the return value of `delete_all`, you just verified that after calling it, users no longer exists. I think that's fine, but the here we should just do the same.

The alternative would be to define a return value for `delete_all` in docs/tests, but I don't think it's needed in this case.
What do you think?

test/unit/user_test.rb
eguzki: maybe `1.upto(num_users).each { |n| ... }` ?? just another option, valid as well.

app/api/internal/users.rb
```diff
@@ -44,6 +44,16 @@ def self.save(service_id, username, usr_attrs, method, headers)
             [400, headers, { status: :error, error: e.message }.to_json]
           end
         end
+
+        delete '/all' do |service_id|
```
eguzki: as @davidor suggested, maybe 

`DELETE /internal/services/:service_id/users` would be more RESTful

can be implemented by

`delete '' do |service_id|`

test/unit/user_test.rb
mikz: If say that the only allowed option here is #times 😆

lib/3scale/backend/user.rb
miguelsorianod: :+1: . I've done this in https://github.com/3scale/apisonator/pull/69

app/api/internal/users.rb
miguelsorianod: :+1:

spec/acceptance/api/internal/users_api_spec.rb
miguelsorianod: :+1:

test/unit/user_test.rb
miguelsorianod: :+1:

test/unit/user_test.rb
miguelsorianod: :+1:

lib/3scale/backend/user.rb
miguelsorianod: I upgraded the dev images and we are now using Redis 3.2.X for our dev images. However, the Redis ruby client version that we use in some of our services is 3.2.2, but the support for spop with the number of elements to pop was added to 3.3.2. At this moment we cannot upgrade to that version in those services due to a bug we found with reconnections https://github.com/redis/redis-rb/issues/668. We'll continue investigating on that and for the moment to be able to proceed we'll leave this with scan.

test/unit/user_test.rb
eguzki: but `#times` is 0-based. If index is needed and 1-based as its first implementation, I'd say it is better

`1.upto(num_users).each { |n| ... }`

spec/acceptance/api/internal/users_api_spec.rb
miguelsorianod: I now check if the user has been deleted or not instead of the return value of the delete_all method

miguelsorianod:
I made some changes and I think could be merged.
How do you see it?

app/api/internal/users.rb
mikz: What happens if I have a following URL constructing code:

```ruby
http.delete("/internal/services/#{service_id}/users/#{username}")
```

and `username` is nil? So the resulting URL is `/internal/services/:service_id/users/` ? Would it delete all users?

miguelsorianod: It shouldn't be a problem.

If you call `/internal/services/:service_id/users/` then it would return a 404 not found because that URL has not been defined in the internal API.

For example, if I change the current internal api test code to execute this code (notice the slash at the end):
```
  delete '/services/:service_id/users/' do
    parameter :service_id, 'Service ID', required: true

    example_request 'Deleting all Users of a Service' do
      expect(status).to eq 200
      expect(response_json['status']).to eq 'deleted'
      expect(ThreeScale::Backend::User.exists?(service_id, username)).to be false
    end
  end
```

and I keep the internal api code as:
```
        delete '' do |service_id|
          begin
            User.delete_all(service_id)
            { status: :deleted}.to_json
          rescue => e
            [400, headers, { status: :error, error: e.message }.to_json]
          end
        end
```

When I execute the test I get the following:
```
Failures:

  1) Users (prefix: /services/:service_id/users) DELETE /services/:service_id/users/ Deleting all Users of a Service
     Failure/Error: expect(status).to eq 200
     
       expected: 200
            got: 404
     
       (compared using ==)
     # ./spec/acceptance/api/internal/users_api_spec.rb:151:in `block (3 levels) in <top (required)>'

Finished in 0.90085 seconds (files took 0.68985 seconds to load)
144 examples, 1 failure

Failed examples:

rspec ./spec/acceptance/api/internal/users_api_spec.rb:150 # Users (prefix: /services/:service_id/users) DELETE /services/:service_id/users/ Deleting all Users of a Service
```

mikz: 👍

lib/3scale/backend/user.rb
```diff
@@ -142,6 +159,17 @@ def self.clear_cache(service_id, user_id)
         Memoizer.clear key
       end
 
+      def self.delete_users_in_batch(service_id, usernames)
+        if usernames.size != 0
+          service_usernames_keys = usernames.map { |username| self.key(service_id, username) }
+          storage.del(service_usernames_keys)
```
davidor: I think these 2 should be grouped in a pipeline:
```ruby
          storage.pipelined do
            storage.del(service_usernames_keys)
            storage.srem(self.service_users_set_key(service_id), usernames)
          end
```

Sorry, I didn't see this in my first review. All the rest looks good to me :+1:

miguelsorianod: Changed

eguzki:
bors r=@davidor,@eguzki

PR #39, opened 2018-06-25
service: Fix set a Service as active when no entry in the DB is found

miguelsorianod:
This is a bugfix commit.

This comes from the change introduced in https://github.com/3scale/apisonator/pull/35

Currently when a Service existed in the
database but had no 'state' entry then was marked as :suspended.
That breaks compatibility with previously existing Services.
With this commit existing Services in the database that have no
'state' entry in the database are marked as :active to ensure
compatibility.

With this changes the following situations will happen:

 * If a Service is loaded from the database and the database does not have the state entry, then it is loaded as an active Service
 * If a Service is loaded from the database and the database has a state entry but the value of the entry is not active then it is marked as suspended
* If a Service is created as an object directly (with the initialize method) and does not have the state attribute in the attributes hash that is received as a parameter then it is marked as active
* **If a Service is created as an object directly (with the initialize method), does have the state attribute and it is nil then the active state is assigned to it.**
* If a Service is created as an object directly (with the initialize method), does have the state attribute and it is a non-nil value that is not active then it is marked as suspended.
* If a Service is created as an object directly (with the initialize method), and does have the state attribute and it is defined as active then it is marked as active
* If the 'state=' instance method is called with a nil or active value then the Service is marked as active. Otherwise is marked as suspended

On the previous points an important situation happens: An input state attribute that is nil is considered as active too. This has been done to follow coherence with the current situation of the Services, where not having a state set to a Service means that it is active.

### DIFF
--- a/lib/3scale/backend/service.rb
+++ b/lib/3scale/backend/service.rb
@@ -238,6 +238,13 @@ module ThreeScale
       def active?
         state == :active
       end
+      alias_method :active, :active?
+
+      def active=(value)
+        self.state = value ? :active : :suspended
+      end
+
+      private
 
       def state=(value)
         # only :active or nil will be considered as :active
@@ -246,8 +253,6 @@ module ThreeScale
         # this change
         @state = value.nil? || value.to_sym == :active ? :active : :suspended
       end
-
-      private
 
       def delete_attributes
         keys = ATTRIBUTES.map { |attr| storage_key(attr) }
--- a/lib/3scale/backend/service.rb
+++ b/lib/3scale/backend/service.rb
@@ -176,8 +176,15 @@ module ThreeScale
       end
 
       def initialize(attributes = {})
-        # default state
-        @state = :active
+        # :state is set as active in this method when:
+        # - The state key is not present in the attributes hash
+        # - The state key is present in the attributes hash but it has
+        #   the nil value
+        # This is done in order to not break compatibility for existing
+        # Services saved in the database, that do not contain the state
+        # key.
+        attributes[:state] ||= :active
+
         super(attributes)
       end
 
@@ -233,8 +240,11 @@ module ThreeScale
       end
 
       def state=(value)
-        # anything but :active will be suspended
-        @state = value.nil? || value.to_sym != :active ? :suspended : :active
+        # only :active or nil will be considered as :active
+        # we assume nil is active because not having a state in an
+        # existing service means that is active in Services created before
+        # this change
+        @state = value.nil? || value.to_sym == :active ? :active : :suspended
       end
 
       private
--- a/spec/acceptance/api/internal/services_api_spec.rb
+++ b/spec/acceptance/api/internal/services_api_spec.rb
@@ -55,7 +55,7 @@ resource 'Services (prefix: /services)' do
         example_request 'Get Service by ID' do
           expect(status).to eq 200
           expect(response_json['service']['id']).to eq id
-          expect(response_json['service']['state']).to eq 'suspended'
+          expect(response_json['service']['state']).to eq 'active'
         end
       end
       context 'when state is suspended' do
--- a/spec/acceptance/api/internal/services_api_spec.rb
+++ b/spec/acceptance/api/internal/services_api_spec.rb
@@ -67,26 +67,13 @@ resource 'Services (prefix: /services)' do
           expect(response_json['service']['state']).to eq state.to_s
         end
       end
-      context 'with an service that has invalid state in db' do
+      context 'when state is set to an invalid state' do
+        let(:state) { :not_valid_state }
         let(:id) { otherid }
-        let(:storage) { ThreeScale::Backend::Storage.instance }
-        example 'Get Service by ID should return inactive state' do
-          storage.set ThreeScale::Backend::Service.storage_key(id, 'state'), 'invalid_state'
-          do_request(id: id)
+        example_request 'Get Service by ID' do
           expect(status).to eq 200
           expect(response_json['service']['id']).to eq id
           expect(response_json['service']['state']).to eq 'suspended'
-        end
-      end
-      context 'with an service that has no state in db' do
-        let(:id) { otherid }
-        let(:storage) { ThreeScale::Backend::Storage.instance }
-        example 'Get Service by ID should return inactive state' do
-          storage.del ThreeScale::Backend::Service.storage_key(id, 'state')
-          do_request(id: id)
-          expect(status).to eq 200
-          expect(response_json['service']['id']).to eq id
-          expect(response_json['service']['state']).to eq 'active'
         end
       end
     end
--- a/spec/acceptance/api/internal/services_api_spec.rb
+++ b/spec/acceptance/api/internal/services_api_spec.rb
@@ -78,6 +78,17 @@ resource 'Services (prefix: /services)' do
           expect(response_json['service']['state']).to eq 'suspended'
         end
       end
+      context 'with an service that has no state in db' do
+        let(:id) { otherid }
+        let(:storage) { ThreeScale::Backend::Storage.instance }
+        example 'Get Service by ID should return inactive state' do
+          storage.del ThreeScale::Backend::Service.storage_key(id, 'state')
+          do_request(id: id)
+          expect(status).to eq 200
+          expect(response_json['service']['id']).to eq id
+          expect(response_json['service']['state']).to eq 'active'
+        end
+      end
     end
   end
 
--- a/spec/unit/service_spec.rb
+++ b/spec/unit/service_spec.rb
@@ -440,7 +440,9 @@ module ThreeScale
             { state: :active, id: '8001' },
             { state: 'active', id: '8001' },
             # even when state is not set
-            { id: '8001' }
+            { id: '8001' },
+            # even when state is intentionally set as nil
+            { state: nil, id: '9001'}
           ].each do |svc_attrs|
             expect(Service.save!(svc_attrs).active?).to be_truthy
           end
@@ -460,6 +462,22 @@ module ThreeScale
           ].each do |svc_attrs|
             expect(Service.save!(svc_attrs).active?).to be_falsy
           end
+        end
+
+        it 'returns true when the service does not have state in the DB' do
+          service_id = '9001'
+          Service.save!({state: :suspended, id: service_id, default_service: false})
+          Service.storage.del ThreeScale::Backend::Service.storage_key(service_id, 'state')
+          expect(Service.load_by_id(service_id).active?).to be true
+          Service.delete_by_id(service_id)
+        end
+
+        it 'returns false when the service has an invalid state in the DB' do
+          service_id = '9001'
+          Service.save!({state: :not_defined_state, id: service_id, default_service: false})
+          Service.storage.set ThreeScale::Backend::Service.storage_key(service_id, 'state'), 'not_defined_state'
+          expect(Service.load_by_id(service_id).active?).to be false
+          Service.delete_by_id(service_id)
         end
       end
     end
--- a/spec/unit/service_spec.rb
+++ b/spec/unit/service_spec.rb
@@ -480,6 +480,40 @@ module ThreeScale
           Service.delete_by_id(service_id)
         end
       end
+
+      describe '#active=' do
+        it 'when set to true, active? returns true' do
+          [
+            { state: :suspended, id: '9001', default_service: false },
+            { state: :active, id: '9001', default_service: false },
+            { state: :something, id: '9001', default_service: false },
+            { state: nil, id: '9001', default_service: false },
+            { state: "", id: '9001', default_service: false },
+            { id: '9001', default_service: false}
+          ].each do |svc_attrs|
+            service = Service.save!(svc_attrs)
+            service.active = true
+            expect(service.active?).to be true
+            Service.delete_by_id('9001')
+          end
+        end
+
+        it 'when set to false, active? returns false' do
+          [
+            { state: :suspended, id: '9001', default_service: false },
+            { state: :active, id: '9001', default_service: false },
+            { state: :something, id: '9001', default_service: false },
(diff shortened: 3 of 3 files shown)


lib/3scale/backend/service.rb
```diff
@@ -56,6 +56,14 @@ def load_by_id(service_id)
 
           return if service_attrs[:provider_key].nil?
 
+          # If service_attrs is nil then it means no entry has been found
+          # in the storage. In order to not break compatibility with existing
+          # Services, by default if no entry is found in the storage for the
+          # state attribute then the :active state is assumed
+          if service_attrs[:state].nil?
+            service_attrs[:state] = :active
+          end
```
unleashed: `service_attrs[:state] ||= :active`

unleashed: why don't you move this to the `initialize` method?
```ruby
def initialize(attributes = {})
  attributes[:state] ||= :active
  @state = attributes[:state]
  super attributes
end
```

eguzki:
required test in *services_api_spec*
```ruby
context 'with an service that has no state in db' do                     
  let(:id) { otherid }                                                   
  let(:storage) { ThreeScale::Backend::Storage.instance }                
  example 'Get Service by ID should return inactive state' do            
    storage.del ThreeScale::Backend::Service.storage_key(id, 'state')    
    do_request(id: id)                                                   
    expect(status).to eq 200                                             
    expect(response_json['service']['id']).to eq id                      
    expect(response_json['service']['state']).to eq 'active'             
  end                                                                    
end                                                                      

```

lib/3scale/backend/service.rb
miguelsorianod: I did not move that to initialize because we want to set the state to suspended if the attribute does not exist (has not been passed as an attributes key) when the initialize method is directly called. The attributes[:state] ||= :active will set it to active it does not exist.

unleashed: that's why I added the two additional comments :) - let's have it sorted out in a single place, please

miguelsorianod: I assumed that we wanted to differentiate this two cases with different behaviour:
* The case where the state entry does not exists in the database for an existing Service
* The case where a new Service is created via the initialize method and a state with nil value is intentionally set.

As we discussed, it is not possible to know if the second previous point is due to this has been done via initialize or because the service has been loaded via the load_by_id method and Redis simply has returned a nil value (meaning that Redis state key for that Service exists). Because of that we have decided to interpret a nil state or a not existing state as 'active'.

I have made the corresponding changes to achieve this and updated the tests.

miguelsorianod: I have also updated the initial comment of this PR listing the possible situations that can happen

spec/acceptance/api/internal/services_api_spec.rb
```diff
@@ -55,7 +55,7 @@
         example_request 'Get Service by ID' do
           expect(status).to eq 200
           expect(response_json['service']['id']).to eq id
-          expect(response_json['service']['state']).to eq 'suspended'
+          expect(response_json['service']['state']).to eq 'active'
```
unleashed: minor nit: the title is inconsistent with the other commit - spec/services_api vs service_api_spec

spec/acceptance/api/internal/services_api_spec.rb
```diff
@@ -78,6 +78,17 @@
           expect(response_json['service']['state']).to eq 'suspended'
         end
       end
+      context 'with an service that has no state in db' do
+        let(:id) { otherid }
+        let(:storage) { ThreeScale::Backend::Storage.instance }
+        example 'Get Service by ID should return inactive state' do
+          storage.del ThreeScale::Backend::Service.storage_key(id, 'state')
```
unleashed: I think that this kind of test belongs in the unit tests/specs for `Service` - this is an API spec and the storage and its representation are implementation details that should not appear here. I know this escaped my radar with the other specs here, but let's have it moved to the unit tests.

spec/unit/service_spec.rb
```diff
@@ -461,6 +463,60 @@ module Backend
...
+        it 'returns false when the service has an invalid state in the DB' do
+          service_id = '9001'
+          Service.save!({state: :not_defined_state, id: service_id, default_service: false})
+          Service.storage.set ThreeScale::Backend::Service.storage_key(service_id, 'state'), 'not_defined_state'
+          expect(Service.load_by_id(service_id).active?).to be false
+          Service.delete_by_id(service_id)
+        end
+      end
+
+      describe '#state' do
```
eguzki: those tests are repetition of ```#active?``` tests

lib/3scale/backend/service.rb
```diff
@@ -176,8 +176,15 @@ def default_service?(provider_key, id)
...
-        # default state
-        @state = :active
+        # :state is set as active in this method when:
+        # - The state key is not present in the attributes hash
+        # - The state key is present in the attributes hash but it has
+        #   the nil value
+        # This is done in order to not break compatibility for existing
+        # Services saved in the database, that do not contain the state
+        # key.
+        attributes[:state] ||= :active
```
eguzki: this change is not required
```
@state = :active
```
is enough

spec/unit/service_spec.rb
```diff
@@ -461,6 +463,60 @@ module Backend
             expect(Service.save!(svc_attrs).active?).to be_falsy
           end
         end
+
+        it 'returns true when the service does not have state in the DB' do
+          service_id = '9001'
+          Service.save!({state: :suspended, id: service_id, default_service: false})
+          Service.storage.del ThreeScale::Backend::Service.storage_key(service_id, 'state')
+          expect(Service.load_by_id(service_id).active?).to be true
+          Service.delete_by_id(service_id)
```
eguzki: clean up not required here,

in case it was needed, better to use ```after```

spec/acceptance/api/internal/services_api_spec.rb
```diff
@@ -67,12 +67,10 @@
           expect(response_json['service']['state']).to eq state.to_s
         end
       end
-      context 'with an service that has invalid state in db' do
+      context 'when state is set to an invalid state' do
+        let(:state) { :not_valid_state }
         let(:id) { otherid }
-        let(:storage) { ThreeScale::Backend::Storage.instance }
-        example 'Get Service by ID should return inactive state' do
-          storage.set ThreeScale::Backend::Service.storage_key(id, 'state'), 'invalid_state'
```
eguzki: if you remove direct db update, db registry will not have invalid state

spec/unit/service_spec.rb
miguelsorianod: They are not. That's because active can return true/false and that's what is checked there, whereas the state getter tests verify what values should be returned.

spec/acceptance/api/internal/services_api_spec.rb
miguelsorianod: That's not what it is being tested anymore. Now it tests that when setgting an invalid state it returns suspended:
```
      context 'when state is set to an invalid state' do
```

The invalid state in the DB test has been moved to spec/service.rb

spec/unit/service_spec.rb
miguelsorianod: I understand cleanup is required there because we are saving it into the database.

Regarding the after hook, I'll implement it.

lib/3scale/backend/service.rb
miguelsorianod: I agree that it's enough with @state = :active but I think it's better for the readability of the code to leave the attributes[:state] ||= :active  operation because it shows the clear intent.

spec/unit/service_spec.rb
eguzki: service class public interface should be ```#active?```. 

IMO test plan should only test ```#active?```. They should not test ```state``` accessor, it should even be private accessor, nobody should access it externally

spec/unit/service_spec.rb
miguelsorianod: I've pushed new changes with the implementation of the after hook

lib/3scale/backend/service.rb
eguzki: You are checking if value is nil in attribute writer accessor, no need to double check here.

spec/unit/service_spec.rb
miguelsorianod: I thought you wanted it to be public too as written in the original commit because the attribute_reader is set as public and the setter too.

If your idea was to only have the #active? method as the public interface, which I agree, then we could do the following thing if the style is correct (let me know):

```
diff --git a/lib/3scale/backend/service.rb b/lib/3scale/backend/service.rb
index 9c46aad2..a766f2a9 100644
--- a/lib/3scale/backend/service.rb
+++ b/lib/3scale/backend/service.rb
@@ -10,6 +10,8 @@ module ThreeScale
       private_constant :ATTRIBUTES
 
       attr_reader :state
+      private :state
+
       attr_accessor :provider_key, :id, :backend_version,
         :default_user_plan_id, :default_user_plan_name
       attr_writer :referrer_filters_required, :user_registration_required,
@@ -239,6 +241,10 @@ module ThreeScale
         state == :active
       end
 
+
+      private
+
+
       def state=(value)
         # only :active or nil will be considered as :active
         # we assume nil is active because not having a state in an
@@ -247,8 +253,6 @@ module ThreeScale
         @state = value.nil? || value.to_sym == :active ? :active : :suspended
       end
 
-      private
-
       def delete_attributes
         keys = ATTRIBUTES.map { |attr| storage_key(attr) }
         keys << storage_key(:user_set)
```

lib/3scale/backend/service.rb
miguelsorianod: I know, it's just a matter of readability of the code.

spec/acceptance/api/internal/services_api_spec.rb
eguzki: This is end to end acceptance test. ```service_spec.rb``` test scope is more functional, unittest. The purpose of the test was: 'What do I get when I request via API a service and in database state value is corrupted". The unittest in ```service_spec.rb``` does not answer that.

In the way it is now, we know from other test in ```service_spec``` that saving service with invalid state attribute, does not propagate to database and state is set as ```:suspended```, the same situation as the previous tests ```when state is suspended```. There is no new use case being tested as it is now.

miguelsorianod: I think that having knowledge on what happens to the database from the point of view of the API specs should not be tested, although I understand the need of looking at what happens when you try to obtain a service that has no db state record. I tried to  cover the specific case of what happens in the DB in the Service spec, when the Service.load_by_id is called (that it's what is called by the API get method at the end)

spec/unit/service_spec.rb
eguzki: ok for the attr reader.

The attr writer I would leave it open.

If the future features require *state* to be multivalue (not just binary value), we can open attr reader

spec/acceptance/api/internal/services_api_spec.rb
eguzki: There is no way to save service info to database leaving state empty using current version of service class. For the end to end acceptance test, we need to do it manually. I agree with you for functional unittests.

spec/unit/service_spec.rb
miguelsorianod: If we leave the attr writer as open, shouldn't we be able to test in some way how the method should behave?

lib/3scale/backend/service.rb
eguzki: this is not coding style. This is different logic implementation. Currently same result for the current implementation of the attribute writer.

spec/unit/service_spec.rb
eguzki: testing ```#active?```

Maybe we should leave ```state``` as private as you suggested and create new public method
```
def set_active(active)
   state = active ? :active : :suspended
end
```

lib/3scale/backend/service.rb
miguelsorianod: With both options the same result would be returned. I prefer to maintain the readability at this moment

spec/unit/service_spec.rb
miguelsorianod: So, the idea would be to make tests of the set_active method because it would be public, and then in the test test that the call to active? returns true???

eguzki: the whole idea is an abstraction over service state. The public api allows you to check whether a service is active or not using ```#active?``` method. The public api allows you to update service activeness doing 
```
service.set_disabled
service.set_enabled
```
or 
```
service.set_enabled(false)
service.set_enabled(true)
```

Internal *state* attr and database serialization is implementation detail.

miguelsorianod: I propose the following:

Implement the methods:

```
service.enable
service.disable
```

Those methods would call the private setter, and the getter would be private too.
Tests for those new methods would be implemented, and inside those methods we would check the results of the #active? method after calling them.

What do you think?

miguelsorianod: I have implemented the enable and disable methods

spec/acceptance/api/internal/services_api_spec.rb
unleashed: We want to specify the internal API here, not do e2e tests. The non-existent state in storage issue is an implementation + storage migration detail that is covered with the unit test, albeit not an integration test. But an integration test covering this is of no value if the unit tests already prove that the storage value does never leak outside, so it becomes an internal detail that we should not care about at that level.

spec/unit/service_spec.rb
miguelsorianod: Finally I removed the implementation of enable/disable methods because could cause confusion implying that calling them would write to the DB. I created the active=(value) setter instead, having value as a boolean.
Also active has been set as an alias of active?

unleashed:
we are releasing a new version - @eguzki if you want to change anything from this please open a PR

bors r+

PR #41, opened 2018-06-29
queue_storage: allow defining Resque configuration without sentinels

miguelsorianod:
This allows to configure a Redis Resque without defining sentinels.
This is the case for example when Redis ElastiCache is used.

While at it also allow picking up the configuration in dev/test modes.

### DIFF
--- a/lib/3scale/backend/queue_storage.rb
+++ b/lib/3scale/backend/queue_storage.rb
@@ -2,24 +2,14 @@ module ThreeScale
   module Backend
     class QueueStorage
       def self.connection(env, cfg)
-        init_params = {}
-        if !%w(development test).include?(env)
-          raise 'Configuration must have a valid queues section' unless valid_cfg? cfg
-          init_params[:url] = cfg.queues.master_name
-        else
+        init_params = { url: cfg.queues && cfg.queues.master_name }
+        if %w(development test).include?(env)
           init_params[:default_url] = '127.0.0.1:6379'
         end
-
         options = Backend::Storage::Helpers.config_with(cfg.queues,
                                                         options: init_params)
 
         Redis.new options
-      end
-
-      private
-
-      def self.valid_cfg?(cfg)
-        cfg.queues && cfg.queues.master_name && cfg.queues.sentinels
       end
     end
   end
--- a/spec/unit/queue_storage_spec.rb
+++ b/spec/unit/queue_storage_spec.rb
@@ -22,7 +22,7 @@ module ThreeScale
 
           context 'with a invalid configuration' do
             it 'returns an exception' do
-              expect { conn }.to raise_error RuntimeError
+              expect { conn }.to raise_error(StandardError)
             end
           end
 


eguzki:
we should not merge anything in master.

If this is something that has nothing to do with apisonator, update patches and PR to *apisonator* branch.

Eventually we should switch from apisonator to master

miguelsorianod:
About the possibility of using an empty array, that's possible because the cfg_sentinels_handler in storage.rb sees empty as not using sentinels:
https://github.com/3scale/apisonator/blob/d6d4d1048140705201f7fdb3c2287d75c63ef709/lib/3scale/backend/storage.rb#L190

However I'd prefer not having to specify an empty array in the configuration file, which seems ugly to me.

About seeing if the valid_cfg? methods makes sense let me take a look :+1:

miguelsorianod:
Both Resque and Storage end calling the methods in storage.rb but the initialization for resque is a specific one in queue_storage.rb. There, it configures some parameters depending on the introduced configuration, which are different than the one for storage (for example, the 'default_url' redis parameter) , or renames the master_name parameter to 'url', which is indeed a Redis config-related parameter,and then indeed calls some storage helpers to generate a configuration.

I still see it makes sense having this method because it references several sublevels of configuration: cfg.queues.master_name and we would not want to receive a NoMethodError.

unleashed:
I just took a look and we could actually do:
```ruby
def self.connection(env, cfg)
  init_params = {
    url: cfg.queues && cfg.queues.master_name,
    default_url: '127.0.0.1:6379'
  }
# [...]
```
...for removing the branching on env and the different configurations. I was about to tell you not to worry about it, as it's not important, but then I remembered [these comments](https://github.com/3scale/pisoni/blob/e9db2a1b077b43311ee1fe588cc2e0c156862e02/docker/docker-compose.apisonator.yml#L32-L33) which would be nice to fix.

miguelsorianod:
I agree. As we discussed I now pushed a change that allows dev/test environments to specify a non-localhost URL.

Furthermore, we now do not crash the program on the storage_queue.rb file anymore but we let the storage.rb to do the crashing. I've updated the tests accordingly.

lib/3scale/backend/queue_storage.rb
```diff
@@ -2,25 +2,15 @@ module ThreeScale
   module Backend
     class QueueStorage
       def self.connection(env, cfg)
-        init_params = {}
-        if !%w(development test).include?(env)
-          raise 'Configuration must have a valid queues section' unless valid_cfg? cfg
-          init_params[:url] = cfg.queues.master_name
-        else
+        init_params = { url: cfg.queues && cfg.queues.master_name}
```
unleashed: typo space between identifier and `}`?

spec/unit/queue_storage_spec.rb
```diff
@@ -22,7 +22,7 @@ module Backend
 
           context 'with a invalid configuration' do
             it 'returns an exception' do
-              expect { conn }.to raise_error RuntimeError
+              expect { conn }.to raise_error
```
miguelsorianod: This has been changed because the RuntimeError is not raised anymore after the changes that have been done in the program.

We know that it should crash but without specifying any specific error now

lib/3scale/backend/queue_storage.rb
miguelsorianod: fixed

miguelsorianod:
commits have been squashed

miguelsorianod:
bors r=@unleashed,@eguzki

PR #43, opened 2018-07-03
docs/alerts: explain how alerts work and their issues

unleashed:


eguzki:
no link from the main README?

davidor:
I wouldn't link this from the README.

In my opinion, the README should be useful to get started with the project. I would consider including a link to the docs section (if it's not already there) for users who want to learn more advanced topics, like the alerts one.

eguzki:
Maybe not from README. My point was to be linked from somewhere. Not requiring people to open the md file manually.

Maybe from DEVELOPMENT.md or some other place.

docs/alerts.md
```diff
@@ -0,0 +1,40 @@
...
+the following utilization percentage of an application _if configured to do so_:
+
+0, 50, 80, 90, 100, 120, 150, 200, 300
+
+However, only one alert percentage is to be notified in any given 24h period.
+
+### Alert notification
+
+Alerts are only sent **once** in any given 24 hours period for each application and
+utilization percentage regardless of the period for which the utilization
```
davidor: We should also mention metrics.

When an alert has been sent for a given app and utilization %, it won't be sent again when going over limits for a different metric.

unleashed: @davidor does the change in the commit below make it all read ok to you?

davidor: :+1:

unleashed:
Ok, linked to the [`docs`](https://github.com/3scale/apisonator/tree/master/docs) folder from the [README.md](https://github.com/3scale/apisonator/blob/82d12f86f23b5371a0b0d636c3daf44034cfcb99/README.md#documentation) - @eguzki DTLGTY?

PR #44, opened 2018-07-03
service: Delete all service errors when removed

miguelsorianod:
When a Service is deleted, all of its errors should be deleted.

### DIFF
--- a/lib/3scale/backend/service.rb
+++ b/lib/3scale/backend/service.rb
@@ -219,6 +219,7 @@ module ThreeScale
       def delete_data
         delete_from_lists
         delete_attributes
+        ErrorStorage.delete_all(id)
       end
 
       def to_hash
--- a/spec/unit/service_spec.rb
+++ b/spec/unit/service_spec.rb
@@ -399,6 +399,20 @@ module ThreeScale
           expect(Service.default_id(service.provider_key)).to eq '7002'
         end
 
+        it 'deletes all service errors' do
+          Service.save! id: 7002, provider_key: 'foo', default_service: true
+          ErrorStorage.store(service.id, ApplicationNotFound.new('foo'))
+          ErrorStorage.store(service.id, ApplicationNotFound.new('foo2'))
+
+          expect { Service.delete_by_id(service.id) }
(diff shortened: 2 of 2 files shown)


lib/3scale/backend/service.rb
```diff
@@ -219,6 +219,7 @@ def storage_key(attribute)
       def delete_data
         delete_from_lists
         delete_attributes
+        ErrorStorage.delete_all(id)
```
miguelsorianod: We call delete directly on the list in ErrorStorage because it has at most 1000 entries (defined  in MAX_NUM_ERRORS in error_storage.rb)

davidor: It should be fine :+1:

spec/unit/service_spec.rb
```diff
@@ -399,6 +399,15 @@ module Backend
           expect(Service.default_id(service.provider_key)).to eq '7002'
         end
 
+        it 'deletes all service errors' do
+          Service.save! id: 7002, provider_key: 'foo', default_service: true
+          ErrorStorage.store(service.id, ApplicationNotFound.new('foo'))
```
davidor: I would store at least 2 to make sure that this works correctly and does not delete just one.

spec/unit/service_spec.rb
```diff
@@ -399,6 +399,15 @@ module Backend
...
+
+          expect(ErrorStorage.count(service.id)).to be > 0
```
davidor: With the pattern: `expect, do_something, expect`, I prefer to use `to_change`:
```ruby
expect { Service.delete_by_id(service.id) }
  .to_change(ErrorStorage.count(service.id))
  .from(2?)
  .to(0)
```

eguzki: I think you mean

```
expect { Service.delete_by_id(service.id) }.to  change { ErrorStorage.count(service.id) }.from(2?).to(0)
```

eguzki:
LGTM

Maybe adding a test trying to delete when there is nothing to delete?? It should not fail or raise any error

miguelsorianod:
Changes have been added

miguelsorianod:
bors r=@davidor,@eguzki

PR #45, opened 2018-07-04
When a Service is deleted, its latest transactions should be deleted

miguelsorianod:
For all the Services, the latest transactions that have been processed
are stored in Redis. The purpose of this is to be queried by other
components that use apisonator to show the latest traffic of the
service. Once a service is removed all the stored transactions
should be removed.


miguelsorianod:
bors r=@davidor,@eguzki

PR #46, opened 2018-07-18
Set sentinel default port when no port is provided

miguelsorianod:
Now when a Redis sentinel configuration is provided without a port
or with a nil or empty port, the sentinel default port is configured
to it.


lib/3scale/backend/storage.rb
```diff
@@ -198,9 +201,8 @@ def cfg_sentinels_handler(options)
                   raise InvalidURI.new("(sentinel #{sentinel.inspect})",
                                        'no host given')
                 end
-                sentinel.fetch(:port) do
-                  raise InvalidURI.new("(sentinel #{sentinel.inspect})",
-                                       'no port given')
+                if !sentinel[:port] || sentinel[:port] == ""
```
eguzki: `sentinel[:port] ||= DEFAULT_SENTINEL_PORT`

miguelsorianod: In that case if sentinel[:port] is "", sentinel[:port] would have the value "":

```
irb(main):003:0> sentinel = { port: "" }
=> {:port=>""}
irb(main):004:0> sentinel[:port] ||= 26379
=> ""
irb(main):005:0> sentinel[:port]
=> ""
```

The idea here was to assign the default sentinel port if originally port is the empty string

eguzki: I would not expect to be empty string, nor empty array or empy hash or any other object. Up to you.

lib/3scale/backend/storage.rb
```diff
@@ -226,8 +228,8 @@ def sentinel_to_hash(sentinel)
             valid_uri_str = to_redis_uri(sentinel)
             # it is safe to perform URI parsing now
             uri = URI.parse valid_uri_str
-
-            { host: uri.host, port: uri.port }
+            sentinel_port = uri.port || DEFAULT_SENTINEL_PORT
```
eguzki: sentinels are iterated. Instead of applying default case in configuration format, we could iterate again checking port exists, otherwise, default port. Double iteration, single place of checking for default sentinel port.

lib/3scale/backend/storage.rb
miguelsorianod: So, in this case we will let the client crash if the user configures {}, "", or [] so they can detect the mistake :+1:

test/unit/storage_test.rb
```diff
@@ -113,6 +113,35 @@ def test_sentinels_simple_url
                                       sentinels: [{ host: '127.0.0.1', port: 26_379 }])
   end
 
+  def test_sentinels_array_hashes_default_port
+    default_sentinel_port = Storage::Helpers.singleton_class.const_get(:DEFAULT_SENTINEL_PORT)
+    config_obj = {
+      url: 'redis://127.0.0.1:6379/0',
+      sentinels: [{ host: '127.0.0.1' }, { host: '192.168.1.1'},
+                  { host: '192.168.1.2', port: nil }]
```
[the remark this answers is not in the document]
miguelsorianod: No, because this is already tested in another test methods like 'test_sentinels_connection_array_hashes', 'test_sentinels_malformed_url', 'test_sentinels_connection_array_strings, etc...

The idea of this specific test is to test sentinels without ports

test/unit/storage_test.rb
```diff
@@ -113,6 +113,35 @@ def test_sentinels_simple_url
...
+                                      sentinels: [{ host: '127.0.0.1', port: default_sentinel_port },
+                                                  { host: '192.168.1.1', port: default_sentinel_port },
+                                                  { host: '192.168.1.2', port: default_sentinel_port }])
+  end
+
+  def test_sentinels_array_strings_default_port
+    default_sentinel_port = Storage::Helpers.singleton_class.const_get(:DEFAULT_SENTINEL_PORT)
+    config_obj = {
+      url: 'redis://127.0.0.1:6379/0',
+      sentinels: ['127.0.0.2', 'redis://127.0.0.1', '192.168.1.1']
```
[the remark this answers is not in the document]
unleashed: also: remember that when we _do_ specify sentinels then the url should contain the master group name - I think we don't check this and let the client complain, can you check?

test/unit/storage_test.rb
```diff
@@ -113,6 +113,35 @@ def test_sentinels_simple_url
                                       sentinels: [{ host: '127.0.0.1', port: 26_379 }])
   end
 
+  def test_sentinels_array_hashes_default_port
+    default_sentinel_port = Storage::Helpers.singleton_class.const_get(:DEFAULT_SENTINEL_PORT)
+    config_obj = {
+      url: 'redis://127.0.0.1:6379/0',
+      sentinels: [{ host: '127.0.0.1' }, { host: '192.168.1.1'},
```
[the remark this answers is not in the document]
miguelsorianod: :+1:

test/unit/storage_test.rb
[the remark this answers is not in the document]
unleashed: the combination of both things at the same time is not tested, no? :) At the very least _some_ test should do it.

test/unit/storage_test.rb
```diff
@@ -113,6 +113,35 @@ def test_sentinels_simple_url
                                       sentinels: [{ host: '127.0.0.1', port: 26_379 }])
   end
 
+  def test_sentinels_array_hashes_default_port
+    default_sentinel_port = Storage::Helpers.singleton_class.const_get(:DEFAULT_SENTINEL_PORT)
+    config_obj = {
+      url: 'redis://127.0.0.1:6379/0',
+      sentinels: [{ host: '127.0.0.1' }, { host: '192.168.1.1' },
+                  { host: '192.168.1.2', port: nil }]
```
unleashed: my comment got outdated with your latest push - I was saying that yes, specifying the port is done elsewhere, but there is still no combination in any previous test of specifying the port and not specifying it and testing correct values, so either add it here or add it in some of the previous tests.

test/unit/storage_test.rb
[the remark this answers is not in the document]
miguelsorianod: I agree that it is a good idea to test the combination of both things at the same time is a good idea to test so I have added it :+1:

If it was not for the possibility of the combination I would not think it was necessary to add an actual port in the test because it is tested in other tests already and I don't see the value on adding it in yet another test where is covered in other already existing tests.

test/unit/storage_test.rb
miguelsorianod: :+1: I answered you in the outdated comment

test/unit/storage_test.rb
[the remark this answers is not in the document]
miguelsorianod: The idea is to let the Redis client crash, because the sentinel master group names is specified in a Redis URI too.

However, I think it's a good idea to use a name more similar to a real use case in our tests. A IP was specified instead of a master group name because I copied it from another existing test that performed other sentinels tests. I have added another commit that changes this and for all the previously existing sentinel-related tests previous to this PR that have a 'url' field and a 'sentinels' field.

miguelsorianod:
bors r=@unleashed, @eguzki

PR #47, opened 2018-07-24
[WIP] Feature/stats deletion

miguelsorianod:
Work in progress. This is a PR to add stats deletion feature


unleashed:
I guess this is still extremely WIP by the looks of it

miguelsorianod:
Yes. Don't review it yet

lib/3scale/backend/stats/keys/generator/application.rb
```diff
@@ -0,0 +1,58 @@
...
+              Enumerator.new do |enum|
+                application_index_start, application_index_end = get_application_index_limits
+                metric_index_start, metric_index_end = get_metric_index_limits
+                self.applications[application_index_start..application_index_end].each_with_index do |app_id, app_range_idx|
+                  self.metrics[metric_index_start..metric_index_end].each_with_index do |metric_id, metric_range_idx|
+                    Timestamp.get_time_generator(self.service_context, self.index_limits, :app).each do |datetime_key, granularity_idx, ts|
+                      idx = Index.new(app: app_range_idx, metric: metric_range_idx, granularity: granularity_idx, ts: ts)
+                      stats_key = generate_stats_key(self.service_context.service_id, app_id, metric_id, datetime_key)
+                      enum << OpenStruct.new(key_index: idx, key_name: stats_key)
+                    end
```
unleashed: this is unreadable - things to do:
* make the `get_*_limits` return `Range` objects
* shorten the variable names, ie `app_range_idx` => `app_idx`, `metric_range_idx` => `m_idx, ...
* `self.<attribute>` is only useful for distinguishing a reader accessor from a local variable and a writer accessor call from a local variable assignment - none applies here, just use `<attribute>`
* I have not evaluated the use of `OpenStruct` yet, but if the fields are known beforehand you should use `Struct`
* this does not implement `Enumerable` (the `each` method) - it is usually enough to do just that and hide the Enumerator as an internal detail - are you perhaps looking for something like [enum_args](https://github.com/unleashed/enum_args)?
* unify this as much as possible with the metric equivalent

lib/3scale/backend/stats/keys/generator/partition.rb
```diff
@@ -0,0 +1,47 @@
...
+            # It is 20*50 because we want 20 batches of 50 keys
+            # to delete in Redis in the resque deletion job
+            DEFAULT_KEYS_SLICE = 20*50
+            private_constant :DEFAULT_KEYS_SLICE
+
+            def initialize(key_generator, options = {})
+              self.keys_slice = options[:keys_slice] || DEFAULT_KEYS_SLICE
+              self.key_generator = key_generator
+              self.key_partition_generator = Enumerator.new do |enum|
+                key_generator().each_slice(keys_slice) do |slice|
```
unleashed: `()` is not idiomatic in Ruby

lib/3scale/backend/stats/keys/generator/partition.rb
```diff
@@ -0,0 +1,47 @@
...
+
+            # It is 20*50 because we want 20 batches of 50 keys
+            # to delete in Redis in the resque deletion job
+            DEFAULT_KEYS_SLICE = 20*50
+            private_constant :DEFAULT_KEYS_SLICE
+
+            def initialize(key_generator, options = {})
+              self.keys_slice = options[:keys_slice] || DEFAULT_KEYS_SLICE
+              self.key_generator = key_generator
+              self.key_partition_generator = Enumerator.new do |enum|
```
unleashed: you don't need this - this is an Enumerator on top of another one to adapt the output. If you don't want to execute it ASAP you need to call `lazy` on any `Enumerable`.

lib/3scale/backend/stats/keys/generator/partition.rb
```diff
@@ -0,0 +1,47 @@
...
+                  enum << [slice.first, slice.last]
+                end
+              end
+            end
+
+            # TODO check if an enumerator can be returned from this if not
+            # called with a block
+            def each(&block)
+              key_partition_generator.each(&block)
+            end
```
unleashed: this whole class is a `collection.lazy.each_slice(...)`

lib/3scale/backend/stats/keys/generator/application.rb
```diff
@@ -0,0 +1,58 @@
+module ThreeScale
+  module Backend
+    module Stats
+      module Keys
+        module Generator
+          class Application
+            def initialize(service_context, index_limits)
+              self.service_context = service_context
+              self.index_limits = index_limits
+              self.applications = service_context.applications || []
```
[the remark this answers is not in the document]
miguelsorianod: This was made this way because the service_context is a private attribute accessor and self cannot be used when having a private attribute. For coherence should be getting the information from the attribute and not from the parameter.

I'll change all the classes to use @ directly in the constructors to set/get attributes of the class, and use the attribute getter/setter methods on the other methods of the class.

The idea to use getter/setter attributes is to be able to change the behaviour of them without having to update the @ references in the class code on the other methods

lib/3scale/backend/stats/keys/generator/partition.rb
miguelsorianod: :+1: I'll use @ for that because it is located in the constructor

lib/3scale/backend/stats/common.rb
```diff
@@ -0,0 +1,35 @@
...
+      module Common
+        SERVICE_GRANULARITIES =
+        [:eternity, :month, :week, :day, :hour].map do |g|
+          Period[g]
+        end.freeze
+
+        # For applications and users
+        EXPANDED_GRANULARITIES = (SERVICE_GRANULARITIES + [Period[:year], Period[:minute]]).freeze
+
+        GRANULARITY_EXPIRATION_TIME = { Period[:minute] => 180 }.freeze
```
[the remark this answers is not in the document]
miguelsorianod: :+1: , I see there is a method to obtain the expire time given a granularity that I moved into the stats/common.rb file:

```
        def self.expire_time_for_granularity(granularity)
```

I'll set that constant to private and that method public

lib/3scale/backend/stats/keys/generator/application.rb
miguelsorianod: * Regarding the get_*_limit returning Range objects: There are some limits where there is no concept of uniform/previsible succession. This is the case for example where the limits are timestamps. Another case is where the limits are the key limits, where the limits are not simple values but the combinations of other limits. A range accepts an initial value and a final value, and the implementation of the 'succ' method to calculate the intermediate values and <=> to compare them.
* About shortening the variable names, I'll do it :+1: 
* I'll change the self. by calls to the accessor method. This was a pending change indeed.
* I have created a Struct named IndexedKey instead of OpenStruct and will now be used

lib/3scale/backend/stats/common.rb
```diff
@@ -0,0 +1,61 @@
...
+        PERMANENT_EXPANDED_GRANULARITIES = (EXPANDED_GRANULARITIES - GRANULARITY_EXPIRATION_TIME.keys).freeze
+
+        #TODO maybe move this to another place??
+        PERMANENT_SVC_GRN_IDX = Stats::Common::get_indexed_granularities(PERMANENT_SERVICE_GRANULARITIES).freeze
+        PERMANENT_EXP_GRN_IDX = Stats::Common::get_indexed_granularities(PERMANENT_EXPANDED_GRANULARITIES).freeze
+
+        # We are not going to send metrics with granularity 'eternity' or
+        # 'week' to Kinesis, so there is no point in storing them in Redis
+        # buckets.
+        EXCLUDED_FOR_BUCKETS = [Period[:eternity], Period[:week]].freeze
```
eguzki: this is only for aggregators, used anywhere else?

miguelsorianod: Not used anywhere else. I'm going to move it into aggregators again :+1:

lib/3scale/backend/stats/remover/job/remover.rb
```diff
@@ -0,0 +1,32 @@
...
+
+            DELETE_BATCH_SIZE = 50
+
+            class << self
+              include Logging
+
+              def perform_logged(serialized_service_context, serialized_start_limit, serialized_end_limit, _enqueue_time)
+                service_context = Serialize.parse_json(serialized_service_context)
+                #service_context should be converted to
+              end
```
eguzki: syntax error

lib/3scale/backend/stats/service_context.rb
```diff
@@ -0,0 +1,14 @@
+# Create a keys partition generator
+module ThreeScale
+  module Backend
+    module Stats
+      class ServiceContext
```
eguzki: minor issue, but I would not call this service context. This is the removal job description.

lib/3scale/backend/stats/keys/generator/timestamp.rb
```diff
@@ -0,0 +1,126 @@
...
+                while from.start <= to.start
+                  yielder << [generate_timestamp_key(from), from.start.to_i]
+                  next_ts = from.finish
+                  from = ThreeScale::Backend::Period[granularity_name].new(next_ts)
+                end
+              end
+            end
+
+            def yield_timestamp(datetime_gen, yielder, granularity_name, granularity_idx)
+              if granularity_name == :eternity
```
eguzki: bad smell. Why not implement in Period class?

lib/3scale/backend/stats/keys/generator/timestamp.rb
```diff
@@ -0,0 +1,126 @@
...
+              limit_end = index_limits[1].granularity || 0
+              Range.new(limit_start, limit_end)
+            end
+
+            def get_granularity_limits
+              [index_limits[0].granularity, index_limits[1].granularity]
+            end
+
+            def get_time_limits
+              Range.new(index_limits[0].ts, index_limits[1].ts)
```
eguzki: Range with timestamps???

lib/3scale/backend/stats/keys/generator/timestamp.rb
```diff
@@ -0,0 +1,126 @@
...
+                  ## we recreate the period. It caches the entire Period instance
+                  ## based on the start time, so because in this case both period
+                  ## objects have the same start time, it retrieves
+                  ## the object from the period cache with the same timestamp.
+                  ## Also, it is not possible to modify the timestamp of an existing
+                  ## period because it is a private field.
+                  start_ts = ThreeScale::Backend::Period[granularity_name].new(Time.at(ts).utc).start
+                  period = ThreeScale::Backend::Period[granularity_name].new(start_ts)
+                end
+                while from.start <= to.start
```
eguzki: while condition based on ts, hence no need to create Period instances for each step

miguelsorianod: The declaration of the Period instances is needed in order to be able to use the 'start' and 'finish' methods available in the ThreeScale::Backend::Period class, which allows us to perform the increments in granularity steps

lib/3scale/backend/stats/keys/generator/timestamp.rb
```diff
@@ -0,0 +1,126 @@
...
+                  ## A problem has been detected here in this two lines. It seems it is caching
+                  ## the @timestamp instance attribute for the Period[:month] even though
+                  ## we recreate the period. It caches the entire Period instance
+                  ## based on the start time, so because in this case both period
+                  ## objects have the same start time, it retrieves
+                  ## the object from the period cache with the same timestamp.
+                  ## Also, it is not possible to modify the timestamp of an existing
+                  ## period because it is a private field.
+                  start_ts = ThreeScale::Backend::Period[granularity_name].new(Time.at(ts).utc).start
+                  period = ThreeScale::Backend::Period[granularity_name].new(start_ts)
```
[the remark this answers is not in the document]
miguelsorianod: :+1: . This part has not been cleant yet due to the problems encountered with the caching of the period mentioned above

lib/3scale/backend/stats/keys/generator/timestamp.rb
miguelsorianod: I'm not sure too of using Ranges in the case of limits/timestamps. Ranges were used at first as recommended by @unleashed and to keep consistency with all other kind of limits this was used here too. However, I'm still evaluating what would be the best way to do it as I keep doing refactors in some parts of the code

miguelsorianod:
Closed in favor of #72, #73, #74

PR #51, opened 2018-09-18
Allow deleting the default service when it's the only one for the provider

### LINKED ISSUE #50: Provide a way to delete default services
Right now there's no way to delete the default service of a provider. It probably makes sense to allow deleting it when it's the only one left.

davidor:
Closes #50

### DIFF
--- a/lib/3scale/backend/service.rb
+++ b/lib/3scale/backend/service.rb
@@ -83,7 +83,10 @@ module ThreeScale
 
         def delete_by_id(service_id)
           service = load_by_id!(service_id)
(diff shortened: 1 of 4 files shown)


davidor:
@miguelsorianod In the acceptance tests, we verify the JSON returned. From there, we cannot check the exception that was raised and captured, we can just verify the error message returned in the JSON.

miguelsorianod:
I see. This kind of tests use the API endpoints directly instead of using pisoni so there is no access to the exceptions :+1:

davidor:
bors r=@miguelsorianod

PR #55, opened 2018-10-09
Fix rate-limit headers information when rate-limited

### LINKED ISSUE #54: '3scale-Limit-Reset' Not useful when rate-limited
The limit headers return the remaining quota and the reset time based on the most constrained limit that applies. 
When the request is rate-limited, these headers return incorrect information.

I tried defining a metric with 2 limits:
1) 5 requests / minute.
2) 100k requests / year.

After making 6 requests in a minute, I would expect to see a reset-time close to 60 seconds. However, it returns the remaining for the year.

davidor:
Fixes #54 

The problem is that when calculating the remaining quota for a `usage_report`, we were taking into account the information from the `status`. In particular, we were returning a remaining of 0 if the status was not authorized. 

Suppose that we have several `usage_reports` for the same `status`. If the status is unauthorized, all of them with return a remaining of 0. This is a problem because we cannot properly compare them to figure out which one of them is causing the "unauthorized" status. We could have a report that goes over limits and another that does not.


spec/unit/limit_headers_spec.rb
```diff
@@ -78,6 +78,31 @@ module Transactor
                 expect(result).to eq expected
               end
             end
+
+            context 'and one of them has a remaining of 0' do
+              let(:time) { Time.now.utc }
+
+              let(:report_with_some_remaining) do
+                double(period: Period::Day.new(time),
```
mikz: From rspec-mock view, I think is is really, really necessary to use object/instance mocks and turn on their verification to prevent them going stale:

https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/verifying-doubles/using-an-object-double
https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/verifying-doubles/using-an-instance-double

If it is too much trouble to set up, ok, but something to bear in mind. When using mocks it is super important to verify the mocked interface.

test/integration/authorize/limit_headers_test.rb
```diff
@@ -71,6 +71,36 @@ def setup
...
+    UsageLimit.save(day_limit)
+
+    hour_limit = { service_id: @service.id,
+                   plan_id: @plan_id,
+                   metric_id: @metric_id,
+                   hour: 10 } # Stricter limit for the hour
+    UsageLimit.save(hour_limit)
+
+    # Go over limits for the hour
+    current_time = Time.now.utc
```
mikz: Isn't better to do this with explicit time? That way it would be less dynamic and more obvious what the actual numbers are.

lib/3scale/backend/transactor/usage_report.rb
```diff
@@ -57,16 +57,12 @@ def remaining_time(from = Time.now)
           # is predicted, not to be reported. The number of identical calls
           # that could be performed is 10/2 = 5.
           #
-          # Returns 0 when not authorized.
           # Returns -1 when there is not a limit in the number of calls.
           def remaining_same_calls
```
mikz: I don't see this change being tested in the unit tests, just integration. Is that fine?

spec/unit/limit_headers_spec.rb
davidor: Definitely, something to take into account for the future. We're using object_doubles only in a couple of places.
For now, I'd keep it as it is. This stubs the `UsageReport` class which is not easy to instantiate because of the params it receives and improving this could take a bit of time.

test/integration/authorize/limit_headers_test.rb
davidor: I agree. It's a bit better because when the tests fail, we'll see a number we can recognize.
I'll change it.

spec/unit/limit_headers_spec.rb
mikz: `instance_double` should skip the initializer (at least that's my understanding), or you can use object double with `UsageReport.allocate` without calling the initializer. It is really just about the interface.

But sure, 👍 if you want to keep it this way.

lib/3scale/backend/transactor/usage_report.rb
davidor: I think it's fine. `UsageReport` does not have any unit tests. I guess it's because it was extracted from another class and its tests were not.

In this case, this method has only one caller, `LimitsHeader`, and I added a unit test for it. That unit test and the integration tests I added should cover everything introduced in this PR. Anyway, I'll consider adding one more for `UsageReport`.

spec/unit/limit_headers_spec.rb
davidor: `instance_double` does not skip the initializer.
I like the `.allocate` solution though, so I'll use that :+1:

davidor:
@mikz I addressed your comments :+1:

davidor:
bors r=@eguzki

PR #61, opened 2018-11-09
Makefile: remove ruby from building steps

eguzki:
Remove requirement from the building environment to have ruby installed. Docker will be the only tool required to build the project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,14 +84,16 @@ dev-build: $(PROJECT_PATH)/Dockerfile
 		$(PROJECT_PATH)
 
 .PHONY: ci-build
-ci-build: export APISONATOR_REL?=v$(shell ruby -r$(PROJECT_PATH)/lib/3scale/backend/version -e "puts ThreeScale::Backend::VERSION")
+ci-build: export APISONATOR_REL?=v$(shell docker run --rm -w /tmp/apisonator -v $(PROJECT_PATH):/tmp/apisonator:z \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -84,14 +84,16 @@ dev-build: $(PROJECT_PATH)/Dockerfile
 		$(PROJECT_PATH)
 
 .PHONY: ci-build
-ci-build: export APISONATOR_REL?=v$(shell ruby -r$(PROJECT_PATH)/lib/3scale/backend/version -e "puts ThreeScale::Backend::VERSION")
+ci-build: export APISONATOR_REL?=v$(shell docker run --rm -e RBENV_VERSION='2.3.6' -v $(PROJECT_PATH):/tmp/apisonator:z \
```
unleashed: can we remove the `RBENV_VERSION` there? If not, perhaps we can just specify `2.3` or even `2` - otherwise we will be depending on the CI image always having a 2.3.6 release.

eguzki: Is there any way to disable rbenv or use default version? Otherwise, using another image could work, like `centos/ruby-23-centos7`

unleashed: also, using a different image would work but would not be practical as another image would need to be download just for this and you would still encode the version here instead of using the app/ci image version

eguzki: used `.ruby-version` of the project :+1:

eguzki:
Maybe this is the cause?
https://circleci.com/docs/2.0/workflows-waiting-status/

PR #64, opened 2018-11-16
Fix issues for CI environments

eguzki:
- [x] Fix sleep time for postgress to start
- [x] docker: remove allocating TTY required for some CI environments


unleashed:
Is there any reason the `-ti` removal is needed other than "things break"?

eguzki:
let me answer with a question:

what is the reason to keep stdin open (`-i`) or allocate tty (`-t`) on docker run when you are running tests?

I think it is better to run in non-interactive mode and potential environments from CI would not fail. 

Maybe we can leave `-t` for TTY support for color formatting.

unleashed:
an answer is not a question and you are not Galician, so not entitled to do that... :)

unleashed:
That said... yes, this is meant to have colors in the `make test` output, and I think `-i` could go away. In any case a special invocation just for CI could leave out the `-t` if really needed.

eguzki:
bors r=@unleashed

PR #71, opened 2019-02-05
[WIP] Feature/delete service stats

eguzki:
3 new components:

* New endpoint to delete stats in internal API

```
```

Request body

```json
{
  "deletejobdef": {
    "applications": ["1"],
    "metrics": ["5"], 
    "from": 1483228800, 
    "to": 1483228800, 
    "users": ["1"]
  }
}
```
* `Stats::PartitionEraserJob`
Background job to delete stat keys
 
* `Stats::PartitionGeneratorJob`
Background job to generate `Stats::PartitionEraserJob` for each stats keys subset

TODO
- [ ] Testing
- [ ] Doc
- [ ] Read some settings from configuration parameters


unleashed:
@eguzki is there a way to introduce these 3 components in 3 different PRs? Even if they are "incomplete" or they would need further tweaking, it would make reviewing easier. Alternatively we can create an integration branch where we get the 3 PRs merged to (as opposed to master) and then we merge the integration branch in another PR (David did this back some time ago for other larger work and it worked quite nicely).

eguzki:
It would make more sense two PR. One of them with the internal endpoint and the other PR with both backend worker jobs. Both jobs cannot be understand standalone. Furthermore, they share stats keys generator (and all code behind it). One of them create partitions and the other delete keys on storage, but the difference is just few lines of code. Most of the code lies on stats keys generator.

Will do it

eguzki:
Created
* Internal endpoint  https://github.com/3scale/apisonator/pull/72
* worker jobs https://github.com/3scale/apisonator/pull/73
* Stats key generator  https://github.com/3scale/apisonator/pull/74

unleashed:
@eguzki what would be the dependencies? #74 depends on #73 depends on #72?

eguzki:
Both #72 and #73 depend on #74.

PR #72, opened 2019-02-06
Feature/delete service stats endpoint

eguzki:
New endpoint to delete stats in internal API

```
```

Request body

```json
{
  "deletejobdef": {
    "applications": ["1"],
    "metrics": ["5"], 
    "from": 1483228800, 
    "to": 1483228800, 
    "users": ["1"]
  }
}
```


app/api/internal/stats.rb
```diff
@@ -0,0 +1,23 @@
...
+        before do
+          respond_with_404('service not found') unless Service.exists?(params[:service_id])
+        end
+
+        delete '' do |service_id|
+          begin
+            delete_stats_job_attrs = api_params Stats::DeleteJobDef
+            delete_stats_job_attrs[:service_id] = service_id
+            Stats::DeleteJobDef.new(delete_stats_job_attrs).run_async
+          rescue Error => e
```
mikz: I find it more readable to refer to internal constants with full name because it is hard to know if it is a global or not.

eguzki:
ping @unleashed @davidor

app/api/internal/stats.rb
```diff
@@ -0,0 +1,23 @@
...
+        before do
+          respond_with_404('service not found') unless Service.exists?(params[:service_id])
+        end
+
+        delete '' do |service_id|
+          begin
+            delete_stats_job_attrs = api_params Stats::DeleteJobDef
+            delete_stats_job_attrs[:service_id] = service_id
+            Stats::DeleteJobDef.new(delete_stats_job_attrs).run_async
+          rescue Backend::Error => e
```
davidor: Could we be more specific here? Is `DeleteServiceStatsValidationError` the only error that can be raised?

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,83 @@
+module ThreeScale
+  module Backend
+    module Stats
+      class DeleteJobDef
```
davidor: We need tests for this class. There are many error cases in the validation that need to be tested.

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,83 @@
...
+        ATTRIBUTES = %i[service_id applications metrics users from to context_info].freeze
+        private_constant :ATTRIBUTES
+        attr_accessor(*ATTRIBUTES)
+
+        def self.attribute_names
+          ATTRIBUTES
+        end
+
+        def initialize(params = {})
+          self.class.attribute_names.each do |key|
```
davidor: Can't we use `ATTRIBUTES.each` ?

app/api/internal/stats.rb
```diff
@@ -0,0 +1,23 @@
+module ThreeScale
+  module Backend
+    module API
+      internal_api '/services/:service_id/stats' do
+        before do
+          respond_with_404('service not found') unless Service.exists?(params[:service_id])
+        end
+
+        delete '' do |service_id|
```
davidor: We need tests for this endpoint.

app/api/internal/stats.rb
```diff
@@ -0,0 +1,23 @@
...
+          begin
+            delete_stats_job_attrs = api_params Stats::DeleteJobDef
+            delete_stats_job_attrs[:service_id] = service_id
+            Stats::DeleteJobDef.new(delete_stats_job_attrs).run_async
```
davidor: I think this is the only line that can raise. If so, it should be the only one in `begin`.

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,83 @@
...
+            send("#{key}=", params[key]) unless params[key].nil?
+          end
+        end
+
+        def run_async
+          validate
```
davidor: Should we wait until this point to validate the attributes?
Wouldn't it make more sense to validate on `initialize()` and on the individual setters (if they need to be exposed) ?

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,83 @@
...
+        end
+
+        def to_hash
+          Hash[self.class.const_get(:ATTRIBUTES).collect { |key| [key, send(key)] }]
+        end
+
+        def validate
+          # from and to valid epoch times
+          unless from.is_a? Integer
+            raise DeleteServiceStatsValidationError.new(service_id, 'from field validation error')
```
davidor: This pattern is repeated several times.
Would it make sense to extract it to a function, something like this?

```ruby
def raise_error(msg)
  raise DeleteServiceStatsValidationError.new(service_id, msg)
end
```

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,83 @@
...
+          if Time.at(to) < Time.at(from)
+            raise DeleteServiceStatsValidationError.new(service_id, 'from < to fields validation error')
+          end
+
+          # application is array
+          unless applications.is_a? Array
+            raise DeleteServiceStatsValidationError.new(service_id, 'applications field validation error')
+          end
+
+          if applications.size > 0
```
davidor: `.all?` returns true for empty collections, so I think that this check is redundant.

app/api/internal/stats.rb
eguzki: :+1:

lib/3scale/backend/stats/delete_job_def.rb
eguzki: :+1:

app/api/internal/stats.rb
eguzki: :+1:

app/api/internal/stats.rb
eguzki: :+1:

lib/3scale/backend/stats/delete_job_def.rb
eguzki: :+1:

lib/3scale/backend/stats/delete_job_def.rb
eguzki: :+1:

lib/3scale/backend/stats/delete_job_def.rb
eguzki: :+1:

eguzki:
all coments resolved. Ready for a new review @davidor

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,59 @@
+module ThreeScale
+  module Backend
+    module Stats
+      class DeleteJobDef
+        ATTRIBUTES = %i[service_id applications metrics users from to context_info].freeze
+        private_constant :ATTRIBUTES
+        attr_accessor(*ATTRIBUTES)
```
davidor: Do we need to expose the setter methods?

Right now, it's possible to initialize a valid instance, modify it with an invalid value, and then call .run_async.

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,59 @@
...
+        def validate
+          # from and to valid epoch times
+          raise_validation_error('from field not integer') unless from.is_a? Integer
+          raise_validation_error('from field is zero') if from.zero?
+          raise_validation_error('to field not integer') unless to.is_a? Integer
+          raise_validation_error('to field is zero') if to.zero?
+          raise_validation_error('from < to fields') if Time.at(to) < Time.at(from)
+          # application is array
+          raise_validation_error('applications field') unless applications.is_a? Array
+          raise_validation_error('applications values') unless applications.all? { |x| x.is_a? String }
```
davidor: Do we need to verify that all these are strings? I think that they could also be Integers.

In the model classes, we do not have this assumption. For example, when loading a service, sending '123' or 123 as the ID makes no difference.

davidor:
@eguzki , I added a couple of comments. The rest looks good to me.

lib/3scale/backend/stats/delete_job_def.rb
eguzki: yeah, that's why validation was done in `.run_async` in the beginning. I will implement read-only class. :+1:

lib/3scale/backend/stats/delete_job_def.rb
eguzki: :+1:

lib/3scale/backend/stats/delete_job_def.rb
```diff
@@ -0,0 +1,65 @@
...
+          to_hash.to_json
+        end
+
+        def to_hash
+          Hash[ATTRIBUTES.collect { |key| [key, send(key)] }]
+        end
+
+        private
+
+        def validate
```
davidor: I forgot to mention that we should probably put some kind of limit on the `to` parameter or on the time interval between `to` and `from`. Otherwise, someone could send a call to delete all the keys from 1970 up until now or something like that. We can fix this in the integration branch later.

PR #73, opened 2019-02-06
Feature/delete service stats worker jobs

eguzki:
* `Stats::PartitionEraserJob`
Background job to delete stat keys
 
* `Stats::PartitionGeneratorJob`
Background job to generate `Stats::PartitionEraserJob` for each stats keys subset


lib/3scale/backend/stats/partition_eraser_job.rb
```diff
@@ -0,0 +1,72 @@
...
+            stats_key_gen.keys.drop(offset).take(length).each_slice(DELETE_BATCH_SIZE) do |slice|
+              storage.del(slice)
+            end
+
+            [true, { job: job.to_hash, offset: offset, lenght: length }.to_json]
+          rescue Error => error
+            ErrorStorage.store(service_id, error, context_info)
+            [false, "#{service_id} #{error}"]
+          rescue Exception => error
+            if error.class == ArgumentError && error.message == 'invalid byte sequence in UTF-8'
```
mikz: Shouldn't this be rescued somewhere closer to where it is actually being raised?

eguzki: This code has been copied from [ReportJob](https://github.com/3scale/apisonator/blob/master/lib/3scale/backend/transactor/report_job.rb), I think it does not apply for this type of job. :+1:

lib/3scale/backend/stats/partition_eraser_job.rb
```diff
@@ -0,0 +1,64 @@
+module ThreeScale
+  module Backend
+    module Stats
+      # TODO: from configuration
+      DELETE_BATCH_SIZE = 50
+      # Job for deleting service stats
+      # Perform actual key deletion from a key partition definition
+      class PartitionEraserJob < BackgroundJob
+        # low priority queue
+        @queue = :main
```
davidor: I think this should be in the `stats` queue or a new one that goes after all the existing ones.

`Main` has `NotifyJob`s that keep updated usage statistics on the master account. That should have precedence over these new jobs which are responsible only for cleanup tasks.

lib/3scale/backend/stats/partition_generator_job.rb
```diff
@@ -0,0 +1,51 @@
+module ThreeScale
+  module Backend
+    module Stats
+      # TODO: from configuration
+      PARTITION_BATCH_SIZE = 1000
+      # Job for deleting service stats
+      # Maps delete job definition to a set of non overlapping key set partitions
+      class PartitionGeneratorJob < BackgroundJob
+        # low priority queue
+        @queue = :main
```
davidor: Same as above.

lib/3scale/backend/stats/partition_eraser_job.rb
```diff
@@ -0,0 +1,64 @@
+module ThreeScale
+  module Backend
+    module Stats
+      # TODO: from configuration
+      DELETE_BATCH_SIZE = 50
```
davidor: In this PR, this constant is only used in `PartitionEraserJob`, so I think it should be defined there and declared private.
I haven't checked all the related PRs yet, so not sure if that will change.

Same for the `PARTITION_BATCH_SIZE` constant defined in the other job class.

davidor:
There are mainly 2 aspects of this PR that I think should be addressed before merging.

The first one is that there are no tests. The second one is that, in my opinion, the responsibilities of the 2 types of jobs (`PartitionGeneratorJob` and `PartitionEraserJob`) are not well defined and they do a lot of duplicated work.

`PartitionGeneratorJob` generates all the keys, but it discards them and only passes to each of the `PartitionEraserJob`s an offset. Then, each `PartitionEraserJob` generates again all the keys and discards all of them except a few ones (just picks N starting from the 'offset' received). It's a lot of duplicated work.

Wouldn't it be better to generate all the keys in the Generator job and just pass X of them to each of the Delete jobs? That way, the jobs would not perform any duplicated work. Also, I think that conceptually it would be easier to understand, because one job type would be responsible for generating all the keys, and the other just for deleting the keys that it received.

lib/3scale/backend/stats/partition_eraser_job.rb
eguzki: :+1:

lib/3scale/backend/stats/partition_generator_job.rb
eguzki: :+1:

lib/3scale/backend/stats/partition_eraser_job.rb
eguzki: both `DELETE_BATCH_SIZE` and `PARTITION_BATCH_SIZE` are designed to be configuration parameters

eguzki:
The design was made to avoid serializing/deserializing to/from redis **all** the keys that have to be deleted. You are proposing to do that.

Besides, resque job size (basically json body size) can be big. The body size is controlled by `PARTITION_BATCH_SIZE` configuration parameter. Having `PARTITION_BATCH_SIZE` ~ 1000 keys, it makes ~ 1000 * (50 bytes per key) = 50k the size of a `PartitionEraserJob` resque job. 

 The trade off is 
* small `PARTITION_BATCH_SIZE` => small job body and lots of `PartitionEraserJob` jobs
* big `PARTITION_BATCH_SIZE` => big job body and few `PartitionEraserJob` jobs. If some job fail, a new worker will be processed and the big partition will be erased again (many keys could already been deleted)

Other implementation would be passing indexes besides jobs, and eraser worker will only have to generate required keys, hence no duplicated work. Resque jobs would be small in size and no keys would be serialized to redis. However the implementation of indexes increases complexity of the key generator and I decided to leave it for now. Maybe future improvement.

WDYT @davidor ? Can we afford big jobs and serializing all the keys including them in job bodies?

eguzki:
Implementation of `PARTITION_BATCH_SIZE` and `DELETE_BATCH_SIZE` as configuration params with default values

davidor:
Serializing/deserializing could be a problem too. This is the kind of thing that needs to be measured.

What if `Stats::PartitionGeneratorJob` generated `to` and `from` timestamps instead of indexes? Would that be a valid option? That way each job would be responsible for generating and deleting a part of the keys without duplication and without needing to serialize/deserialize the keys.

For example, if we need to delete a whole year of stats, we could generate one job for each month, or for each day. Not sure what would be the most appropriate granularity. Or maybe split by apps or metrics?

eguzki:
Partitioning by time, applications, metrics or users is rough way of partitioning. You cannot control the number of keys received by a worker to be deleted. With the current implementation, the size of a partition is controlled by a configuration parameter.

Anyway, it is a partitioning, yes.

We could leave it like it is now and write  an issue to implement indexes. WDYT @davidor ?

lib/3scale/backend/stats/partition_eraser_job.rb
```diff
@@ -0,0 +1,63 @@
...
+              users: users,
+              from: from,
+              to: to
+            )
+
+            validate_job(job, offset, length)
+
+            stats_key_types = KeyTypesFactory.create(job)
+
+            stats_key_gen = KeyGenerator.new(stats_key_types)
```
davidor: I think it would make sense to define an interface that returns the keys with one call instead of 2 like here.

eguzki: :+1:

spec/integration/job_queues_spec.rb
```diff
@@ -14,6 +14,7 @@ module Backend
       let(:queue) { 'priority' }
 
       before do
+        Resque::Failure.clear
```
mikz: This is something that should be done between any two tests no? So, a global hook ?

eguzki: If I am not wrong, default hook of `before` is `:each`, to between any two tests

mikz: I meant globally, for the whole test suite. This is something that can happen and leak across different files, right?

eguzki: oh, you mean in `RSpec.configure` so it will be cleared between any two tests of the whole test suite.

I will put  there

eguzki:
@davidor ready for a review

openshift/3scale_backend.conf
```diff
@@ -34,6 +34,8 @@ ThreeScale::Backend.configure do |config|
   config.kinesis_stream_name = "#{ENV['CONFIG_KINESIS_STREAM_NAME']}"
   config.kinesis_region = "#{ENV['CONFIG_KINESIS_REGION']}"
   config.stats.bucket_size = ENV['CONFIG_STATS_BUCKET_SIZE'].to_i
+  config.stats.delete_batch_size = "#{ENV['CONFIG_DELETE_STATS_BATCH_SIZE']}"
```
davidor: Should we use `parse_int_env` here? Like in other int envs above like `CONFIG_REDIS_CONNECT_TIMEOUT`, `CONFIG_REDIS_WRITE_TIMEOUT`, etc.

lib/3scale/backend/stats/key_generator.rb
```diff
@@ -4,7 +4,7 @@ module Stats
       class KeyGenerator
         attr_reader :service_id, :applications, :metrics, :users, :from, :to
 
-        def initialize(service_id:, applications: [], metrics: [], users: [], from:, to:)
+        def initialize(service_id:, applications: [], metrics: [], users: [], from:, to:, **)
```
davidor: Why do we need `**` here?

openshift/3scale_backend.conf
eguzki: it is parsed here, in `Configuration` class and assigned default value if empty: https://github.com/3scale/apisonator/pull/73/files#diff-5c109eb6e5b153d75018f097b9887048R111

lib/3scale/backend/stats/key_generator.rb
eguzki: `context_info` is passed as well. But I will check it required.

lib/3scale/backend/stats/partition_eraser_job.rb
```diff
@@ -0,0 +1,59 @@
...
+            end
+
+            [true, { job: job.to_hash, offset: offset, lenght: length }.to_json]
+          rescue Backend::Error => error
+            [false, "#{service_id} #{error}"]
+          end
+
+          private
+
+          def validate_job(job, offset, length)
```
davidor: Can this raise here?

I think the code is already performing these validations when the request comes through the internal API right? If it finds a problem there it returns an error to the caller so that they can fix the issue.

If we reach this point without returning any error before, the caller will not have a chance to fix the issue. This will just log an error in backend.

lib/3scale/backend/stats/key_generator.rb
eguzki: not required now. But `DeleteJobDef` has `context_info` to be used here
```ruby
Resque.enqueue(PartitionGeneratorJob, Time.now.getutc.to_f, service_id, applications,
               metrics, users, from, to, context_info)                               
```

lib/3scale/backend/stats/partition_eraser_job.rb
davidor: I think that as an extra protection this might be fine, but if this ever raises, we will need to fix some code in the above layers.

eguzki: this method validates extra params new and specific for this worker job:

`offset` and `length`

They are created by `PartitionGeneratorJob`. Internal API does not receive then

davidor: Right, I was confused.

lib/3scale/backend/stats/partition_generator_job.rb
```diff
@@ -1,12 +1,39 @@
...
+              metrics: metrics,
+              users: users,
+              from: from,
+              to: to
+            )
+
+            stats_key_gen = KeyGenerator.new(job.to_hash)
+
+            # Generate partitions
+            0.step(stats_key_gen.keys.count, configuration.stats.delete_partition_batch_size).each do |idx|
```
davidor: If this fails in some call to `Resque.enqueue`, I think that the job will be marked as failed, re-enqueued at some point, and then it will start scheduling duplicated `PartitionEraserJob`s. Is that right?

openshift/3scale_backend.conf
davidor: :+1:

lib/3scale/backend/stats/partition_generator_job.rb
eguzki: totally right! 

happy to hear improvements

davidor: I don't see an easy solution that fits the current approach.

You could rescue from Redis errors and retry. That could work if there's minimal downtime and retrying a few times work. If after retrying a few times the problem persists, we'd need to mark the job as failed.

What if we added an optional index to the `job` so it remembers what's already been enqueued? Would that work?

lib/3scale/backend/stats/partition_eraser_job.rb
```diff
@@ -0,0 +1,59 @@
...
+              users: users,
+              from: from,
+              to: to
+            )
+
+            validate_job(job, offset, length)
+
+            stats_key_gen = KeyGenerator.new(job.to_hash)
+
+            stats_key_gen.keys.drop(offset).take(length).each_slice(configuration.stats.delete_batch_size) do |slice|
```
davidor: I'm still not convinced about the idea of generating the whole set of keys on every `PartitionEraserJob`. It'd be good to hear other opinions on this. @unleashed ? @miguelsorianod ?

This might be good enough for a first version but we'll need to closely measure the impact on CPU time consumed by this kind of worker with something like `stackprof`. It's difficult to tell now because there are many factors that can have an impact like number of this kind of jobs generated, the average `from`-`to` interval, etc.

At least we should add a TODO here explaining what this does and mention that this is probably something that could be improved.

miguelsorianod: isn't just a subset of the keys generated/removed (length) ?

eguzki: it is generating all keys up to `offset + length`.

I will write an issue about improving the algorithm to use indexes. 

Deleting service stats is not a job that will be created often

davidor: Only a subset is deleted, but all of them are generated `stats_key_gen.keys`.

lib/3scale/backend/stats/partition_generator_job.rb
eguzki: you mean, if fails, rescue it and add `idx` to this job, so when requeued from failed jobs can resume from it? is that possible?

lib/3scale/backend/stats/partition_eraser_job.rb
miguelsorianod: :+1: , I was thinking on the old implementation, when enumerators were used and not all keys were generated.

lib/3scale/backend/stats/partition_generator_job.rb
davidor: I think so. I see 2 possible approaches:

- Set the index in the job and let it fail so the failed jobs scheduler picks it. We'd need to make sure that when rescheduling the failed job, Resque picks the mutated one instead of the original one.
- Set the index in the job, discard the current job, and enqueue the new one normally instead of letting it fail.

I don't have a strong preference, but I think the first approach might be better because it will allow  us to see some activity in the `failed` queue and notice that there might be something wrong with our infra.

eguzki: But if it fails, it will be by network issue. So jobs could not be re-scheduled in a "short" term.

I mean, the only error in this `PartitionGeneratorJob` can be originated enqueuing jobs. So if fails, how can be re-scheduled?

davidor: If there's a long downtime the job will simply fail and never be retried unless we implement something like retries with back-offs. I don't think it's worth it right now.

spec/integration/stats_partition_generator_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      from: from,
+      to: to
+    )
+  end
+  let(:stats_queue) { :stats }
+  let(:configuration) { ThreeScale::Backend.configuration }
+  let(:num_keys_generated) { ThreeScale::Backend::Stats::KeyGenerator.new(job).keys.count }
+
+  before :each do
+    ThreeScale::Backend::Worker::QUEUES.each { |queue| Resque.remove_queue(queue) }
```
davidor: We could move this to a global before like you did with `Resque.clear`.

spec/integration/stats_partition_generator_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    configuration.stats.delete_partition_batch_size = 100
```
davidor: Is the configuration reset on every test? I don't remember right now, but I don't see anything in the test helpers.

If this is not reset, we should add an `after` clause to restore the original value or add a reset of the config to the global `before`.

spec/integration/stats_partition_eraser_job_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      from: from,
+      to: to
+    }
+  end
+  let(:stats_queue) { :stats }
+  let(:storage) { ThreeScale::Backend::Storage.instance }
+  let(:keys) { ThreeScale::Backend::Stats::KeyGenerator.new(job_params).keys }
+
+  before :each do
+    storage.flushdb
```
davidor: This is redundant. It's already done in the global `before` in the helper.

spec/integration/stats_partition_eraser_job_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  it 'partition keys are deleted' do
+    without_resque_spec do
+      Resque.enqueue(described_class, Time.now.getutc.to_f, service_id, applications,
+                     metrics, users, from, to, offset, length, nil)
+      expect(Resque.size(stats_queue)).to eq 1
+      # Try to process the job.
+      ThreeScale::Backend::Worker.work(one_off: true)
+      expect(keys.drop(offset).take(length).count).to be > 0
+      keys.drop(offset).take(length).each do |key|
+        expect(storage.get(key)).to be_nil
```
davidor: Nitpick: you can use `exists` instead of `get` for this. Also, this can be rewritten to use a single expect.

```ruby
keys_to_be_deleted = keys.drop(offset).take(length)
expect(keys_to_be_deleted.none? { |key| storage.exists(key) })
```

lib/3scale/backend/stats/partition_generator_job.rb
eguzki: I think if this job fails, handing off to the `Resque::Failure` will also probably fail.

Anyway, I do not see any way to "update" de job with some index.

if `perform_logged` looks like this

```ruby
def perform_logged(_enqueue_time, service_id, applications, metrics, users,
                   from, to, partition_idx, context_info = {})
```
and I update `partition_idx`,
```ruby
partition_idx = partition_idx + 1
```

On re-scheduling, `partition_idx` will have the new value???

spec/integration/stats_partition_generator_job_spec.rb
eguzki: I think when testing, each test has to `set` its own values.  

assuming a certain configuration key X will have value V is an error and goes agains test isolation paradigm

lib/3scale/backend/stats/partition_generator_job.rb
davidor: I don't know. I'll need to check the code that marks the job as failed and stores it in the failed queue.

If that does not work, we always have the option of discarding the job and enqueueing a new one in the 'stats' queue instead of the failed one with Resque.enqueue. However, as you said, this might be a bit overkill, because it only protects against small network hiccups. Not sure if we should tackle this for a first version. You can give it a try and see whether it complicates the code a lot.

eguzki: tried above, updating `partition_idx` and does not work.

On calling `Resque::Job#fail` instance method,  job is handed off to the `Resque::Failure`, but we do not have access to the job_params.

Unless, anybody knows a way to handle this, I suggest we go on and write an issue to implement this kind of "smart retry" strategy.

davidor: OK :+1:

spec/integration/stats_partition_generator_job_spec.rb
davidor: I agree with that, but in this case, we set the config with these values in a test helper: https://github.com/3scale/apisonator/blob/master/test/test_helpers/configuration.rb

I think it's reasonable to expect that those defaults apply to each test. However, in order to ensure that each test needs to restore the config to its original values if they modify them, or we need to do that in a global `before`.

You don't have to fix this now because it's a more general problem. I see that we have more tests with the same case, but I think we should fix this at some point.

davidor:
Good job @eguzki  :+1: 
I think this can be merged to the integration branch.

spec/integration/stats_partition_generator_job_spec.rb
eguzki: that test_helper is for Test::Unit. We are discussing a RSpec test

davidor: But it's used in RSpec too https://github.com/3scale/apisonator/blob/bd63f9d59476d7913fa662ceebc33625dc07a982/spec/spec_helper.rb#L37

It's confusing. We should probably separate the helpers so it's clear that they're used from both test suites.

lib/3scale/backend/stats/partition_generator_job.rb
atzorvas: composing a unique key based on params and the parent job ID could be a possible solution here? :wave: 

Duplicate keys shouldn't fail so the exception should be handled gracefully. But on the other side, there could be some kind of logging in place (most probably with a small retention period) in order to be able to act upon such occurrences ie if too frequent

PR #78, opened 2019-03-01
Feature/delete service stats

eguzki:
Given a Service, delete all its Stats

The number of stats from a given service A can be huge. It is roughly estimated that there are 300K keys of stats per metric, application, user and year. Deleting all stat keys at once is not a good idea.

Stats deletion task partitioning system has been implemented using resque async background jobs

Internal endpoint has been published for that task:
```
```
Request body
```yaml
openAPIV3Schema:
  properties:
    deletejobdef:
        properties:
          applications:
            type: array
          metrics:
            type: array
          users:
            type: array
          from:
            type: int64
          to:
            type: int64
        required:
        -  applications
        -  metrics
        -  users
        -  from
        -  to
```

Example
```sh
$ cat delete_job.json
{ 
  "deletejobdef": {
     "applications": ["1"],
     "metrics": ["5"],
     "from": 1483228800,
     "to": 1483228800, 
     "users": []
  }
}

$ curl -v -X DELETE -d '@delete_job.json' -H "Content-Type: application/json" http://127.0.0.1:3000/internal/services/c2f28d6c-aa75-46d4-b5d8-5bc8d2dc0cff/stats
```

Doc about implementation design is included


davidor:
This is the integration branch that contains #72 , #73  and #74 .

@eguzki  Have you introduced any new changes in this branch? Just so we know if there's anything new that we should review.

eguzki:
@davidor I have squashed several commit as @mikz requested

davidor:
@eguzki I just tried this in staging and noticed a problem.

Some jobs fail with `KeyErrorlib` in `/3scale/backend/stats/codes_commons.rb:12`

The problem is that this raises an exception when a user reports a response code that Apisonator does not track, for example, 304.

Instead of raising, we should just ignore the code.

eguzki:
thnx @davidor for reporting.

Fix: https://github.com/3scale/apisonator/pull/82

PR #80, opened 2019-03-03
 extensions/time: make to_not_compact_s 2x faster

unleashed:


davidor:
I don't think we need to optimize `to_not_compact_s`.

This case is very different from #81 where `to_compact_s` was optimized.

`to_compact_s` is called multiple times per job and using profiling tools we could show that it appears as the second method that takes more CPU time.

`to_not_compact_s` is only called from https://github.com/3scale/apisonator/blob/a99715d10f5816908bdd4184f8c48663728a4255/lib/3scale/backend/stats/aggregator.rb#L55 which is called at most once per job and it does not appear in the list of methods that take a significant amount of CPU time.

unleashed:
We don't _need_ to do it, but we _have_ it now. :)

davidor:
If I were a new contributor and saw a benchmark for that method I'd think that it's one of the important ones, but it's not. I don't see any reason to merge that benchmark.

unleashed:
I agree, don't merge the benchmark! :)

unleashed:
@davidor benchmark removed from this PR

PR #82, opened 2019-03-04
Fix/response code processing

eguzki:
* Fix https://github.com/3scale/apisonator/pull/78#issuecomment-469279602
Do not raise an exception when a user reports a response code that Apisonator does not track, for example, 304.

* Do not account for untracked http response codes, currently, for example 304

### DIFF
--- a/lib/3scale/backend/stats/codes_commons.rb
+++ b/lib/3scale/backend/stats/codes_commons.rb
@@ -3,13 +3,10 @@ module ThreeScale
     module Stats
       module CodesCommons
         TRACKED_CODES = [200, 404, 403, 500, 503].freeze
-
-        HTTP_CODE_GROUPS_MAP = TRACKED_CODES.each_with_object({}) do |code, hsh|
-          hsh[code / 100] = "#{code / 100}XX"
-        end
+        TRACKED_CODE_GROUPS = ['2XX'.freeze, '4XX'.freeze, '5XX'.freeze].freeze
 
         def self.get_http_code_group(http_code)
-          HTTP_CODE_GROUPS_MAP.fetch(http_code / 100)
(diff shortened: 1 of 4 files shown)


lib/3scale/backend/stats/codes_commons.rb
```diff
@@ -4,13 +4,13 @@ module Stats
...
-        HTTP_CODE_GROUPS_MAP = TRACKED_CODES.each_with_object({}) do |code, hsh|
-          hsh[code / 100] = "#{code / 100}XX"
-        end
-
         def self.get_http_code_group(http_code)
-          HTTP_CODE_GROUPS_MAP.fetch(http_code / 100)
+          "#{http_code / 100}XX"
         end
+
+        TRACKED_CODE_GROUPS = TRACKED_CODES.each_with_object({}) do |code, group|
```
davidor: I'd change this to `TRACKED_CODE_GROUPS = ['2xx', '4xx', '5xx']` for simplicity given that the set of `TRACKED_CODES` is unlikely to change.

lib/3scale/backend/stats/codes_commons.rb
```diff
@@ -3,13 +3,10 @@ module Backend
     module Stats
       module CodesCommons
         TRACKED_CODES = [200, 404, 403, 500, 503].freeze
-
-        HTTP_CODE_GROUPS_MAP = TRACKED_CODES.each_with_object({}) do |code, hsh|
-          hsh[code / 100] = "#{code / 100}XX"
-        end
+        TRACKED_CODE_GROUPS = %w[2XX 4XX 5XX].freeze
```
mikz: Those are not frozen strings in that array.

PR #87, opened 2019-03-08
Disable stats deletion endpoint

davidor:
This endpoint has performance issues that will need to be fixed before we enable it again.

### DIFF
--- a/app/api/internal/stats.rb
+++ b/app/api/internal/stats.rb
@@ -6,6 +6,9 @@ module ThreeScale
           respond_with_404('service not found') unless Service.exists?(params[:service_id])
         end
 
+        # This is very slow and needs to be disabled until the performance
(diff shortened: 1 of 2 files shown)


app/api/internal/stats.rb
```diff
@@ -6,6 +6,9 @@ module API
           respond_with_404('service not found') unless Service.exists?(params[:service_id])
         end
 
+        # This is very slow and needs to be disabled until the performance
+        # issues are solved. In the meanwhile, the job will just return OK.
```
Martouta: > In the meanwhile, the job will just return OK

Good for me then 😜

davidor: Yes, it's done this way to avoid breaking compatibility with the latest version of Pisoni.

PR #93, opened 2019-03-20
Non-blocking Redis calls: run tests with both blocking/non-blocking drivers

davidor:
This PR continues the work started in #77

It just adapts `script/test` to run all the tests with both blocking/non-blocking storage drivers. The reason is that we'll need both for some time.

Notice that the base branch of this PR is the integration branch `async`, not `master`.

Note: CircleCI fails because the CI image contains a ruby version that is not compatible with this feature (async lib requires ruby >= 2.2.7).


test/test_helpers/storage.rb
```diff
@@ -87,7 +105,14 @@ def respond_to_missing?(m)
               attr_reader :inner
 
               def non_proxied_instances
-                ::ThreeScale::Backend::StorageAsync::Client.non_proxied_instances
+                klass = if inner.class == ThreeScale::Backend::StorageAsync::Client
```
mikz: wouldn't this could be nicer as a switch/case?

davidor: :+1:

PR #96, opened 2019-04-24
Non-blocking redis calls using redis-async lib

davidor:
This is an integration branch. It contains: #77 , #86 , #92 , #93 .
~It cannot be merged yet. We need to drop support for Ruby < 2.2.7 first.~


davidor:
I've deleted the commits related with Ruby updates that were merged in a different PR. I've also rebased on top of master adapting the code to the changes that were introduced in recent PRs (sentinels passwords, update of redis-rb, etc.).

davidor:
@unleashed This is ready for review now that we no longer need to support Ruby 2.2.

You already reviewed all the PRs of this integration branch except for this one https://github.com/3scale/apisonator/pull/92

I spent some time with @slopezz trying this in our Openshift cluster and it's working well.

Let me know what you think we need to be able to merge this. I think it's a very good starting point and we can keep improving later, for example, investigating alternative ways to fetch the jobs from the queue.

test/test_helpers/storage.rb
```diff
@@ -25,6 +28,12 @@ module Mock
       DEFAULT_NODES = ["127.0.0.1:7379", "127.0.0.1:7380"].freeze
       private_constant :DEFAULT_NODES
 
+      STORAGE_CLASSES = [
+          ThreeScale::Backend::StorageAsync::Client,
+          ThreeScale::Backend::StorageSync
+      ].freeze
```
unleashed: I'd like to be able to run the tests with sync without loading the async client and its dependencies. Not a big issue, but would be nice to have. Just referring to it here probably requires the interpreter to load up the dependencies.

test/test_helpers/storage.rb
```diff
@@ -34,9 +43,18 @@ def nodes
...
-        def mock_storage_client!
-          require '3scale/backend/storage_async'
-          class << ::ThreeScale::Backend::StorageAsync::Client
+        def mock_storage_clients
+          STORAGE_CLASSES.each { |klass| mock_storage_client!(klass) }
+        end
+
+        def unmock_storage_clients
+          STORAGE_CLASSES.each { |klass| unmock_storage_client!(klass) }
+        end
```
unleashed: This commit is a candidate for squashing with the one introducing the async client in tests.

test/test_helpers/configuration.rb
```diff
@@ -7,5 +7,5 @@
   config.redshift.port = 5432
   config.redshift.dbname = 'test'
   config.redshift.user = 'postgres'
-  config.redis.async = true
+  config.redis.async = ENV['CONFIG_REDIS_ASYNC'].to_s == 'true' ? true : false
```
unleashed: This can also be squashed.

script/test
```diff
@@ -19,7 +19,16 @@ function do_license_check()
...
-    bundle_exec rake "$@"
+    echo "Running tests using the $1 Redis driver"
+
+    case $1 in
+    sync)
+        CONFIG_REDIS_ASYNC=false bundle_exec rake
+        ;;
+    async)
+        CONFIG_REDIS_ASYNC=true bundle_exec rake
+        ;;
```
unleashed: Would benefit from a catch-all case in which it blows up as the argument is unknown (as opposed to not doing anything).

spec/unit/storage_async/pipeline_spec.rb
```diff
@@ -10,6 +10,11 @@ module StorageAsync
 
           subject { Pipeline.new }
 
+          # Spec helpers make sure to cleanup instances of Storage, but in this
+          # case, we are using the client directly with a host/port so we need
+          # this to call flushdb().
+          before { async_client.flushdb! }
```
unleashed: Squash! ;)

spec/spec_helper.rb
```diff
@@ -39,11 +39,11 @@ def committed_at
     require_relative '../test/test_helpers/configuration'
     require_relative '../test/test_helpers/storage'
 
-    TestHelpers::Storage::Mock.mock_storage_client!
+    TestHelpers::Storage::Mock.mock_storage_clients
```
unleashed: Not a fan of letting the detail that there are now two storage clients leak into the naming. Also, this is meant to use just one storage per invocation, isn't it?

lib/3scale/backend/worker.rb
```diff
@@ -30,7 +26,15 @@ def self.new(options = {})
           WorkerMetrics.start_metrics_server
         end
 
-        super
+        if options[:async]
+          # Conditional require is done to require async-* libs only when
+          # needed and avoid possible side-effects.
```
unleashed: Looks good - must ensure the same happens when loading the async storage in tests, or in general in the listeners when choosing just the sync one (I say this because I haven't checked it, but you probably know).

lib/3scale/backend/storage_async/client.rb
```diff
@@ -120,6 +120,18 @@ def initialize(opts)
...
+          call_args = ['BLPOP'] + args
+
+          # redis-rb accepts a Hash as last arg that can contain :timeout.
+          if call_args.last.is_a? Hash
+            timeout = call_args.pop[:timeout]
+            call_args << timeout
+          end
+
+          @redis_async.call(*call_args.flatten)
+        end
```
unleashed: squash?

test/test_helpers/configuration.rb
davidor: Yes, I wanted to keep the original commits of the PRs included in this integration branch so you could know what you reviewed already and what were your comments back then. Now that you've already reviewed this I can squash them to clean the history a bit.

test/test_helpers/storage.rb
davidor: I prefer to leave this one like it is.

script/test
davidor: :+1:

spec/unit/queue_storage_spec.rb
```diff
@@ -2,8 +2,10 @@
 
 module ThreeScale
   module Backend
-    describe QueueStorage do
-      describe "#connection" do
+    # Tests are run only when redis.async = false. These tests only check
+    # sentinels, which are not supported by the async client.
+    describe QueueStorage, if: false do
```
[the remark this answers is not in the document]
davidor: Right. I don't know how that ended up there :face_with_head_bandage: 
Fixed.

lib/3scale/backend/storage_async/pipeline.rb
```diff
@@ -23,12 +31,20 @@ def initialize
...
+          @fiber_id = Fiber.current.object_id
         end
 
         # In the async-redis lib, all the commands are run with .call:
         # client.call('GET', 'a'), client.call('SET', 'b', '1'), etc.
         # This method just accumulates the commands and their params.
         def call(*args)
+          if @fiber_id != Fiber.current.object_id
+            raise PipelineSharedBetweenFibers
+          end
```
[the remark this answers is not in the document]
davidor: * No. It didn't show up in any of my profiling results.
* No. We need this because pipelines are implemented a bit differently in the upstream. It's another case where async-redis does not provide the same interface as redis-rb. We'll see in the future if we can contribute that "compatibility-layer" to async-redis.

lib/3scale/backend/storage_async/async_redis.rb
```diff
@@ -8,12 +8,42 @@ class Client
...
+        # for performance reasons.
+        def write_request_without_flush(arguments)
+          write_lines("*#{arguments.count}")
+
+          arguments.each do |argument|
+            string = argument.to_s
+
+            write_lines("$#{string.bytesize}", string)
+          end
+        end
```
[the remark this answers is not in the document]
davidor: Yes, in master. I asked the maintainer to release a new version.

spec/spec_helper.rb
```diff
@@ -53,6 +54,19 @@ def committed_at
...
     ThreeScale::Backend::Memoizer.reset!
   end
+
+  config.after :each do
+    ThreeScale::Backend::Storage.instance.close
+  end
+
+  config.around :each do |example|
+    Async.run do
+      # TODO: This is needed for the acceptance specs. Not sure why.
```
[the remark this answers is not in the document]
davidor: I'm not even sure this is a bug, and it only happens in the acceptance specs, so it makes me think that the problem might be in `rspec_api_documentation`. I investigated a bit and didn't understand why.

lib/3scale/backend/storage_async/client.rb
```diff
@@ -0,0 +1,144 @@
...
+        #      For example, Redis returns 0 or 1 for the EXISTS command, but
+        #      redis-rb transforms that into a boolean.
+        #   3) There are a few methods that need to be treated differently and
+        #      do not fit in any of the previous categories. For example, SSCAN
+        #      which accepts a hash of options in redis-rb.
+        #
+        # All of this might be simplified a bit in the future using the
+        # "methods" in async-redis
+        # https://github.com/socketry/async-redis/tree/master/lib/async/redis/methods
+        # but there are some commands missing, so for now, that's not an option.
```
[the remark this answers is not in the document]
davidor: https://github.com/socketry/protocol-redis/issues/1

lib/3scale/backend/storage_async/async_redis.rb
```diff
@@ -0,0 +1,19 @@
+# Monkey-patches the async-redis lib to provide a 'call_pipeline' method that
+# sends multiple commands at once and returns an array of the responses for
+# each of them.
+
+module Async
+  module Redis
+    class Client
+      def call_pipeline(commands)
```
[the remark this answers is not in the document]
davidor: Same case. See comment above.

lib/3scale/backend/storage_async/pipeline.rb
```diff
@@ -0,0 +1,63 @@
...
+        def run(redis_async_client)
+          responses = redis_async_client.call_pipeline(@commands)
+
+          responses.zip(@commands).map do |resp, cmd|
+            command_name = cmd.first.to_s.upcase
+
+            if CHECK_EQUALS_ONE.include?(command_name)
+              resp.to_i == 1
+            elsif CHECK_GREATER_THAN_0.include?(command_name)
+              resp.to_i > 0
```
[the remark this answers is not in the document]
davidor: Same case. See comment above.

davidor:
@unleashed We have different views on what this PR should be.

This is a big change in how apisonator works. It's true that in our tests this increases performance substantially. However, this relies on the async-redis lib. The maintainers have done an amazing job in the lib and the async ecosystem in general. However, It's also true that the lib is not as widely used as redis-rb. Given that it's such an important piece in this project, there's some risk. I think we need to try this with real workloads to see how it behaves. I think it's important to discard problems such as unexpected crashes, memory leaks, etc. The kind of problems that you only realize when running for a long time on a real production scenario.

This feature is opt-in. For users absolutely nothing changes unless they explicitly decide that they want to give this as try. That's why I think we should try to merge this and let @slopezz and others experiment with it.

My opinion is that, rather than try to solve all the minor problems in the first PR, we should go step by step. Try a first version, see how it behaves, try different configuration params to evaluate performance in different environments, etc. If all that goes well, we can put the effort to, for example, contribute some of the code that we have here to other projects in the async ecosystem.

As I said before, I don't think we need to solve all the issues now. For example, contributing a redis-rb-compatible interface for pipelines in async-redis is not a trivial effort. This PR brings a substantial performance improvement to the table. Also, the rules to scale in environments like Openshift become way easier to understand. That's why I'm willing to go more in tech debt that in other cases, and that includes:

- I'm willing to add a line to make the acceptance tests pass: https://github.com/3scale/apisonator/pull/96#discussion_r317624586
- I think that having an adapter for the redis-async so we can use both redis-rb and redis-async at the same time is not a big deal. Sure, ideally that part of the code would be in redis-rb as a driver for redis-async or in redis-async as a dsl with a compatibility layer, but we don't have that right now. Also, having adapters like the one introduced in this PR is a common and widely accepted pattern to solve this kind of problem: https://github.com/3scale/apisonator/pull/96#discussion_r317629557
- I think that changing a lazy enumerator in a part of the code that's deprecated and that does not really benefit from that lazy enumerator is not a big issue: https://github.com/3scale/apisonator/pull/96#discussion_r317664008
- I'm also OK with running this test only with the sync driver https://github.com/3scale/apisonator/pull/96#discussion_r317661143

Interestingly, you didn't mention anything about running all the tests twice, which I think is the biggest problem that this PR introduces. Again, the ideal scenario would be to have redis-async - redis-rb compatibility tests in any of those projects. But that does no exist right now, and because of how our test suite works and its coupling with Redis, I don't feel confident running the tests just with async or sync and assuming that the other will work just fine.

I think that these trade-offs are reasonable, and we can document them so there are no surprises in the future. I just don't think that waiting to merge this is worth it.

There are other small issues that you mentioned that I'll address before merging, but I wanted to talk about the important things first.

ioquatix:
@davidor if you see any issues feel free to keep me in the loop.

davidor:
Rebased on top of master. I'll try to add a document with the known limitations tomorrow.

davidor:
@unleashed I added a document with the limitations, goals, and other info in my latest commit.

docs/async.md
```diff
@@ -0,0 +1,73 @@
...
+- `redis-rb` and `async-redis` have different interfaces for some methods and we
+want to be compatible with the two libraries at least for a while. That means
+that we need an adapter `lib/3scale/backend/storage_async/client.rb`. We could
+simplify that and other things in our codebase by contributing to the
+`async-redis` project.
+
+- We need this line `RSpec.current_example = example` in `config.around :each`
+of `spec_helper.rb` in order to make the acceptance tests pass.
+
+- There's one test of the suite that fails with the async client. It's in the
```
unleashed: This a symptom of some issue we think might be related to _threading_. So the limitation would be: potential issues with threading we haven't figured out yet.

docs/async.md
```diff
@@ -0,0 +1,73 @@
...
+client based on the configuration.
+
+- The majority of the code required to make async work is under
+`lib/3scale/backend/storage_async`. That includes a wrapper for the
+`async-redis` client, and some monkey-patching to optimize a few things. Some
+tests helpers also needed to be adapted to choose between the two types of
+clients.
+
+
+## Limitations
```
unleashed: There is no mention of what the async gem does, and it is a potential source of problems. In particular, it monkey-patches core IO modules to use a reactor, which means that interaction with any IO code is to be suspected in case of issues, and C extensions (usually) won't use the reactor for their IO (so we need to watch out for blocking there).

docs/async.md
```diff
@@ -0,0 +1,73 @@
...
+
+- The whole test suite needs to be run both with the async client and the sync
+one. The reason is that our tests are highly coupled with Redis, even the unit
+ones. Running the test suite with only one of the clients is risky.
+
+- `redis-rb` and `async-redis` have different interfaces for some methods and we
+want to be compatible with the two libraries at least for a while. That means
+that we need an adapter `lib/3scale/backend/storage_async/client.rb`. We could
+simplify that and other things in our codebase by contributing to the
+`async-redis` project.
```
unleashed: You are also missing the `avoid IO in the middle of a pipeline` limitation.

unleashed:
bors delegate+

docs/async.md
ioquatix: It doesn’t do that (monkey patching).

ioquatix:
What primitives are broken by async?

unleashed:
> What primitives are broken by async?

@ioquatix according to @davidor our usage of lazy enumerators and a test using two threads break when loading the async gems. He might be able to provide further details.

docs/async.md
unleashed: that was my understanding from when the first PR was introduced - are you not modifying the behaviour of `IO` methods?

I'd then assume async-redis is using async{,-io} on its own and then nothing else would be touching the reactor, right? That is preferable for us as it lowers the exposure to issues in unrelated code.

ioquatix: > that was my understanding from when the first PR was introduced - are you not modifying the behaviour of IO methods?

@unleashed we are not modifying IO. We add `Async::IO` which provides (almost) identical wrappers which you can inject into other code if needed.

> I'd then assume async-redis is using async{,-io} on its own and then nothing else would be touching the reactor, right? That is preferable for us as it lowers the exposure to issues in unrelated code.

At the moment, that's correct. However in the future, we may make it possible for the user to allow other forms of IO to use the reactor: https://github.com/ruby/ruby/pull/1870 - just FYI.

ioquatix:
> @ioquatix according to @davidor our usage of lazy enumerators and a test using two threads break when loading the async gems. He might be able to provide further details.

Lazy enumerators are buggy when combined with user fibers. It's a known issue: https://github.com/ruby/ruby/pull/2002

The solution is not to use async in code which can be used from an enumerator, until the bug/issue is resolved within Ruby itself.

Regarding threads, if you use thread level locks/mutex within async code, you might have strange behaviour, especially if you mix the reactor between threads. Reactors should be per-thread. If you need thread synchronisation, you need to use a higher level construct e.g. a pipe. Using thread synchronisation with an async reactor is blocking by definition.

unleashed:
@davidor that is not a lazy enumerator. There could be problems with our code or *dependencies* using fibers (some gems use fibers under the hood), but AIUI non-lazy enumerators should not be affected. The threading issue might be resolved by adapting the offending test, and the remaining problem is whether we want to live with these issues.

A closer look to the problems or a link to existing or new issues is what I was requesting in my review. Thanks for providing those, @ioquatix. Now we have more information, and it would be good to add it to the document on limitations before this would be merged.

davidor:
@unleashed I know it's not a lazy enum. I read the comments in the PR of the Ruby repo and it looked to me that other kind of enumerators have problems too, but maybe I'm missing something.

davidor:
I see. I misunderstood some things. My last comment is not correct.
I've updated the `async.md` with the new things we've learned.

docs/async.md
```diff
@@ -0,0 +1,84 @@
...
+GitHub](https://github.com/socketry) is very likely to block.
+
+- For now, this feature is only implemented in the workers. The idea is to
+enable it in the listeners too after we verify that everything works well.
+
+- The whole test suite needs to be run both with the async client and the sync
+one. The reason is that our tests are highly coupled with Redis, even the unit
+ones. Running the test suite with only one of the clients is risky.
+
+- Cannot use `Fiber.yield` in an enumerator. See [ruby PR
```
unleashed: I would specifically call out lazy enumerators, because that is where you are most likely to inadvertently hit this problem.

ioquatix: It's only lazy enumerators that allocate a fiber. Some lazy enumerators can avoid it. The typical situation is when you invoke zip, the argument will use a fiber internally.

davidor:
@unleashed and I have decided to merge this.
The feature is opt-in and all the limitations are documented in `docs/async.md`.

davidor:
bors r=@unleashed

PR #101, opened 2019-06-07
Add support for password-protected sentinels

davidor:
Ref: https://issues.jboss.org/browse/THREESCALE-2620

### DIFF
--- a/lib/3scale/backend/storage.rb
+++ b/lib/3scale/backend/storage.rb
@@ -174,8 +174,8 @@ module ThreeScale
           #   "redis_url0,redis_url1,redis_url2,....,redis_urlN"
           # Or an Array of Strings representing one URL each:
           #   ["redis_url0", "redis_url1", ..., "redis_urlN"]
-          # Or an Array of Hashes with ":host" and ":port" keys:
-          #   [{ host: "srv0", port: 7379 }, { host: "srv1", port: 7379 }, ...]
+          # Or an Array of Hashes with ":host", ":port", and ":password" (optional) keys:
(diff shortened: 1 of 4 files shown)


lib/3scale/backend/storage.rb
```diff
@@ -174,8 +174,8 @@ def ensure_url_param(options)
           #   "redis_url0,redis_url1,redis_url2,....,redis_urlN"
           # Or an Array of Strings representing one URL each:
           #   ["redis_url0", "redis_url1", ..., "redis_urlN"]
-          # Or an Array of Hashes with ":host" and ":port" keys:
-          #   [{ host: "srv0", port: 7379 }, { host: "srv1", port: 7379 }, ...]
+          # Or an Array of Hashes with ":host", ":port", and ":password" (optional) keys:
```
unleashed: I assume sentinesl don't have a `:user`?

davidor: They don't https://redis.io/topics/sentinel#configuring-sentinel-instances-with-authentication

davidor:
It turns out that there was a bug in redis about this: https://github.com/redis/redis-rb/pull/856
They included the fix in redis-rb 4.1.2, but that version also drops support for Ruby < 2.3.0, so we cannot merge this.

davidor:
Added a new commit that monkey patches the Sentinel client class to fix the issue without upgrading the gem.

davidor:
Created an issue so we don't forget about this: https://github.com/3scale/apisonator/issues/102

PR #103, opened 2019-06-11
Introduce new limit header with the new quota after the limit period ends

davidor:
This PR introduces a new limit header, "3scale-limit-value-after-reset" that returns the new quota after the current limiting period ends.

This is needed by APIcast. When the request is limited, APIcast needs to know whether it was because the limit was 0 (which means that the metric is disabled). APIcast needs to do so without parsing the body, so it relies on the limit headers. However, using the current ones there's no way to know if the limit is 0.

Returning a boolean instead of the new quota is another option, but I thought that returning the new quota gives more information without adding any complexity to the code.

Refs:
[THREESCALE-2756](https://issues.jboss.org/browse/THREESCALE-2756)
[THREESCALE-2755](https://issues.jboss.org/browse/THREESCALE-2755)

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -55,6 +55,8 @@ non-existing metrics occur):
 * `3scale-limit-reset`: An integer stating the amount of seconds left for the
   current limiting period to elapse. A negative integer value means there is no
   limit in time.
+* `3scale-limit-max-value`: An integer stating the maximum total number of hits
+  allowed in the current limiting period.
 
 When a `usage` is specified, only the metrics specified in that usage and their
 parent metrics are taken into account when calculating the limit headers.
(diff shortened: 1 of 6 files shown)


davidor:
Makes  sense. Renamed :+1:

docs/extensions.md
```diff
@@ -55,6 +55,8 @@ non-existing metrics occur):
 * `3scale-limit-reset`: An integer stating the amount of seconds left for the
   current limiting period to elapse. A negative integer value means there is no
   limit in time.
+* `3scale-limit-max-value`: An integer stating the quota after the current
```
unleashed: `s/quota after the current limiting period ends/maximum number of hits allowed in this limiting period/`

unleashed:
r=me when the docs are fixed.

docs/extensions.md
davidor: That's better :+1: 
Fixed.

PR #107, opened 2019-07-17
openshift/Dockerfiles: apply bundler fix for .so dir

davidor:
The problem still happens in the bundler version included in rh-scl 2.5

### DIFF
--- a/openshift/distro/centos/7/Dockerfile
+++ b/openshift/distro/centos/7/Dockerfile
@@ -111,12 +111,11 @@ RUN source scl_source enable ${RUBY_SCL} \
       yum -y install ${BUILD_DEP_PKGS} ; \
     fi \
  && bundle install --deployment --jobs $(grep -c processor /proc/cpuinfo) \
-# Bundler < 1.12.0 needs fixing for git gems with extensions (Puma)
- && if ruby -e "begin; require 'rubygems'; Gem::Specification.find_by_name('bundler').version < Gem::Version.new('1.12.0') && exit(0); rescue; end; exit(1)"; then \
-      echo "[WARNING] Old Bundler requires fixing git gems with extensions (Puma)" ; \
-      PUMA_DIR=$(bundle show puma) \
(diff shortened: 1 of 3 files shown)


openshift/distro/centos/7/Dockerfile
```diff
@@ -111,12 +111,11 @@ RUN source scl_source enable ${RUBY_SCL} \
...
     fi \
  && bundle install --deployment --jobs $(grep -c processor /proc/cpuinfo) \
-# Bundler < 1.12.0 needs fixing for git gems with extensions (Puma)
- && if ruby -e "begin; require 'rubygems'; Gem::Specification.find_by_name('bundler').version < Gem::Version.new('1.12.0') && exit(0); rescue; end; exit(1)"; then \
-      echo "[WARNING] Old Bundler requires fixing git gems with extensions (Puma)" ; \
-      PUMA_DIR=$(bundle show puma) \
-      && ln -s $(find "${PUMA_DIR}" -name puma_http11.so) "${PUMA_DIR}"/lib/puma ; \
-    fi \
+
+ ## bundler does not place the .so in the correct dir
```
unleashed: just one `#` is enough ;)

unleashed:
which bundler version is this?

davidor:
1.16.1

openshift/distro/centos/7/Dockerfile
unleashed: this is rather weird - something from the ruby configuration, rubygems or bundler (or even puma using some wrong directory?) might be changing from a local installation (non-RHSCL) which I can't reproduce

davidor: I can't reproduce it either on a non-rhscl env.

unleashed:
bors flip-flip-all-the-tables

unleashed:
@davidor we should probably dig a bit to find out the cause of this and open an issue somewhere... as time permits.

davidor:
Sure, it was important to merge this because the listener cannot even boot after the ruby upgrade using these images. Let's try to understand why that bundler version is broken later.

PR #111, opened 2019-07-31
Prometheus metrics for workers

davidor:
This PR adds Prometheus metrics for workers.

It uses the [yabeda-prometheus gem](https://github.com/yabeda-rb/yabeda-prometheus). It basically creates a web service on a separate thread that exposes `/metrics`.

This feature is opt-in. There are a couple of new config params. One for enabling the feature (`ENV['CONFIG_WORKER_PROMETHEUS_METRICS_ENABLED']`) and another one for selecting the port for the metrics web service (`ENV['CONFIG_WORKER_PROMETHEUS_METRICS_PORT']`).

For now, I added a couple of metrics (see `docs/prometheus_metrics.md`):

```
| Metric                                | Description                        | Type      | Labels                           |
|---------------------------------------|------------------------------------|-----------|----------------------------------|
| apisonator_worker_job_count           | Number of jobs processed           | counter   | type(ReportJob, NotifyJob, etc.) |
| apisonator_worker_job_runtime_seconds | How long the jobs take to complete | histogram | type(ReportJob, NotifyJob, etc.) |
```

/cc @slopezz

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,10 +36,10 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - rbenv-dependencies-{{ checksum ".ruby-version" }}
+          - v1-rbenv-dependencies-{{ checksum ".ruby-version" }}
           - v1-dependencies-{{ checksum "Gemfile.lock" }}
           # fallback to using the latest cache if no exact match is found
-          - rbenv-dependencies-
+          - v1-rbenv-dependencies-
           - v1-dependencies-
 
       - run:
@@ -50,7 +50,7 @@ jobs:
       - save_cache:
           paths:
             - ~/.rbenv
-          key: rbenv-dependencies-{{ checksum ".ruby-version" }}
+          key: v1-rbenv-dependencies-{{ checksum ".ruby-version" }}
 
       - run:
           name: bundle install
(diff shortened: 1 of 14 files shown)


miguelsorianod:
Hi,

The code looks good.
I have some comments about how metrics are calculated with Yabeda that I think should be taken into account when generating the dashboards for our saas/useful information for our on-premises customers:

I understand that the way it works is that we are using approach 2 from here https://github.com/yabeda-rb/yabeda-prometheus#usage. This is, another thread is started and somehow when you report metric data all is aggregated from there. This works well in pre-fork environments because all the data is aggregated before being calculated.
It is interesting to have this kind of data at a worker level, but another interesting layer is to have metrics is at "service" level. Our environment can also have multiple machines running the workers. This means that another layer of aggregation has to be added on Prometheus server side. I had serious doubts about that being possible but I see you have used histograms and not summaries, and histograms seem to be aggregable on prometheus server-side if I understand the following documentation correctly https://prometheus.io/docs/practices/histograms/. So, we should make sure that we perform counter and histogram aggregation on Prometheus server-side so we have metrics at the service level and not only at the instance level.

slopezz:
Yes, information will be obtained at individual pod level, but in order to create useful grafana dashboards and prometheus alerts, we will aggregate pod data.

Then, in order to add monitoring at pod level with prometheus-operator, it will be needed to be created:
- Headless metrics Service object (to send all service requests to each pod, and not do any kind of load balacing) pointing to DeploymentConfig object.
- ServiceMonitor object pointing to previous Headless Service

davidor:
@miguelsorianod As you mentioned, histograms can be aggregated without problems. It's just a matter of adding the values for each bucket across all the instances. I had some doubts about which buckets to choose, but we can tune that later.

davidor:
bors r=@miguelsorianod

PR #120, opened 2019-09-02
Fix bug that makes worker crash when fetching a job

davidor:
This reverts commit bacb0ba3507ceba75987ca232d2606296b825fb5.

This commit made the worker crash after fetching a job from the queue. The reason is that `Array#blank?` is not defined in our code.

Our tests did not catch this because one of the dependencies that we only use in tests requires `activesupport`, which defines `Array#blank?`


davidor:
No, just for String and nil.

PR #128, opened 2019-09-19
Drop support for end-users

davidor:
Support for end-users was deprecated some releases ago.
Now, the product team has confirmed that we can drop support for the feature.

@unleashed The diff of this PR is quite big. End-users were present in many parts of the codebase. Here's how I organized it:
- Internal API
- Code
- Oauth code: dropping support in oauth tokens was not as straightforward as in the rest of the code so I decided to separate it.
- One commit for each of the test types.
- Update of active docs


lib/3scale/backend/stats/keys.rb
```diff
@@ -22,12 +22,6 @@ def applications_key_prefix(prefix)
           "#{prefix}/cinstances"
         end
 
-        # @note For backwards compatibility, the key is called uinstance.
-        # It will be eventually renamed to user.
```
unleashed: Never did. :)

lib/3scale/backend/stats/keys.rb
```diff
@@ -51,37 +45,21 @@ def application_usage_value_key(service_id, app_id, metric_id, period)
...
-
         def service_response_code_value_key(service_id, response_code, period)
           service_key        = service_key_prefix(service_id)
           response_code_key  = response_code_key_prefix(service_key, response_code)
 
           encode_key(counter_key(response_code_key, period))
         end
 
-       def application_response_code_value_key(service_id, app_id, response_code, period)
+        def application_response_code_value_key(service_id, app_id, response_code, period)
```
unleashed: extra whitespace?

lib/3scale/backend/transactor.rb
```diff
@@ -58,34 +58,23 @@ def validate(oauth, provider_key, report_usage, params, extensions)
...
         params[:app_id] = nil if app_id && app_id.empty?
-        params[:user_id] = nil if user_id && user_id.empty?
-
-        # Now OAuth tokens also identify users, so must check tokens anyway if
-        # at least one of app or user ids is missing.
-        #
-        # NB: so what happens if we call an OAuth method with user_key=K and
-        # user_id=U? It is effectively as if app_id was given and no token would
-        # need to be checked, but we do... That is not consistent. And madness.
-        # Each time I try to understand this I feel I'm becoming dumber...
```
unleashed: A piece of wisdom unfortunately gone forever.

lib/3scale/backend/stats/partition_eraser_job.rb
```diff
@@ -11,13 +11,12 @@ class << self
           include StorageHelpers
           include Configurable
 
-          def perform_logged(_enqueue_time, service_id, applications, metrics, users,
+          def perform_logged(_enqueue_time, service_id, applications, metrics,
```
unleashed: This is dangerous. How are we going to run pending jobs from an upgrade?

We should make the jobs' signatures callable with newer and older enqueued jobs for a while, otherwise this will break when upgrading from a previous version.

docs/oauth_tokens.md
```diff
@@ -100,8 +83,7 @@ This is the feature that requires walking the token set.
 > support anywhere in the codebase to avoid blocking the database.
 
 When we list the tokens for a given service and application, we want to list
-*all* of them, that is, including user-specific tokens. We also support listing
-those that apply to a single user alone.
+*all* of them.
```
unleashed: This sentence makes no sense now.

spec/integration/stats_partition_eraser_job_spec.rb
```diff
@@ -31,7 +29,7 @@
   it 'partition keys are deleted' do
     without_resque_spec do
       Resque.enqueue(described_class, Time.now.getutc.to_f, service_id, applications,
-                     metrics, users, from, to, offset, length, nil)
+                     metrics, from, to, offset, length, nil)
```
unleashed: We need to test that enqueuing the old job (with a users param) works with the new job signature.

lib/3scale/backend/stats/keys.rb
davidor: No, it was wrong before.

lib/3scale/backend/stats/partition_eraser_job.rb
davidor: There are no partition_eraser jobs running. We disabled them for performance reasons.

lib/3scale/backend/stats/keys.rb
davidor: :trollface:

docs/oauth_tokens.md
davidor: Someone could assume that there's pagination :eyes:

spec/integration/stats_partition_eraser_job_spec.rb
davidor: Same as https://github.com/3scale/apisonator/pull/128#discussion_r326092663

davidor:
I think that what you mentioned about rejecting any request containing a user_id is a good idea.

There should not be anyone using this, but just in case, if we detect that, we can alerts users so they can check their integration settings.

andrewdavidmackenzie:
can we look at saas logs to see how many requests are using that parameter now?

docs/oauth_tokens.md
unleashed: The sentence still doesn't make sense. Just delete it or add one about not having pagination.

lib/3scale/backend/stats/partition_eraser_job.rb
unleashed: No any other job affected? IIRC there is at least another one `perform_logged`. If also not enabled then ok to go.

docs/oauth_tokens.md
davidor: :+1:

lib/3scale/backend/transactor.rb
davidor: It will always be in the git history :)

lib/3scale/backend/stats/partition_eraser_job.rb
davidor: It should be fine. `ReportJob` and `NotifyJob` receive transactions. There isn't a `user_id` param.

davidor:
@unleashed the last 2 commits raise an error when the call contains a user ID.

@andrewdavidmackenzie Nobody was using this when we checked some time ago. However, yesterday, I checked the logs of the previous day, just be sure, and it turns out that there were _6_ calls from the same user. I spoke with @andreumasferrer to decide how we should proceed.

lib/3scale/backend/errors.rb
```diff
@@ -333,5 +333,11 @@ def initialize(service_id, msg)
         super "Delete stats job context validation error. Service: #{service_id}. Error: #{msg}"
       end
     end
+
+    class EndUsersNoLongerSupported < BadRequest
+      def initialize
+        super 'End-users are no longer supported'.freeze
```
unleashed: `"[...], do not specify the user_id parameter"`?

davidor: Done :+1:

unleashed:
r=me when @andreumasferrer comes up with the decision

davidor:
I spoke with @andreumasferrer . We can go ahead.

davidor:
bors r=@unleashed

PR #129, opened 2019-09-25
Worker metrics tests: check if Errno::ENETUNREACH is raised

davidor:
This PR fixes one of the tests of the worker metrics server.

We've seen that both `Errno::ENETUNREACH` and `Errno::EADDRNOTAVAIL` can be raised depending on the environment, but only the first was being considered.

/cc @gsaslis

### DIFF
--- a/spec/integration/worker_metrics_spec.rb
+++ b/spec/integration/worker_metrics_spec.rb
@@ -96,7 +96,7 @@ apisonator_worker_job_runtime_seconds_count{type="ReportJob"} \d+\.\d+
           sleep(1) # The web server takes a bit to start
 
           expect { Net::HTTP.get('localhost', metrics_endpoint, metrics_port) }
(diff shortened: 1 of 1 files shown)


gsaslis:
@davidor / @unleashed i'm not going into the details, but if this is going to take a few days, could I suggest we delete this test for now and get the tests passing again, while we work on this in the background? (this is kind of blocking me right now - and my workaround would be to completely skip tests, which i would like to avoid)

unleashed:
@gsaslis up to @davidor - we can easily go two ways: either remove the test for now or widen the kind of errors caught to something like `SystemCallError`. I believe the latter would provide a better compromise than no tests or just catching specific instances of `errno`. Later we can think about how to have better guarantees than `sleep(1)`.

davidor:
I've changed it check `SystemCallError`.

davidor:
bors r=@gsaslis

PR #133, opened 2019-10-10
Update rubyzip to v2.0.0

davidor:
Fixes https://nvd.nist.gov/vuln/detail/CVE-2019-16892

Does not affect in production because this is a dependency of license-finder


unleashed:
it is a major update rather than a CVE fix

davidor:
I know it's a major update, but it's not a big deal. It's a dependency used only by license finder and it works well after updating rubyzip.

PR #137, opened 2019-10-15
Make Rake tasks compatible with the async redis client

davidor:
There are a couple of Rake tasks that were not working correctly when using the async redis client: reschedule of failed jobs and the connectivity checks.


davidor:
bors r=@miguelsorianod

PR #141, opened 2019-11-06
Support docker-compatible CLIs

unleashed:
This lets you define the DOCKER variable to specify a docker
compatible CLI, and in addition defaults to looks up podman in
$PATH, falling back to docker otherwise, either by full path or by
the plain "docker" name.


davidor:
Did you upgrade to Fedora 31? :)

unleashed:
Yes, but that has nothing to do with this patch.

bors r=@davidor

PR #143, opened 2019-12-17
Alternative approach to delete the stats keys of a service

davidor:
This PR implements an alternative approach to delete the stats keys of a service.

The summary is as follows: when backend receives a request to delete the stats of a service, instead of deleting them right away, it stores the service ID in a Redis set so it can be deleted later. There's a rake task that fetches the IDs in that set and deletes all the keys belonging to those services in all the Redis servers received via params.

I've documented all the design decisions and the implementation details in the new `Stats::Cleaner` class. I've also documented how to use this feature in `docs/delete_stats.md`.


lib/3scale/backend/stats/cleaner.rb
```diff
@@ -0,0 +1,156 @@
...
+        # request to delete the stats of a service, we generated all the
+        # possible stats keys that could exist for it. That approach was not
+        # efficient because it ended up generating many keys that didn't exist
+        # and thus, unnecessary delete calls to Redis. That approach is also
+        # more complex and error prone. You can find the details here:
+        # https://github.com/3scale/apisonator/issues/90
+
+        include Storable
+
+        SET_WITH_SERVICES_MARKED_FOR_DELETION = 'set_with_services_marked_for_deletion'.freeze
```
unleashed: Well, the actual name could be just `service_to_delete` or something shorter and simpler, since in the end _if_ you go inspecting the actual key you might as well know (or query) its type.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -0,0 +1,156 @@
...
+        private_constant :SET_WITH_SERVICES_MARKED_FOR_DELETION
+
+        SLEEP_BETWEEN_SCANS = 0.01 # In seconds
```
unleashed: Make these a default fallback and allow users to specify them (sleep, slice size, ...)?

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -0,0 +1,156 @@
...
+          # Note 1: keys deleted cannot be restored.
+          # Note 2: this method can take a long time to finish as it needs to
+          # scan all the keys in several DBs.
+          def delete!(redis_conns, log_deleted_keys: false)
+            services = services_to_delete
+            logger.info("Going to delete the stats keys for these services: #{services.to_a}")
+
+            redis_conns.each do |redis_conn|
+              delete_keys(redis_conn, services, log_deleted_keys)
+            end
```
unleashed: We could run this in parallel for all the instances. Given how much it takes and all the sleeping, I think it's something we would want to have.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -0,0 +1,156 @@
...
+            loop do
+              cursor, keys = redis_conn.scan(cursor, count: SCAN_SLICE)
+
+              to_delete = keys.select { |key| delete_key?(key, services) }
+
+              unless to_delete.empty?
+                if log_deleted_keys
+                  values = redis_conn.mget(*(to_delete.to_a))
+                  to_delete.each_with_index do |k, i|
+                    puts "#{k} #{values[i]}"
```
unleashed: Have `log_deleted_keys` be an `IO` object where you'd `log_deleted_keys.puts "..."`?

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -0,0 +1,156 @@
...
+                  end
+                end
+
+                redis_conn.del(to_delete) unless to_delete.empty?
```
unleashed: `to_delete` is guaranteed to not be empty here

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -0,0 +1,156 @@
...
+
+          def delete_keys(redis_conn, services, log_deleted_keys)
+            cursor = 0
+
+            loop do
+              cursor, keys = redis_conn.scan(cursor, count: SCAN_SLICE)
+
+              to_delete = keys.select { |key| delete_key?(key, services) }
+
+              unless to_delete.empty?
```
unleashed: You can mix this with the above with `tap do |to_delete|` (or a few other options) and remove the `to_delete` assignment.

spec/unit/stats/cleaner_spec.rb
```diff
@@ -0,0 +1,142 @@
+require_relative '../../spec_helper'
+
+module ThreeScale
+  module Backend
+    module Stats
+      describe Cleaner do
+        let(:storage) { Backend::Storage.instance }
```
unleashed: Would be nice to be able to mock Redis here, especially since these are unit tests (it's ok).

lib/3scale/backend/stats/cleaner.rb
davidor: I prefer to start with something simple. I'll leave that for a future PR. I suspect most users only have one shard anyway.

spec/unit/stats/cleaner_spec.rb
davidor: Right. But we are not doing this in most of the other unit tests that we have, which in my opinion is something that we should solve, but not in this PR.

lib/3scale/backend/stats/cleaner.rb
davidor: In my opinion, it comes down to personal preference. I don't see any obvious reason why any of those 2 approaches is better than the other.

lib/3scale/backend/stats/cleaner.rb
davidor: Right. Fixed :+1:

lib/3scale/backend/stats/cleaner.rb
davidor: Fixed :+1:

lib/3scale/backend/stats/cleaner.rb
unleashed: Less cognitive load and lower complexity. One assignment and a conditional (negated, slightly more complex) vs enum chaining with fewer lines and no conditionals. The error you made, not being sure the collection was or not empty is an indication that this is more error prone, just like writing manual loops vs using a functional style.

That said, up to you, no big deal.

lib/3scale/backend/stats/cleaner.rb
unleashed: @davidor what about this?

spec/unit/stats/cleaner_spec.rb
unleashed: It's ok 👍

lib/3scale/backend/stats/cleaner.rb
davidor: The defaults here are similar to the ones that we use in the rest of the codebase. I don't know if moving this responsibility to the user is the right decision.

unleashed: They are fallbacks, to align to what's used elsewhere but giving the user the opt-in possibility (rather than the forced responsibility) of overriding them. Not a blocker.

lib/3scale/backend/stats/cleaner.rb
davidor: Done :+1:

lib/3scale/backend/stats/cleaner.rb
davidor: The conditional is also needed when using `tap`

lib/3scale/backend/stats/cleaner.rb
davidor: I propose to address this in a future PR.
First, let's see how this behaves in a real use case scenario. Let's see how often we get requests to delete stats, how long it takes for each shard, etc, and then come back to this, adjust those params if needed, and offer some options to tune them. What do you think?

davidor:
Looks like bors is not working today :crying_cat_face:

PR #145, opened 2019-12-20
Monkey-patch Resque.enqueue when using async

davidor:
The async lib does not work well with the Resque gem. It crashes when running a pipeline in the enqueue method. This PR mokey-patches that method when using async.

I tried running the listeners with [Falcon](https://github.com/socketry/falcon) and the change introduced in this PR is the only one needed to make it work. After applying this change, the listeners can be executed with Falcon and the async redis lib with:
`CONFIG_REDIS_ASYNC=true bundle exec falcon serve -b http://0.0.0.0:3000 -c config.ru -n 4`


davidor:
@unleashed added :+1:

davidor:
Looks like bors is going to time-out again. As this is based on top of master, it can be safely merged.

PR #147, opened 2020-01-08
docs/rfcs/error_responses: clarify exception with 409 auth failed

### LINKED ISSUE #146: Documented behaviour not consistent for rate limiting
In the [error responses document](https://github.com/3scale/apisonator/blob/master/docs/rfcs/error_responses.md#currently-known-error_codes-and-proposed-classification), it suggest thats backend will return a `409` when rate limits have been exceeded which is behaviour I am observing. However it also suggests that the `error_code` tag in the xml response will be populated with the code `limits_exceeded` which I have not been able to observe.

This is the response I am seeing when an imposed limit has been breached:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<status>
  <authorized>false</authorized>
  <reason>usage limits are exceeded</reason>
  <plan>app-plan</plan>
  <usage_reports>
    <usage_report metric="hits" period="minute">
      <period_start>2020-01-06 14:52:00 +0000</period_start>
      <period_end>2020-01-06 14:53:00 +0000</period_end>
      <max_value>2</max_value>
      <current_value>2</current_value>
    </usage_report>
  </usage_reports>
</status>
```

unleashed:
`409 Conflict` responses don't specify the `error_code` and `message` fields. They instead contain a `reason` field and potentially the code of the "error" via an extension in the `3scale-rejection-reason` header.

Fixes #146.

### DIFF
--- a/docs/rfcs/error_responses.md
+++ b/docs/rfcs/error_responses.md
@@ -74,9 +74,7 @@ Backend responses:
 * `409 Conflict`:
   - The request's reported usage goes over the limits set for that specific
   application or user or some other condition for this application (not enabled
-  or invalid/missing key) or service (not enabled or not OAuth-enabled). Note that *this is not
-  an application or client error* but the reporting of a normal request telling
-  the client that the usage exceeds limits.
+  or invalid/missing key) or service (not enabled or not OAuth-enabled).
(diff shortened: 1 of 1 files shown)


davidor:
bors r=@PhilipGough,@davidor

PR #149, opened 2020-01-17
Add instructions to configure HPA on openshift

davidor:
This PR adds a document that explains how to setup Horizontal pod Autoscaling (HPA) in Openshift.

### DIFF
--- a/docs/openshift_horizontal_scaling.md
+++ b/docs/openshift_horizontal_scaling.md
@@ -1,0 +1,109 @@
+# Openshift Horizontal Pod Autoscaling (HPA)
+
+It is possible to configure Apisonator to automatically scale when deployed in
+Openshift. This document details the required setup.
+Note that this guide focuses on Openshift, but with minimal changes, it's
+possible to have the same setup on Kubernetes. You can learn about Kubernetes
+HPA
+[here](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/).
+
+## Pre-requisites
+
+- Have a working installation of 3scale in an Openshift cluster. You can use [this
+guide](https://github.com/3scale/3scale-operator/blob/master/doc/template-user-guide.md).
+
+## Openshift HPA
+
+Openshift Horizontal Pod Autoscaling (HPA) allows to scale pods horizontally by
+specifying the minimum and maximum number of pods that should be deployed and
+the target CPU utilization. For more details check the [official Openshift
+docs](https://docs.openshift.com/container-platform/4.1/nodes/pods/nodes-pods-autoscaling.html).
+
+## Async
+
+Apisonator can be configured to use a non-blocking Redis client. With this
+feature enabled, a single process should use a whole CPU core when properly
+tuned. This means that we can enable Openshift HPA using the CPU utilization as
+the scaling metric.
+
+To learn more about how the async Redis client works, check this [design
+doc](./async.md).
+
+## Setup
+
+### Redis databases
+
+The standard 3scale deployment uses a single Redis pod and sets both the resque
+DB and the data DB in the same Redis process under different DB indexes. This is
+not supported by the async redis library, so we need to either place both DBs in
+the same Redis process or use two separate ones. To configure this, change the
+`CONFIG_REDIS_PROXY` and `REDIS_QUEUES_URL` envs in the listener and worker
+deployment configs.
+
+### Enable async
+
+- Set the `CONFIG_REDIS_ASYNC` to true in the listener and worker deployment configs.
+- Configure listeners to use the Falcon web-server instead of Puma. In order to
+do this, modify the command in the listener deployment config. Replace this:
+```yaml
+spec:
+  containers:
+    - args:
+        - bin/3scale_backend
+        - start
+        - '-e'
+        - production
+        - '-p'
+        - '3000'
+        - '-x'
+        - /dev/stdout
+```
+
+with:
+```yaml
+spec:
+  containers:
+    - args:
+        - bin/3scale_backend
+        - '-s'
+        - falcon
+        - start
+        - '-e'
+        - production
+        - '-p'
+        - '3000'
+        - '-x'
+        - /dev/stdout
+```
+- Set the `LISTENER_WORKERS` env to 1 in the listener deployment. This will
+deploy 1 Falcon worker per pod.
+
+### Set up horizontal pod autoscaling
+
+We need to decide the target CPU usage. It is defined as a percentage of the
+requested CPU cores. In a default installation, the backend-listener pod
+requests 0.5 CPU cores. In an ideal case, we know that both the listener and the
+worker should use 1 CPU core when using the async redis lib. However, we need to
+leave some margin for possible inefficiencies, so we can target 0.8 CPU cores,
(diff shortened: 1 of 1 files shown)


docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
...
+oc autoscale dc/backend-listener --min 1 --max 10 --cpu-percent=160
+oc autoscale dc/backend-worker --min 1 --max 10 --cpu-percent=533
+```
+
+To check the decisions made by the auto-scaler as well as information about the
+current number of pods vs the desired number:
+```bash
+oc describe hpa backend-listener
+oc describe hpa backend-worker
+```
```
andrewdavidmackenzie: Add a little bit about seeing it work when testing, load increases?

docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
...
+not supported by the async redis library, so we need to either place both DBs in
+the same Redis process or use two separate ones. To configure this, change the
+`CONFIG_REDIS_PROXY` and `REDIS_QUEUES_URL` envs in the listener and worker
+deployment configs.
+
+### Enable async
+
+- Set the `CONFIG_REDIS_ASYNC` to true in the listener and worker deployment configs.
+- Configure listeners to use the Falcon web-server instead of Puma. In order to
+do this, modify the command in the listener deployment config. Replace this:
```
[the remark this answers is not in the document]
davidor: I agree. In the future, this should be implemented in the operator, but I wanted to document all the apisonator-specific knowledge here and validate the idea first.

docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Setup
+
+### Redis databases
+
+The standard 3scale deployment uses a single Redis pod and sets both the resque
+DB and the data DB in the same Redis process under different DB indexes. This is
+not supported by the async redis library, so we need to either place both DBs in
+the same Redis process or use two separate ones. To configure this, change the
+`CONFIG_REDIS_PROXY` and `REDIS_QUEUES_URL` envs in the listener and worker
```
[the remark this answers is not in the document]
davidor: This variable was introduced a long time ago. We can discuss ENV name changes in a separate PR because they require other changes to ensure backwards compatibility.

docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
...
+        - '-x'
+        - /dev/stdout
+```
+- Set the `PUMA_WORKERS` env to 1 in the listener deployment. This will deploy 1
+Falcon worker per pod.
+
+### Set up horizontal pod autoscaling
+
+We need to decide the target CPU. It is defined as a percentage of the requested
+CPU. In a default installation, the backend-listener pod requests 0.5 CPUs. In
```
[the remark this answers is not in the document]
davidor: I added some changes to clarify this.

docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
+# Openshift Horizontal Pod Autoscaling (HPA)
+
+It is possible to configure Apisonator to auto scale when deployed in Openshift.
```
unleashed: nit:
```suggestion
It is possible to configure Apisonator to automatically scale when deployed on Openshift.
```

unleashed: perhaps mention this can also be done in bare k8s?

docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
...
+        - falcon
+        - start
+        - '-e'
+        - production
+        - '-p'
+        - '3000'
+        - '-x'
+        - /dev/stdout
+```
+- Set the `PUMA_WORKERS` env to 1 in the listener deployment. This will deploy 1
```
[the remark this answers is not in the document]
unleashed: How about introducing `LISTENER_WORKERS`, look it up with higher precedence in the code, and if not present fall back to `PUMA_WORKERS`?

docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
...
+DB and the data DB in the same Redis process under different DB indexes. This is
+not supported by the async redis library, so we need to either place both DBs in
+the same Redis process or use two separate ones. To configure this, change the
+`CONFIG_REDIS_PROXY` and `REDIS_QUEUES_URL` envs in the listener and worker
+deployment configs.
+
+### Enable async
+
+- Set the `CONFIG_REDIS_ASYNC` to true in the listener and worker deployment configs.
+- Configure listeners to use the Falcon web-server instead of Puma. In order to
```
unleashed: This depends on #150.

docs/openshift_horizontal_scaling.md
```diff
@@ -0,0 +1,105 @@
...
+We need to decide the target CPU usage. It is defined as a percentage of the
+requested CPU cores. In a default installation, the backend-listener pod
+requests 0.5 CPU cores. In an ideal case, we know that both the listener and the
+worker should use 1 CPU core when using the async redis lib. However, we need to
+leave some margin for possible inefficiencies, so we can target 0.8 CPU cores,
+for example. 0.8/0.5 = 1.6, so we need to define a CPU % of 160 for the
+listener. In a default installation, the backend-worker pod requests 0.150 CPU
+cores. Following the same logic, we need to define a target CPU of 533
+(0.8/0.150 = 5.33). Pay attention to the CPU request of each pod, as they might
+change between versions, and adapt the target CPU % values accordingly.
```
unleashed: So here comes the logical question: why would we want to scale so much more the workers CPU demand (to 0.8) when we only request initially 0.150 millicores (when compared to the listeners which start way higher)? We could add something, even if a little vague, about why this is the case.

davidor: Even if you request only 0.150, you can ask for a limit of 1.

davidor: In my opinion, it's a bit confusing to define the target CPU as a function of the pod request. It'd be more intuitive to just define the number of cores.

docs/openshift_horizontal_scaling.md
davidor: Yes, go approve it :smile:

docs/openshift_horizontal_scaling.md
unleashed: And the docs are not helping. Will look them up in k8s docs and see if there is any clearer explanation.

docs/openshift_horizontal_scaling.md
davidor: Done :+1:

davidor:
@unleashed I added `LISTENER_WORKERS` in the other PR (#150) as you suggested.

davidor:
@andrewdavidmackenzie I've addressed your comments.

@unleashed I have replaced `PUMA_WORKERS` with `LISTENER_WORKERS`. Regarding `--cpu-percent`, I prefer to leave it as it is. If the reader wants to know more about that parameter, there are links to the kubernetes and openshift docs that he can check.

andrewdavidmackenzie:
Great! Go for it.

PR #150, opened 2020-01-17
Add option to deploy listeners using Falcon

davidor:
This PR adds the option to deploy listeners using the Falcon web server as an alternative to Puma. Falcon is a web-server built on top the async lib so we can use it when the listeners are configured to use the async Redis lib instead of the blocking one.

This is still experimental. It works, but needs more testing. Also, there's some functionality provided by Puma that's not in Falcon, you can see that in `lib/3scale/backend/server/falcon.rb`. The good news is that this allows us to configure auto-scaling. I've added a guide to configure that in Openshift (#149 ).


Gemfile.base
```diff
@@ -63,3 +63,4 @@ gem 'bugsnag', '~> 6', require: nil
 gem 'prometheus-client'
 gem 'yabeda-prometheus', '= 0.1.4'
 gem 'async-redis', '~> 0.4.1'
+gem 'falcon', '~> 0.34'
```
andrewdavidmackenzie: License of Falcon as a new dependency is OK?
Passes our license_checker?

davidor: Yes, it's MIT.

Gemfile.on_prem.lock
```diff
@@ -218,6 +255,7 @@ DEPENDENCIES
   builder (= 3.2.3)
   codeclimate-test-reporter (~> 0.6.0)
   daemons (= 1.2.4)
+  falcon (~> 0.34)
```
unleashed: This pulls in loads of new gems related to async. For a future PR: look up [`install_if`](https://bundler.io/v2.0/man/gemfile.5.html#INSTALL_IF) in Gemfile to only pull these deps conditional on some manifest/configuration/envvar to enable async or sync (could also help ensure we don't depend on async code for sync, or on puma code when using falcon).

lib/3scale/backend/server/falcon.rb
```diff
@@ -0,0 +1,48 @@
...
+    module Server
+      class Falcon
+        extend ThreeScale::Backend::Server::Utils
+
+        def self.start(global_options, options, args)
+          # Falcon does not support:
+          # - options[:daemonize]
+          # - options[:logfile]
+          # - options[:errorfile]
+          # - options[:pidfile]
```
unleashed: We can emulate all of those. A potential TODO for the future, likely in common code elsewhere.

lib/3scale/backend/server/falcon.rb
```diff
@@ -0,0 +1,48 @@
...
+          argv_add argv, true, '--count', server_model[:workers].to_s
+        end
+
+        def self.restart(global_options, options, args)
+          argv = ['falcon', 'supervisor']
+          argv_add argv, true, 'restart'
+        end
+
+        def self.stop(global_options, options, args)
+          puts 'Not implemented'
```
unleashed: Have you considered using `STDERR`? Or even `raise`? Also, there should be some means to stop the server gracefully without `R`esetting TCP connections (even if it is just sending a SIGTERM?).

Gemfile.on_prem.lock
davidor: :+1:

lib/3scale/backend/server/falcon.rb
davidor: Yes, I'll leave those as future improvements for now, because they're not needed for the kind of testing we want to do know (#149)

lib/3scale/backend/server/falcon.rb
unleashed: ping @davidor

davidor: I'll change this to raise instead :+1:

davidor: Done

PR #151, opened 2020-01-27
Add script that checks that reqs are not mixed in async mode

davidor:
This PR adds a script that verifies that requests are not mixed in async mode.
This is needed because Apisonator was not designed with Fibers in mind.


davidor:
@unleashed as you mentioned, this does not prove that there are no issues, but that's not the intention of the PR. My goal was to write some small tests that guarantee with a high-level of confidence that there are no errors.

To have more confidence in this, we can try some of the methods that you mentioned :+1:

davidor:
bors r=@unleashed

PR #154, opened 2020-01-31
stats/cleaner: handle invalid keys

davidor:
Fixes a minor bug in the `Stats::Cleaner` class. We need to handle invalid keys including corrupted ones or using a format that we no longer use.

While at it, added a couple of minor improvements.

### DIFF
--- a/lib/3scale/backend/stats/cleaner.rb
+++ b/lib/3scale/backend/stats/cleaner.rb
@@ -77,11 +77,30 @@ module ThreeScale
             logger.info("Going to delete the stats keys for these services: #{services.to_a}")
 
             unless services.empty?
+              delete_successful = true
               redis_conns.each do |redis_conn|
-                delete_keys(redis_conn, services, log_deleted_keys)
+                begin
+                  delete_keys(redis_conn, services, log_deleted_keys)
+                # If it's a connection error, mark as failed and continue
+                # cleaning other shards. If it's another kind of error, it
+                # could be a bug, so better re-raise.
+                rescue Redis::BaseConnectionError, Errno::ECONNREFUSED, Errno::EPIPE => e
+                  logger.error("Error while deleting stats of server #{redis_conn}: #{e}")
+                  delete_successful = false
+                rescue Redis::CommandError => e
+                  # Redis::CommandError from redis-rb can be raised for multiple
+                  # reasons, so we need to check the error message to distinguish
+                  # connection errors from the rest.
+                  if e.message == 'ERR Connection timed out'.freeze
+                    logger.error("Error while deleting stats of server #{redis_conn}: #{e}")
+                    delete_successful = false
+                  else
+                    raise e
+                  end
+                end
               end
 
-              remove_services_from_delete_set(services)
+              remove_services_from_delete_set(services) if delete_successful
             end
 
             logger.info("Finished deleting the stats keys for these services: #{services.to_a}")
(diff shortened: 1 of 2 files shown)


lib/3scale/backend/stats/cleaner.rb
```diff
@@ -143,15 +143,27 @@ def delete_key?(key, services_to_delete)
...
           end
 
+          # Returns nil when there's not a service encoded in the key or when
+          # the stats key has an invalid format.
           def service_from_stats_key(stats_key)
             StatsParser.parse(stats_key, nil)[:service]
+          rescue StatsParser::StatsKeyValueInvalid
+            # This could happen with legacy stats keys. For example, a long time
+            # ago some stats keys had a "city" and a "country" encoded, but
+            # always empty. That format has not been used in a long type. We'll
```
unleashed: typo `time`

davidor: :+1:

davidor:
@unleashed I added a new commit with better error handling.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -76,12 +76,31 @@ def delete!(redis_conns, log_deleted_keys: nil)
...
               redis_conns.each do |redis_conn|
-                delete_keys(redis_conn, services, log_deleted_keys)
+                begin
+                  delete_keys(redis_conn, services, log_deleted_keys)
+                # If it's a connection error, mark as failed and continue
+                # cleaning other shards. If it's another kind of error, it
+                # could be a bug, so better re-raise.
+                rescue Redis::BaseConnectionError, Errno::ECONNREFUSED, Errno::EPIPE
+                  delete_successful = false
+                  logger.error("Error while deleting stats of server: #{redis_conn}")
```
unleashed: assign the error and show it?

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -76,12 +76,31 @@ def delete!(redis_conns, log_deleted_keys: nil)
...
             unless services.empty?
               redis_conns.each do |redis_conn|
-                delete_keys(redis_conn, services, log_deleted_keys)
+                begin
+                  delete_keys(redis_conn, services, log_deleted_keys)
+                # If it's a connection error, mark as failed and continue
+                # cleaning other shards. If it's another kind of error, it
+                # could be a bug, so better re-raise.
+                rescue Redis::BaseConnectionError, Errno::ECONNREFUSED, Errno::EPIPE
+                  delete_successful = false
```
unleashed: can't replace with `find_all` + `any?`/`all?`?

davidor: Not sure what you mean.

davidor: I guess you meant something like `delete_successful = redis_conns.all? ... `

unleashed: yes

davidor: It's a bit better. Changed :+1:

lib/3scale/backend/stats/cleaner.rb
davidor: Done :+1:

lib/3scale/backend/stats/cleaner.rb
davidor: In the end, I left it as it was at the beginning. We cannot use `all?` because it stops when there's a false, and I think that `find_all` + `any?` is more complex than the original solution.

PR #156, opened 2020-02-03
Add delete stats to "backend-cron"

davidor:
What the title says.

### DIFF
--- a/openshift/backend-cron
+++ b/openshift/backend-cron
@@ -3,6 +3,11 @@
 DEFAULT_RESCHEDULE_JOBS_FREQ = 300
 DEFAULT_DELETE_STATS_FREQ = 60*60*24 # 1 day
 
+# The logger used in the delete stats code defaults to /tmp/backend_logger.log
+# When running this executable STDOUT is more convenient to check that
+# everything is working correctly.
+ENV['CONFIG_LOG_PATH'] ||= 'dev/stdout'
+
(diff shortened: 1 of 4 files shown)


openshift/backend-cron
```diff
@@ -1,7 +1,38 @@
...
+DEFAULT_RESCHEDULE_JOBS_FREQ = 300
+DEFAULT_DELETE_STATS_FREQ = 60*60*24 # 1 day
+
+# The logger used in the delete stats code defaults to /tmp/backend_logger.log
+# When running this executable STDOUT is more convenient to check that
+# everything is working correctly.
+ENV['CONFIG_LOG_PATH'] ||= 'dev/stdout'
+
+def run_rake_task_in_thread(name, freq, check_once = false)
+  Thread.new do
```
unleashed: There is nothing shared between the threads. Have you considered doing a `fork` + `exec`? It would avoid the pitfalls of GIL, although this seems like not much of an opportunity to block.

davidor: Yes, there's not much blocking. Delete stats has sleeps between `SCAN` calls and the rescheduling only runs a few redis commands every 5 min by default.

davidor:
@unleashed I force-pushed a minor change to keep the task under the `stats` namespace. I think it's more clear this way.

unleashed:
LGTM

PR #160, opened 2020-02-25
Gemfile{,.on_prem}.lock: update Nokogiri to 1.10.8

unleashed:
We are not obviously affected by the CVEs, but updating nonetheless.

### DIFF
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -17,7 +17,7 @@ end
 group :test do
   gem 'benchmark-ips', '~> 2.7.2'
   gem 'mocha',         '~> 1.3'
-  gem 'nokogiri',      '~> 1.10.4'
+  gem 'nokogiri',      '~> 1.10.8'
   gem 'pkg-config',    '~> 1.1.7'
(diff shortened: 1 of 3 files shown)


davidor:
I think we should update the `Gemfile.base` to require `~>1.10.8`.

unleashed:
> I think we should update the `Gemfile.base` to require `~>1.10.8`.

We don't need to, and it's questionable we'd even want to specify a base patchlevel version. `~> 1.10.4` will always pick the latest `1.10.Z` available, even if you deleted `Gemfile.lock`. It's a safety net useful only for internal repositories, which we don't use since we always vendor our gems straight from rubygems.org. I'd even change that and use `~> 1.10`, since the guarantees will be the same and we don't depend on anything present in 1.10.4/1.10.8 that was not there in 1.10.

davidor:
In this case, I think it would be useful to signal that we require a minimum version which we know is the first one not affected by the CVE.

unleashed:
The text above is meant to say that it is not. :)

PR #162, opened 2020-03-04
Update rake to 13.0.1

davidor:
Fixes https://github.com/advisories/GHSA-jppv-gw3r-w3q8


davidor:
I don't see why not jump directly to the latest version. The only breaking change in 13.0 is that it drops support for ruby 2.2

PR #167, opened 2020-03-10
Transactions with UNIX timestamps

unleashed:
This adds support for specifying transaction timestamps as UNIX Epoch timestamps, ie. number of seconds since Jan 1, 1970.

### DIFF
--- a/test/integration/report_test.rb
+++ b/test/integration/report_test.rb
@@ -160,6 +160,20 @@ class ReportTest < Test::Unit::TestCase
     end
   end
 
+  test 'successful report with UNIX timestamped transactions' do
+    Timecop.freeze(Time.utc(2010, 4, 23, 00, 00)) do
+      post '/transactions.xml',
+        :provider_key => @provider_key,
+        :transactions => {0 => {:app_id    => @application.id,
+                                :usage     => {'hits' => 1},
+                                :timestamp => '1271980765'}} # UNIX ts for 22/04/2010 23:59
+      Resque.run!
+
+      key = service_key(@service_id, @metric_id, :hour, '2010042223')
+      assert_equal 1, @storage.get(key).to_i
+    end
+  end
+
   test 'report uses current time if timestamp is blank' do
     Timecop.freeze(Time.utc(2010, 8, 19, 11, 24)) do
       post '/transactions.xml',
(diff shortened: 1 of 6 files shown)


test/integration/report_test.rb
```diff
@@ -591,7 +605,7 @@ def setup
     Timecop.freeze(Time.utc(2010, 5, 12, 13, 33)) do
       post '/transactions.xml',
         :provider_key => @provider_key,
-        :transactions => {0 => {:app_id => @application.id, :usage => {'hits' => 1}, :timestamp => '2012'}}
+        :transactions => {0 => {:app_id => @application.id, :usage => {'hits' => 1}, :timestamp => '2012garbage2012'}}
```
davidor: Can you adapt the comment belong accordingly?

test/integration/report_test.rb
```diff
@@ -160,6 +160,20 @@ def setup
     end
   end
 
+  test 'successful report with UNIX timestamped transactions' do
+    Timecop.freeze(Time.utc(2010, 4, 23, 00, 00)) do
+      post '/transactions.xml',
+        :provider_key => @provider_key,
+        :transactions => {0 => {:app_id    => @application.id,
+                                :usage     => {'hits' => 1},
+                                :timestamp => '1271980765'}}
```
davidor: Would be good to add a comment with the date.
`:timestamp => '1271980765'}} # 22/04/2010 @ 11:59pm`

test/integration/report_test.rb
unleashed: Mean below? :+1:

unleashed: The comment and even the test are also utterly wrong... `2012` would mean `20120101`, not `20121001`... :woman_facepalming:

unleashed:
So... `2012garbage2012` is an invalid date, but `'201210garbage201210'` is totally valid, no questions asked.

test/integration/report_test.rb
davidor: The comment looks OK. It says that because `2012` is invalid, the keys increased correspond to the current timestamp (defined by the Timecop call above) and not by the timestamp specified `2012`. Am I missing something?

davidor:
What? :dizzy_face:

test/integration/report_test.rb
unleashed: It is testing the negative case as well: that `20121001`, October 2012, monthly granularity, is not incremented. If 2012 _was_ understood, then it would have been January 2012.

davidor: 2012 is invalid. That's why Jan 2012 it's not increased.

unleashed:
> What? dizzy_face

fliptableagain

test/integration/report_test.rb
unleashed: Yes, I'm saying it's a _negative_ test case: it's testing that October 2012 is not increased. For that matter, it could have been testing that November 1990, or February 9001, wouldn't be increased, because it has nothing to do with what's being fed to the test.

unleashed: IOW, it could have used year 1234567890 and it would still be wrong, because if that year was being tested, then why choose October? If anything, specifying just a year is done with month 01.

unleashed:
So I think the only reasonable way out of this _surprising legacy from the past_, especially given this is behaviour is not documented, would be to remove this offending test, since it does not make sense any way to do the negative testing (and even then it was so broken), and accept (for now) the discovery that completely broken timestamps like the one I specified can actually mean some random thing to the system, just the same way it's always been, and then consider whether to fix what was already broken adding validation.

Ideas?

test/integration/report_test.rb
davidor: I didn't see that `20121001`. It should have been `20120101` :joy:

davidor:
@unleashed That sounds good to me :+1:

unleashed:
So in the end I added a pending test that is way more likely to bug people when running the suite to go and fix this issue one way or another, and also changed the "negative" testing to look for changes in `201201`.

davidor:
Is there something we can do about that? Have you taken a look to see if we can replace that `Date._parse`call with a similar call that does not cause other issues?

@unleashed It'd be good to open an issue so we remember to address that.

PR #168, opened 2020-03-12
Makefile: use docker by default

davidor:
The development container does not work well with Podman


unleashed:
There is a PR adding this which you'd be reverting. However, there is a known fix, which needs to be adapted to Apisonator. See https://github.com/3scale/echo-api/pull/137

davidor:
The PR in echo-api that you referenced changes many things, what's the fix that should be adapted to Apisonator? More importantly, why didn't you fix it already or at least opened an issue :trollface: 

I didn't want to revert all the changes in https://github.com/3scale/apisonator/pull/141 just wanted to change the default.

davidor:
Let's close this then and adapt the fixes you mentioned in https://github.com/3scale/echo-api/pull/137

PR #172, opened 2020-03-19
Move "huge payload" test to the "integration" test group

davidor:
The "special" category was not being executed by default in "script/test" nor "rake test". Also, it does not make much sense to separate just this test from the rest of the suite because it does not
take that long to run.


unleashed:
They weren't because they are "special", possibly because of adding a bit of time to the suite. Personally I don't think we need to separate the only test it contains from the rest unless it indeed adds a lot of time. How about moving it with the rest and either removing the special directory or adding a text file stating the purpose is to have expensive tests there?

davidor:
On my machine it runs in ~0.7s. I wouldn't separate it from the rest of the test suite because we have others that take longer (there are some that sleep for 1s).

PR #173, opened 2020-03-19
Delete unused "background reports" test

davidor:
It does nothing. Has a setup, but not tests.


davidor:
Not really. This file only had one test from the beginning and it was about end-users: https://github.com/3scale/apisonator/commit/24af813b753d7d850e32b131666d7cb7df800eca#diff-0fb07a335a966eed4d944bec40dd298b
When we removed the code to support end-users, we deleted that test and didn't realize that it was the only one in the file.

PR #178, opened 2020-03-24
Enable Prometheus metrics when using Falcon

### LINKED ISSUE #176: Add Prometheus metrics for Falcon
Puma already has Prometheus metrics (added in #174), We need to do the same for Falcon.

davidor:
Closes #176 

Same as we did for Puma in #174 but now for Falcon. This PR updates Falcon to the latest version available, because the one that we were using does not support the `--preload <path>` option that we need. Also, it seems that there's a bug in the 0.35.x versions of Falcon that makes it ignore the first SIGTERM received. I've opened an issue about that (https://github.com/socketry/falcon/issues/109) and implemented the workaround in this PR.


spec/server/listener_metrics_spec.rb
```diff
@@ -135,23 +147,33 @@ def start_puma(config_file, metrics_enabled, puma_port, metrics_port)
...
-      def stop_puma(puma_port)
-        system("bundle exec bin/3scale_backend stop -p #{puma_port}")
+      def stop_listener(port, server)
+        if server == :puma
+          system("bundle exec bin/3scale_backend stop -p #{port}")
+        else # stop not implemented in Falcon
+          # Need to send 2 SIGTERMs because a bug in falcon v0.35.x
+          # https://github.com/socketry/falcon/issues/109
+          2.times do
+            system("ps axf | grep falcon | grep -v \"grep\" | tr -s ' ' | cut -d ' ' -f 2 | xargs kill")
```
unleashed: Uuuuuugh. Note that `ps` does not come standard with some container images, in particular not with UBI 8 minimal. What you do here can be done in a single command `pidof falcon` with better guarantees (`pidof` is included in the procps package, just like `ps`).

davidor: Not sure why that provides better guarantees. `pidof falcon` does not work, we'd need `pidof ruby` which could clash with other ruby processes in the dev container.

unleashed: Fine, falcon does not set its own process title. Still that line is too broad. At least you should drop `a` (since it lists other users' processes).

unleashed: (also, perhaps you'd want `pgrep`?)

davidor: The `a` can be dropped. Regarding the other 2 I see the same problem that I mentioned above.

davidor: pgrep -f :+1:

davidor: Used `2.times { system("pkill -u #{Process.euid} -f \"ruby .*falcon\"") }` in the end.

PR #180, opened 2020-03-26
Avoid propagating unused log attrs unnecessarily

davidor:
In the past, the log field in a transaction could also include "response" and "request". Those fields are not used any more, but some callers are still sending them. We want to filter them to avoid storing them in the job queues, decoding them, etc. unnecessarily.

### DIFF
--- a/lib/3scale/backend/listener.rb
+++ b/lib/3scale/backend/listener.rb
@@ -174,6 +174,8 @@ module ThreeScale
         # not null/empty.
         params[:provider_key] = provider_key
 
+        log_without_unused_attrs(params[:log]) if params[:log]
+
         auth_status = Transactor.send method_name, provider_key, params, request: request_info
         response_auth_call(auth_status)
       rescue ThreeScale::Backend::Error => error
@@ -420,6 +422,10 @@ module ThreeScale
         transactions = params[:transactions]
         check_transactions_validity(transactions)
 
+        transactions.values.each do |tr|
+          log_without_unused_attrs(tr['log']) if tr['log']
+        end
+
         Transactor.report(provider_key, params[:service_id], transactions, response_code: 202, request: request_info)
         202
       end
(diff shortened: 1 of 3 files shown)


lib/3scale/backend/listener.rb
```diff
@@ -420,6 +422,10 @@ def do_api_method(method_name)
         transactions = params[:transactions]
         check_transactions_validity(transactions)
 
+        transactions.values.each do |tr|
```
unleashed: `tr` for transaction? I think we've been using `txn` or `tx` elsewhere, in case you'd want to check

lib/3scale/backend/listener.rb
```diff
@@ -576,6 +582,14 @@ def check_transactions_validity(transactions)
         end
       end
 
+      # In the past, the log field in a transaction could also include
+      # "response" and "request". Those fields are not used anymore, but some
+      # callers are still sending them. We want to filter them to avoid storing
+      # them in the job queues, decoding them, etc. unnecessarily.
+      def log_without_unused_attrs(log)
+        log.select { |k| k == 'code' }
```
unleashed: Using `log.select!` or `log.reject!` avoids duplicating the whole hash (note that they don't return the same as `select`/`reject`)

lib/3scale/backend/listener.rb
davidor: I don't see any `txn` or `tx`.

lib/3scale/backend/listener.rb
davidor: Yeah, but the problem with `select!` is that it returns nil when it doesn't change the hash.

unleashed: Why is that a problem in this case?

lib/3scale/backend/listener.rb
unleashed: Ok, we have been using just `t`.

lib/3scale/backend/listener.rb
davidor: Changed :+1:

PR #182, opened 2020-03-27
Fix flaky tests in prometheus metrics tests

davidor:
This PR fixes the two issues that were causing the tests failures that we have seen lately in CircleCI.

The two problems:
1) By accident, we were trying to instantiate a metrics server on each call to `process_jobs`. That's because we were calling `Worker.work` on each call to `process_jobs` and `Worker.work` starts the metrics server when configured to do so.
2) We did not wait enough after starting Puma. Sometimes we made requests before the control socket became ready.

### DIFF
--- a/spec/server/listener_metrics_spec.rb
+++ b/spec/server/listener_metrics_spec.rb
@@ -149,8 +149,14 @@ module ThreeScale
         # Send logs to /dev/null to avoid cluttering the output
         start_ok = system("#{envs_str} bundle exec bin/3scale_backend " +
                           "-s #{server} start -p #{listener_port} 2> /dev/null &")
-        raise 'Failed to start Puma' unless start_ok
-        sleep(2) # Give it some time to be ready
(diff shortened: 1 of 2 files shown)


davidor:
@unleashed I found a second issue and added a fix for it in the second commit. I ran the tests multiple times in circleCI and it seems to be stable now.

spec/server/listener_metrics_spec.rb
```diff
@@ -162,6 +168,14 @@ def stop_listener(port, server)
         sleep(2) # Give it some time to stop
       end
 
+      def wait_for_puma_control_socket
+        Timeout::timeout(5) do
+          until system("bundle exec bin/3scale_backend status")
```
unleashed: Improvement: wait until the control socket can be opened and is writable.

davidor: That's basically what the status command does. It returns an error when the control socket cannot be opened.

PR #186, opened 2020-04-06
Avoid leaking exception messages in response body

### LINKED ISSUE #153: Async: Avoid leaking exception messages in response body
When there's an exception, Falcon returns its description in the response body. Puma only does that in development mode. I opened an issue in Falcon to fix this https://github.com/socketry/falcon/issues/98

davidor:
Fixes #153

### DIFF
--- a/lib/3scale/backend/rack/internal_error_catcher.rb
+++ b/lib/3scale/backend/rack/internal_error_catcher.rb
@@ -1,0 +1,23 @@
+module ThreeScale
+  module Backend
+    module Rack
+
+      # This middleware should be the last one to run. If there's an exception,
+      # instead of propagating it to the web server, we set our own error
+      # message. The reason is that each web server handles this differently.
+      # Puma returns a generic error message, while Falcon returns the message
+      # of the exception.
+      class InternalErrorCatcher
+        def initialize(app)
+          @app = app
+        end
+
+        def call(env)
+          @app.call(env)
+        rescue
+          [500, {}, ["Internal Server Error\n".freeze]]
+        end
+      end
+    end
+  end
+end
--- a/lib/3scale/backend/rack.rb
+++ b/lib/3scale/backend/rack.rb
@@ -3,6 +3,7 @@ require '3scale/backend/logging/middleware'
 require '3scale/backend/util'
 require '3scale/backend/rack/exception_catcher'
 require '3scale/backend/rack/prometheus'
+require '3scale/backend/rack/internal_error_catcher'
 require '3scale/backend'
 
 require 'rack'
@@ -12,6 +13,8 @@ module ThreeScale
     module Rack
       def self.run(rack)
         rack.instance_eval do
+          use Rack::InternalErrorCatcher if Backend.production?
+
           Backend::Logging::External.setup_rack self
 
           if Backend.configuration.listener_prometheus_metrics.enabled


unleashed:
Rack and/or Sinatra should already be doing this. Have you tried running with `APP_ENV=production` and `RACK_ENV=production` in the environment?

davidor:
@unleashed Yes, I've tried with those.

Our code simply raises the error, puma catches it, and when configured with `RACK_ENV=production`, it replaces the exception message with a generic one to avoid leaking potentially sensitive information:
https://github.com/puma/puma/blob/81c2ccb6d4c040de03119109d8784ef28f04035e/lib/puma/server.rb#L770

What I'm proposing in this PR is to do that work in our application. The reason is that Falcon, unlike Puma, does not replace the error message with a generic one.

unleashed:
> What I'm proposing in this PR is to do that work in our application. The reason is that Falcon, unlike Puma, does not replace the error message with a generic one.

I know what you are trying to do. What I am saying is that Sinatra and Rack should be able to handle this for us, but we might have the wrong configuration (and handling). I think the code we have for dealing with exceptions can be added to a Sinatra/Rack error handling mechanism rather than in custom middlewares. For example in:
https://github.com/3scale/apisonator/blob/v2.99.0/lib/3scale/backend/listener.rb#L9

...we could do this:
```ruby
configure :production do
  disable :raise_errors
end
```
...and figure out what is breaking and fix it. We could add this to a new issue to tackle later if you want. See :raise_errors in http://sinatrarb.com/configuration.html.

As for this workaround, why can't this be added just for Falcon?

davidor:
Ah yes, you're right. We could simply delete the line that you linked and add `disable :raise_errors` in the `configure :production` block.

davidor:
We can't just `disable :raise_errors` with our codebase. Currently, we rely on raising those errors, so the ExceptionCatcher middleware can catch the `Backend::Error` ones (`ApplicationKeyInvalid`, `UserKeyInvalid`, etc.) and set the appropriate status code and message.

After that step, the logging middleware writes the error and later, the external logging middleware sends it to bugsnag/airbrake.

With that structure, I think that the easiest solution is the one I proposed. Add another middleware at the end of the chain that checks if something raised, and if so, sets a generic response message instead of bubbling the exception up to Puma/Falcon and relying on them to set the error.

unleashed:
That's why I said to open a new issue and analyze how to best use the facilities instead of adding more middlewares. Btw, Falcon should have such a provision, like Puma and Thin have, so maybe open an issue there as well?

Wouldn't the easiest thing be to replace https://github.com/3scale/apisonator/blob/v2.99.0/lib/3scale/backend/logging/middleware.rb#L35 with:
```ruby
rescue Exception => e
  raise e if !env.production?
  [500, "internal server error"]
end
```

Alternatively, when setting up each server, we could provide server specific middlewares. In any case, up to you now.

lib/3scale/backend/rack/internal_error_catcher.rb
```diff
@@ -0,0 +1,23 @@
...
+      # Puma returns a generic error message, while Falcon returns the message
+      # of the exception.
+      class InternalErrorCatcher
+        def initialize(app)
+          @app = app
+        end
+
+        def call(env)
+          @app.call(env)
+          rescue
```
unleashed, line 17: wrong indentation

davidor:
@unleashed I opened a PR a while ago in Falcon, but it has not been merged.

I don't like the idea of dealing with this in the logging middleware as you propose. First, it should be in the external middleware logging rather than in the logging middleware. Also, I don't think it should be the responsibility of either of those 2 to change the response body. I'd rather make this explicit even if it means adding an extra trivial middleware. I think that ideally this should be in the exception catcher one, but that requires more work and potentially other problems. I don't think reorganizing all of that is worth it at this point just to add such a small feature.

unleashed:
@davidor you made the point of the middleware being the easiest, not what is or not something to like - that is just doing an ugly hack, just like the way we are handling these now, which is why I proposed to open an issue to investigate options. :)

lib/3scale/backend/rack/internal_error_catcher.rb
davidor: Fixed :+1:

PR #187, opened 2020-04-06
Delete listeners limitation from async doc

davidor:
The listeners can now be configured to use the redis async lib.


davidor:
Two runs failed. We'll need to investigate why those tests are flaky.
I realized that it normally fails when running Ruby 2.6. I wonder if that's mere coincidence.

PR #188, opened 2020-04-06
Fix flaky Prometheus metrics test

davidor:
This PR adds a workaround for what seems to be a bug in Falcon.
I've run this multiple times to make sure it fixes the test flakiness: https://app.circleci.com/pipelines/github/3scale/apisonator?branch=flaky-metrics-test

### DIFF
--- a/spec/server/listener_metrics_spec.rb
+++ b/spec/server/listener_metrics_spec.rb
@@ -162,10 +162,18 @@ module ThreeScale
       def stop_listener(port, server)
         if server == :puma
           system("bundle exec bin/3scale_backend stop -p #{port}")
+          sleep(2) # Give it some time to stop
         else # stop not implemented in Falcon
           system("pkill -u #{Process.euid} -f \"ruby .*falcon\"")
+          sleep(2) # Give it some time to stop
+
+          # TODO: investigate why occasionally Falcon does not kill its children
+          # processes ("Falcon Server").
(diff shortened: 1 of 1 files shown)


spec/server/listener_metrics_spec.rb
```diff
@@ -166,6 +166,10 @@ def stop_listener(port, server)
           system("pkill -u #{Process.euid} -f \"ruby .*falcon\"")
         end
         sleep(2) # Give it some time to stop
+
+        # TODO: investigate why occasionally Falcon does not kill its children
+        # processes ("Falcon Server").
+        system("pkill -u #{Process.euid} -f \"Falcon Server\"")
```
unleashed, line 172: Does "Falcon Server" also match the parent process? Also, this should be moved to the `else` branch above, and now that I see this, perhaps best to extract the whole `else` to a method that does the following:

1. Signal via SIGTERM the parent process
2. Sleep
3. Signal via SIGTERM any remaining processes.
4. __IFF__ the step above was successful (`system` will return true) then there were remaining processes, so:
  4.1. Sleep a couple seconds
  4.2. Kill the processes via `pkill --signal SIGKILL ...`

unleashed: Note: pkill will error out if it can't find such processes, and in that case step 4 should be skipped.

davidor: OK. That's more robust :+1: 
Changed.

PR #189, opened 2020-04-06
Check SystemCallError in listener metrics spec

davidor:
Same reason as in https://github.com/3scale/apisonator/pull/129/


davidor:
@gsaslis This fixes one of the issues you're seeing in your pipeline.
It's the same issue we had some months ago in #129

Issue #18, opened 2018-04-16
Total job process time logged is not correct


davidor:
For each processed job, the workers log some information like the process time, the total time (process time + time in the queue), etc.

The second is not correct. The reason is that the enqueue time is stored in Redis by a listener when it enqueues a job:

https://github.com/3scale/apisonator/blob/18edf5c00bb727f4d09ab9a71768556249d08953/lib/3scale/backend/transactor.rb#L204

Whereas the total time is calculated by the worker when it finished processing a job:

https://github.com/3scale/apisonator/blob/18edf5c00bb727f4d09ab9a71768556249d08953/lib/3scale/backend/background_job.rb#L43

That's the reason why sometimes we get inconsistent numbers in the logs. For example, sometimes we see that the process time is greater than the process time plus the time in the queue. In many deployments, those two timestamps always come from different machines (listeners vs background workers) so comparing them is not reliable.

PR #191, opened 2020-04-14
Fix concurrency issue in BackgroundJob

davidor:
This PR replaces the usage of a class variable in `BackgroundJob`.

That class variable causes problems when using the async mode because fibers can modify it while others still need to read it. Luckily, with the current code, the only attribute affected is "enqueue_time", which is not accurate anyway (see https://github.com/3scale/apisonator/issues/18).

Even without taking the async mode into consideration using a class var is counter-intuitive. The arguments of a particular job (service ID, app ID, etc.) should belong to an instance, they shouldn't be stored in a class var.

### DIFF
--- a/lib/3scale/backend/background_job.rb
+++ b/lib/3scale/backend/background_job.rb
@@ -9,11 +9,10 @@ module ThreeScale
 
       class << self
         def perform(*args)
-          @args = args
-          perform_wrapper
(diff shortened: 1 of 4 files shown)


davidor:
Yes, I checked. I didn't see any other usages of `@args`.

PR #205, opened 2020-04-22
Use 3scale's redis-rb fork

davidor:
The fork is based on the 4.1.3 redis-rb version and includes a couple of fixes: https://github.com/3scale/redis-rb/pull/1 and https://github.com/3scale/redis-rb/pull/2


Gemfile.base
```diff
@@ -63,3 +62,8 @@ gem 'bugsnag', '~> 6', require: nil
 gem 'yabeda-prometheus', '~> 0.5.0'
 gem 'async-redis', '~> 0.4.1'
 gem 'falcon', '~> 0.35'
+
+# Use a patched redis-rb that fixes an issue when trying to connect with
+# sentinels and avoids retrying calls when there's a timeout to prevent
+# duplicated commands. It's based on version 4.1.3.
+gem 'redis', git: 'https://github.com/3scale/redis-rb', ref: '35301b3d952975300a4cb30d408ae3b5969d0440'
```
unleashed, line 69: I prefer this to refer to a branch, then locking the revision in the lockfile.

davidor: I just followed what we're doing for Puma, but I can change this, I see that in others we use branches.

unleashed: Use the example of resque - the puma one is wrong, it has a branch that we should use.

davidor: Changed :+1:

PR #208, opened 2020-04-24
Adjust prometheus histogram buckets for the worker

davidor:
The current ones are not very informative, because the majority of run times fall in the first or second bucket.

### DIFF
--- a/lib/3scale/backend/worker_metrics.rb
+++ b/lib/3scale/backend/worker_metrics.rb
@@ -11,7 +11,8 @@ Yabeda.configure do
       comment "How long jobs take to run"
       unit :seconds
       tags %i[type]
-      buckets [0.005, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]
+      # Most requests will be under 100ms, so use a higher granularity from there
+      buckets [0.01, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.09, 0.1, 0.25, 0.5, 0.75, 1]
     end
   end
 end


lib/3scale/backend/worker_metrics.rb
```diff
@@ -11,7 +11,8 @@
       comment "How long jobs take to run"
       unit :seconds
       tags %i[type]
-      buckets [0.005, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]
+      # Most requests will be under 100ms, so use a higher granularity from there
+      buckets [0.01, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.09, 0.1, 0.25, 0.5, 0.75, 1]
```
unleashed, line 15: I'd keep the 0.005 for the super nice requests. And then I'd smooth the big step in intervals jumping from 0.01 to 0.150 then 0.250 in two consecutive steps. Wdyt about `0.1, 0.15, 0.2, 0.3, 0.5, 0.75, 1, (...rest?)`?

davidor: Commented one second late :sweat_smile: 
@slopezz and me are trying this now. We'll refine this values as needed when we get more experience from our tests.

slopezz: Right now on staging with a not real environment, we are havinf super low times (even lower than 5ms), so maybe 5ms could be de lowest one

You can see current values on  a test we did some hours ago: https://grafana-route-application-monitoring.apps.stg-saas-ocp4-3.stg-saas.3sca.net/d/0aHddS3Wz/3scale-backend?orgId=1&from=1587721676575&to=1587730412335

PR #214, opened 2020-05-12
Make max number of Redis connections configurable

davidor:
Only applies to the async mode.


davidor:
I ran several tests on my development machine and the throughput and latencies stop improving at 10 connections more or less. As always, the optimal value depends on many factors, that's why it's configurable.

PR #217, opened 2020-05-27
Update activesupport to  5.2.4.3

davidor:
This version fixes https://github.com/advisories/GHSA-2p68-f74v-9wc6
apisonator is not affected by the CVE.


davidor:
I ran `bundle update activesupport`. All the gems updated are dependencies of `activesupport`. The only one that's also a dependency of other gems is concurrent-ruby, but it's fine, it's not a major bump. The tests pass and everything looks good.

PR #218, opened 2020-05-27
Delete unused Status#value_for_application_usage_limit

davidor:
This method is never called and it does the same as `value_for_usage_limit` defined just above.


davidor:
Right :D

PR #224, opened 2020-06-04
Fix bug when loading the usage limits and no_body=true

davidor:
This PR fixes a bug introduced in #221 

The problem is that I assumed that in `Application.load_usage_limits_affected_by` all the metrics passed were already validated, but that was a mistake because the listener does not validate the metrics until it runs the `Limits` validator, which happens later in the process.

More specificaly, when a metric does not exist, `Metric.load_id` returns nil and then, the sort here crashes:
https://github.com/3scale/apisonator/blob/e965a2228c331c21b159925d16888a5f473ef26c/lib/3scale/backend/application.rb#L241

We didn't caught this because we didn't have tests that used both `no_body=true`, and reported more than 1 metric and at least an invalid one. I have included tests for that in this PR.

In the future, maybe we should make the metrics validation more explicit as a separate test in the transactor rather than have it hidden in the `Limits` validator.

### DIFF
--- a/lib/3scale/backend/application.rb
+++ b/lib/3scale/backend/application.rb
@@ -230,11 +230,12 @@ module ThreeScale
       # Loads the usage limits affected by the metrics received, that is, the
       # limits that are defined for those metrics plus all their ancestors in
       # the metrics hierarchy.
+      # Raises MetricInvalid when a metric does not exist.
       def load_usage_limits_affected_by(metric_names)
         metric_ids = metric_names.flat_map do |name|
           [name] + Metric.ascendants(service_id, name)
         end.uniq.map do |name|
-          Metric.load_id(service_id, name)
+          Metric.load_id(service_id, name) || raise(MetricInvalid.new(name))
         end
 
         # IDs are sorted to be able to use the memoizer
--- a/test/integration/authorize/basic_test.rb
+++ b/test/integration/authorize/basic_test.rb
@@ -779,4 +779,17 @@ class AuthorizeBasicTest < Test::Unit::TestCase
 
     assert_error_resp_with_exc(ThreeScale::Backend::MetricInvalid.new('non_existing'))
   end
+
+  test 'returns error when the usage includes a metric that does not exist and no_body=true' do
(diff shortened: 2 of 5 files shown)


lib/3scale/backend/application.rb
```diff
@@ -230,11 +230,12 @@ def load_all_usage_limits
       # Loads the usage limits affected by the metrics received, that is, the
       # limits that are defined for those metrics plus all their ancestors in
       # the metrics hierarchy.
+      # Raises MetricInvalid when a metric does not exist.
       def load_usage_limits_affected_by(metric_names)
         metric_ids = metric_names.flat_map do |name|
           [name] + Metric.ascendants(service_id, name)
         end.uniq.map do |name|
-          Metric.load_id(service_id, name)
+          Metric.load_id(service_id, name) || raise(MetricInvalid.new(name))
```
unleashed, line 238: I think this is problematic - any time the above fails it will assume the issue is an invalid metric. Is the method returning falsey always going to be the case of invalid metrics? Are we using this method in any other context where embedding the raised exception in `Metric.load_id` for falsey values would not work?

davidor: `Metric.load_id` can return:
1) An ID when the metric exists
2) Nil when the metric does no exist.
3) A Redis Exception.
So yes, we can assume that if it's nil, it's because the metric does no exist.

unleashed: Sure - the last question is still unanswered. Does it make sense to raise within `Metric.load_id`?

davidor: I remember that we had this discussion in the past and the answer was no, but now there's only one caller apart from this one and it also raises: https://github.com/3scale/apisonator/blob/e965a2228c331c21b159925d16888a5f473ef26c/lib/3scale/backend/metric/collection.rb#L78

So it looks like it should work, but it needs some testing to be sure. Let's address that in a separate PR.

unleashed: Sounds good! :+1:

PR #226, opened 2020-06-18
Remove native oauth feature

davidor:
This is a feature that has been deprecated for a long time.
@andreumasferrer gave us the :+1: to delete this.


davidor:
I don't see how those methods can be reused.

`require_params!` can't be reused because it raises a generic error whereas the rest of methods in the `Listener` class raise specific errors such as `ServiceIdMissing`.
`ensure_authenticated!` is not needed in the rest of the methods of the class.

PR #230, opened 2020-07-13
Fix Prometheus counter of 5xx errors

davidor:
Ref: https://issues.redhat.com/browse/THREESCALE-5622

### DIFF
--- a/lib/3scale/backend/rack/prometheus.rb
+++ b/lib/3scale/backend/rack/prometheus.rb
@@ -8,7 +8,15 @@ module ThreeScale
 
         def call(env)
           began_at = Time.now.getutc
-          status, header, body = @app.call(env)
+
+          begin
+            status, header, body = @app.call(env)
+          rescue Exception => e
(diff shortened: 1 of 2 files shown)


lib/3scale/backend/rack/prometheus.rb
```diff
@@ -8,7 +8,15 @@ def initialize(app)
...
           began_at = Time.now.getutc
-          status, header, body = @app.call(env)
+
+          begin
+            status, header, body = @app.call(env)
+          rescue StandardError => e
+            ListenerMetrics.report_resp_code(env['REQUEST_PATH'], 500)
+            ListenerMetrics.report_response_time(env['REQUEST_PATH'], Time.now - began_at)
+            raise e
+          end
```
unleashed, line 18: This won't catch `Exception.new "abc"`, which will end in a 500 status code.

davidor: Yeah, at first I thought that increasing the counter in situations like a signal or a syntax error is not going to be that important, but let's add it for completeness.
Changed :+1:

PR #235, opened 2020-09-07
Handle corner case with "TransactionTimestampNotWithinRange" in notify jobs

davidor:
This handles an error that's very unlikely to happen. I think that today it's the first time I've seen this.

We were not handling `TransactionTimestampNotWithinRange` in notify jobs. At first glance, this doesn't look necessary because the timestamps in those jobs come from the listener processes and not the user. However, there are a few corner cases that make this necessary. Please check the explanation I added [here](https://github.com/3scale/apisonator/compare/notify-job-handle-timestamp-not-in-range?expand=1#diff-a1238ecbec4e11a641deb8a67f9f2a74R27).

We need to handle `TransactionTimestampNotWithinRange`, otherwise, notify jobs are re-enqueued even though they can't be processed correctly.

### DIFF
--- a/lib/3scale/backend/transactor/notify_job.rb
+++ b/lib/3scale/backend/transactor/notify_job.rb
@@ -14,12 +14,30 @@ module ThreeScale
             if application_id && Application.exists?(master_service_id, application_id)
               master_metrics = Metric.load_all(master_service_id)
 
-              ProcessJob.perform([{
-                service_id: master_service_id,
-                application_id: application_id,
-                timestamp: timestamp,
-                usage: master_metrics.process_usage(usage)
-              }])
+              begin
+                ProcessJob.perform([{
+                  service_id: master_service_id,
+                  application_id: application_id,
+                  timestamp: timestamp,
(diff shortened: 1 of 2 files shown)


lib/3scale/backend/transactor/notify_job.rb
```diff
@@ -16,12 +16,30 @@ def perform_logged(provider_key, usage, timestamp, _enqueue_time)
...
+                  timestamp: timestamp,
+                  usage: master_metrics.process_usage(usage)
+                }])
+              rescue TransactionTimestampNotWithinRange => e
+                # This is very unlikely to happen. The timestamps in a notify
+                # job are not set by users, they are set by the listeners. If
+                # this error happens it might mean that:
+                # a) The worker started processing this job way after the
+                # listener produced it. This can happen for example if we make
+                # some requests to a listener but no workers. The listeners will
```
unleashed, line 32:
```suggestion
                # some requests to a listener with no workers. The listeners will
```

unleashed (approved):
Check typo and :+1:.

PR #236, opened 2020-09-14
Add Prometheus metrics for the internal API

davidor:
This PR adds prometheus metrics for the internal API. The metrics exposed are the request type (services, applications, metrics, etc.) and the response times. It's similar to what we did for the auth and report endpoints in #174

### DIFF
--- a/lib/3scale/backend/listener_metrics.rb
+++ b/lib/3scale/backend/listener_metrics.rb
@@ -4,14 +4,34 @@ require 'rack'
 module ThreeScale
   module Backend
     class ListenerMetrics
-      REQUEST_TYPES = {
+      AUTH_AND_REPORT_REQUEST_TYPES = {
         '/transactions/authorize.xml' => 'authorize',
         '/transactions/oauth_authorize.xml' => 'authorize_oauth',
         '/transactions/authrep.xml' => 'authrep',
         '/transactions/oauth_authrep.xml' => 'authrep_oauth',
         '/transactions.xml' => 'report'
       }
-      private_constant :REQUEST_TYPES
+      private_constant :AUTH_AND_REPORT_REQUEST_TYPES
+
+      INTERNAL_API_PATHS = [
+        [/\/services\/.*\/alert_limits/, 'alerts'],
+        [/\/services\/.*\/applications\/.*\/keys/, 'application_keys'],
+        [/\/services\/.*\/applications\/.*\/referrer_filters/, 'application_referrer_filters'],
+        [/\/services\/.*\/applications/, 'applications'],
+        [/\/services\/.*\/errors/, 'errors'],
+        [/\/events/, 'events'],
+        [/\/services\/.*\/metrics/, 'metrics'],
+        [/\/service_tokens/, 'service_tokens'],
+        [/\/services/, 'services'],
+        [/\/services\/.*\/stats/, 'stats'],
+        [/\/services\/.*\/plans\/.*\/usagelimits/, 'usage_limits'],
+        [/\/services\/.*\/applications\/.*\/utilization/, 'utilization'],
+      ].freeze
+      private_constant :INTERNAL_API_PATHS
+
+      # Most requests will be under 100ms, so use a higher granularity from there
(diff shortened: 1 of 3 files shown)


lib/3scale/backend/listener_metrics.rb
```diff
@@ -93,6 +132,24 @@ def code_group(resp_code)
             'unknown'.freeze
           end
         end
+
+        def req_type(path)
+          AUTH_AND_REPORT_REQUEST_TYPES[path] || internal_api_req_type(path)
+        end
+
+        def internal_api_req_type(path)
+          (_regex, type) = INTERNAL_API_PATHS.find { |(regex, _)| regex.match path }
```
unleashed, line 141: This is very unfortunate. With this we need to match against regexes a path which Sinatra already matched for us, so double work, but this time much more expensive than matchings strings.

We could use a [Custom Route Matcher](http://sinatrarb.com/intro.html) although that seems a bit too much for this case (and it's not like it's greatly documented), or in a more practical way we could use `request.env['prometheus.iapi.req_type'] = 'alerts'` in the relevant Internal API endpoints?

lib/3scale/backend/listener_metrics.rb
```diff
@@ -4,14 +4,34 @@
...
+        [/\/services\/.*\/applications\/.*\/utilization/, 'utilization'],
+      ].freeze
+      private_constant :INTERNAL_API_PATHS
+
+      # Most requests will be under 100ms, so use a higher granularity from there
+      TIME_BUCKETS = [0.01, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.09, 0.1, 0.25, 0.5, 0.75, 1]
+      private_constant :TIME_BUCKETS
 
       class << self
         ERRORS_4XX_TO_TRACK = Set[403, 404, 409].freeze
```
unleashed, line 37: FYI, I know this is not part of this PR (it's from #174 and I overlooked this), but this is a case for an ordered (by frequency) `Array` rather than a `Set` - rationale is it performs better for sets in which the 1st and 2nd elements are queried with high frequency compared to the rest - I just unearthed [this gist](https://gist.github.com/unleashed/764aa1af85b122bc9646) and enabled `perf` and disabled `accumulated` for numbers.

Maybe a case for a follow-up PR.

davidor: That was an interesting investigation :smile_cat: 

We can take a look in a different PR. Honestly, I don't think it's too important and I'm not sure we can apply your results to this specific case without doing some measurements first. First of all, I'm not sure about the distribution of hits in the set. I think it depends a lot on the use case. Also, this is the kind of thing that can change between Ruby versions.

lib/3scale/backend/listener_metrics.rb
davidor: Your solution would be more efficient, that's true. However, it implies adding a similar line to the one you pasted in all the files that define the endpoints. Also, this is part of an optional feature, and it only affects the internal API, not the main one. For those reasons, I'd rather merge it as it is, and only change this if it proves to be a problem when we run our benchmarks.

lib/3scale/backend/listener_metrics.rb
unleashed: > [...] apply your results to this specific case without doing some measurements first.

That's easy to check with the gist and the frequencies.

> Also, this is the kind of thing that can change between Ruby versions.

It hasn't changed from 2.1 up to 2.7, and it fundamentally should not change because `Array` and `Set` should have basic guarantees about their structure. You can read the [details](https://github.com/3scale/backend/pull/359#issuecomment-86990515) [private link] for some fun.

unleashed: Let's keep this for another PR if you are willing to take it :)

lib/3scale/backend/listener_metrics.rb
unleashed: :+1:

unleashed (approved):

PR #237, opened 2020-09-16
NotifyJob: handle case when metrics do not exist in Porta

davidor:
This PR handles an error case in notify jobs. Those jobs report some metrics to the master account. The problem is that Apisonator can't guarantee that those metrics have been created, it relies on Porta to do that, and we've found that in some environments those metrics were not being created.

### DIFF
--- a/lib/3scale/backend/configuration.rb
+++ b/lib/3scale/backend/configuration.rb
@@ -22,6 +22,9 @@ module ThreeScale
       end
     end
 
+    MASTER_SERVICE_ID_DEFAULT = 1
+    private_constant :MASTER_SERVICE_ID_DEFAULT
+
     NOTIFICATION_BATCH_DEFAULT = 10000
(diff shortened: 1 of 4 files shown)


lib/3scale/backend/configuration.rb
```diff
@@ -121,8 +121,8 @@ def parse_int(value, default)
         config.master.metrics.transactions_authorize = CONFIG_MASTER_METRICS_TRANSACTIONS_AUTHORIZE_DEFAULT
       end
 
-      if config.master_service_id.nil? || config.master_service_id.to_s.empty?
-        raise InvalidMasterServiceId
```
unleashed, line 125: doesn't this removal render this error completely unused?

davidor: Good catch. Fixed :+1:

unleashed (approved):

PR #238, opened 2020-09-17
Run notify job only when the master service ID is explicitly defined

davidor:
This is related to the previous PR: #237 

We assumed that we could set a default for the master service ID, but after talking with the Porta team, we realized that's not a safe assumption.

This PR changes the configuration setup so it no longer sets a default for the service ID, and also runs notify jobs only when that ID has been set.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -158,14 +158,15 @@ send events to Porta.
 ### CONFIG_MASTER_SERVICE_ID
 
 - The service ID of the master account in Porta.
-- Optional. Defaults to 1.
-- Applies to: worker.
+- Optional. If not provided, Apisonator does not report metrics to the master
+account of Porta.
+- Applies to: listener, worker.
 - Format: string.
 
 ### CONFIG_MASTER_METRICS_TRANSACTIONS
 
-- Name of the metric configured in the master account of Porta to
-track the number of report calls.
+- Name of the metric configured in the master account of Porta to track the
+number of report calls. Applies only when `CONFIG_MASTER_SERVICE_ID` is set.
 - Optional. Defaults to "transactions".
 - Applies to: listener.
 - Format: string.
@@ -173,7 +174,7 @@ track the number of report calls.
 ### CONFIG_MASTER_METRICS_TRANSACTIONS_AUTHORIZE
 
 - Name of the metric configured in the master account of Porta to track the
-number of authorize calls.
+number of authorize calls. Applies only when `CONFIG_MASTER_SERVICE_ID` is set.
 - Optional. Defaults to "transactions/authorize".
 - Applies to: listener.
 - Format: string.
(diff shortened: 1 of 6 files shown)


docs/configuration.md
```diff
@@ -158,22 +158,23 @@ send events to Porta.
 ### CONFIG_MASTER_SERVICE_ID
 
 - The service ID of the master account in Porta.
-- Optional. Defaults to 1.
-- Applies to: worker.
+- Optional. If not provided, Apisonator does not report metrics to the master
+account of Porta.
```
unleashed, line 162: I think we might want to emit a warning about this in the logs.

lib/3scale/backend/transactor/notify_batcher.rb
```diff
@@ -30,9 +30,13 @@ def key_for_notifications_batch
...
         def notify(provider_key, usage)
-          # batch several notifications together so that we can process just one
+          # We need the master service ID to report its metrics. If it's not
+          # set, we don't need to notify anything.
+          # Batch several notifications together so that we can process just one
           # job for a group of them.
-          notify_batch(provider_key, usage)
+          unless configuration.master_service_id.to_s.empty?
+            notify_batch(provider_key, usage)
+          end
```
unleashed, line 39: could have been a nice oneliner :)

unleashed (commented):
What if the worker picks up a notification job and it does not have a master id configured? Will it raise?

lib/3scale/backend/transactor/notify_batcher.rb
davidor: :tophat:

docs/configuration.md
davidor: I'll answer the 2 questions here.

If a worker picks up a job and it does not have a master service ID configured, it logs an error and notifies it with the external error manager configured (bugsnag, etc.). That was included in the previous PR linked above. That case shouldn't be common though. If the listener has a master service ID configured and the worker does not, it means that something is not configured correctly.

About emitting a warning when the master service ID, I don't think it's a good idea. The reason is that there are many customers with this configuration already. If we change this they'll suddenly start seeing some warnings about a feature they didn't even know existed. I think we should emit a warning only if product decides that this feature should be enabled by default in following versions.

unleashed: I think I did not convey the context correctly, so here I meant to raise a warning in the worker if a job is found with no master id configured, not in the listener, since that would depend on whether this becomes mandatory or not.

> it means that something is not configured correctly

For example, upgrading from a previous release where the listener's previous version enqueued the job and the workers new version picks it up without a configured master svc id. We could have this to signal a problem with a fully configured and upgraded release, although there is this window of time in which in-flight jobs could give this warning (I don't think investing in solving that is worth it).

> If a worker picks up a job and it does not have a master service ID configured, it logs an error and notifies it with the external error manager configured (bugsnag, etc.).

So that's what I meant, thanks.

unleashed (approved):
bors r+

davidor:
bors retry

PR #243, opened 2020-11-18
Add endpoint in the internal API to get the provider key from a (token, service_id) pair

davidor:
Requests need to include a provider key or a service token. If they include a service token, we can't map a request to a provider. The endpoint introduced in this PR allows us to do so.


unleashed (approved):
bors r+ - @davidor let's open an issue in pisoni for implementing access to this just in case.

PR #245, opened 2021-01-18
Use case insensitive strings when parsing booleans in config

davidor:
It seems reasonable to expect that something like "True" should work the same as "true".

For example, the kubernetes libs used by our saas-operator sets booleans to "True" and "False".

### DIFF
--- a/openshift/3scale_backend.conf
+++ b/openshift/3scale_backend.conf
@@ -5,6 +5,11 @@ def parse_int_env(env_name)
   !env_var.nil? && !env_var.empty? ? env_var.to_i : nil
 end
 
+def parse_boolean_env(env_name)
+  value = ENV[env_name]
+  !value.nil? && value.downcase == 'true'
+end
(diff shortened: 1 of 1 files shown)


openshift/3scale_backend.conf
```diff
@@ -5,6 +5,11 @@ def parse_int_env(env_name)
   !env_var.nil? && !env_var.empty? ? env_var.to_i : nil
 end
 
+def parse_boolean_env(env_name)
+  value = ENV[env_name]
+  !value.nil? && value.downcase == 'true'
```
unleashed, line 10: Why not also check for the string `"1"`?
```suggestion
  value == '1'.freeze || value&.downcase == 'true'.freeze
```

davidor: OK. Committed your suggestion :+1:

unleashed (approved):
bors r+

PR #247, opened 2021-01-22
Stop reporting usages of 0

davidor:
This PR makes sure that no usages of 0 are reported in the authrep and the report endpoints.

The problem was that they were reported, and report jobs ended up creating stats keys with a value of 0. A stats key set to 0 is equivalent to a non-existing one when rate-limiting, so the feature was working correctly but wasting space in the DB.

### DIFF
--- a/lib/3scale/backend/transactor.rb
+++ b/lib/3scale/backend/transactor.rb
@@ -20,8 +20,14 @@ module ThreeScale
       def report(provider_key, service_id, transactions, context_info = {})
         service = Service.load_with_provider_key!(service_id, provider_key)
 
-        report_enqueue(service.id, transactions, context_info)
-        notify_report(provider_key, transactions.size)
+        # A usage of 0 does not affect rate-limits or stats, so we do not need
+        # to report it.
+        filtered_transactions = filter_usages_with_0(transactions.clone)
+
+        return if filtered_transactions.empty?
+
+        report_enqueue(service.id, filtered_transactions, context_info)
+        notify_report(provider_key, filtered_transactions.size)
       end
 
       def authorize(provider_key, params, context_info = {})
@@ -180,6 +186,24 @@ module ThreeScale
         else
           application.load_all_usage_limits
         end
+      end
+
+      def filter_usages_with_0(transactions)
+        res = transactions
+
+        # There are plenty of existing tests using both a string and a symbol
+        # when accessing the usage.
+        usage_key = ['usage'.freeze, :usage]
+
+        res.each do |_idx, tx|
+          usage_key.each { |k| tx[k]&.reject! { |_metric, delta| delta.to_s == '0'.freeze } }
+        end
+
+        res.reject! do |_idx, tx|
+          usage_key.all? { |k| tx[k].nil? || tx[k].empty? }
+        end
+
+        res
       end
 
       def storage
--- a/lib/3scale/backend/transactor.rb
+++ b/lib/3scale/backend/transactor.rb
@@ -143,7 +143,7 @@ module ThreeScale
 
         usage = params[:usage]
 
-        filtered_usage = usage&.reject { |_metric, delta| delta.to_s == '0' }
+        filtered_usage = filter_metrics_without_inc(usage.clone) if usage
 
         if ((filtered_usage && !filtered_usage.empty?) || params[:log]) && status.authorized?
           application_id = status.application.id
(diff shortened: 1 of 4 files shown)


lib/3scale/backend/transactor.rb
```diff
@@ -182,6 +188,24 @@ def preload_usage_limits(application, no_body_enabled, usage_in_params)
         end
       end
 
+      def filter_usages_with_0(transactions)
+        res = transactions.clone
```
unleashed, line 200: Why this call to `clone`?

davidor: To avoid mutating the original `params[:usage]`, I think that would make the code and the tests harder to follow.

lib/3scale/backend/transactor.rb
```diff
@@ -182,6 +188,24 @@ def preload_usage_limits(application, no_body_enabled, usage_in_params)
...
+
+        # There are plenty of existing tests using both a string and a symbol
+        # when accessing the usage.
+        usage_key = ['usage'.freeze, :usage]
```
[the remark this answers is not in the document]
davidor: `params` is an `IndifferentHash` when set up by Sinatra, but as the comment says, we have _plenty_ of tests calling this method directly and using both options with standard ruby Hashes.

lib/3scale/backend/transactor.rb
unleashed: But you're already mutating the original contents regardless of this - the clone is not a deep clone. The only reason I can think of to want to keep the original `params[:usage]` is for logging/monitoring, and at that point you can still let the callers decide to pass a clone or not. I don't think we are testing anywhere that the usage in params remains untouched anyway.

lib/3scale/backend/transactor.rb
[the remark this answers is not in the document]
davidor: Because the tests don't send an IndifferentHash and changing all of them looks like a lot of work that maybe we should do but it's not related to this PR.

lib/3scale/backend/transactor.rb
davidor: Yes, I was thinking about aspects like logging. Would you be OK with this if I just change the call to `filter_usages_with_0(transactions.clone)` instead of cloning inside the function?

unleashed: Yes. But are we doing any such "aspects"?

lib/3scale/backend/transactor.rb
```diff
@@ -182,6 +196,24 @@ def preload_usage_limits(application, no_body_enabled, usage_in_params)
...
+
+        res.each do |_idx, tx|
+          usage_key.each { |k| tx[k]&.reject! { |_metric, delta| delta.to_s == '0'.freeze } }
+        end
+
+        res.reject! do |_idx, tx|
+          usage_key.all? { |k| tx[k].nil? || tx[k].empty? }
+        end
+
+        res
```
unleashed, line 214: This looks like it works a bit too much - how about this other version walking the txns and doing lookups just once?
```suggestion
        res.delete_if do |_idx, tx|
          usage = tx['usage'.freeze] || tx[:usage] or next
          usage.delete_if { |_metric, delta| delta.to_s == '0'.freeze }.empty?
        end
```

lib/3scale/backend/transactor.rb
davidor: To answer that question you need to follow the code to be sure, and think about it every time you make a change in the callers chain. Right now that includes the few lines of code after the call to `filter_usages_with_0`, the caller method in `Listener` and all the rack middlewares that might run after it. I think it's easier to avoid mutating it.

lib/3scale/backend/transactor.rb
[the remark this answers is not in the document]
unleashed: Ok.

lib/3scale/backend/transactor.rb
```diff
@@ -137,9 +143,17 @@ def do_authrep(method, provider_key, params, context_info)
 
         usage = params[:usage]
 
-        if (usage || params[:log]) && status.authorized?
+        filtered_usage = usage&.reject { |_metric, delta| delta.to_s == '0' }
```
unleashed, line 146: I think we can split the helper method to filter 0's so that you can use it for both report and authrep?

lib/3scale/backend/transactor.rb
unleashed: Ok. :+1:

lib/3scale/backend/transactor.rb
davidor: Done. Check new commit.

lib/3scale/backend/transactor.rb
davidor: Done. Check new commit.

lib/3scale/backend/transactor.rb
davidor: Changed this as explained above.

lib/3scale/backend/transactor.rb
```diff
@@ -182,6 +196,19 @@ def preload_usage_limits(application, no_body_enabled, usage_in_params)
         end
       end
 
+      def filter_usages_with_0(transactions)
+        # There are plenty of existing tests using both a string and a symbol
+        # when accessing the usage.
+        transactions.delete_if do |_idx, tx|
+          (usage = tx['usage'.freeze] || tx[:usage]) or next
```
unleashed, line 203:
```ruby
puts 'The parens are not needed' or
  puts 'because the natural language operators have the lowest precendence' or
  puts 'even lower than assignment' and
  nil.call.causes_no_method_error and
  no_worries.it_is_never_executed_anyway and
  (ツ) or
  puts "ok, well, whatever!"
```

unleashed (approved):
bors r+

unleashed:
bors retry

PR #249, opened 2021-01-25
Provide helper for "first traffic" set in Stats::Keys

davidor:
Minor refactor. Adds a helper in the `Stats::Keys` module to get the "first traffic" set for a given service ID.


unleashed (approved):
bors r+

PR #250, opened 2021-01-26
Add rake task to delete stats keys set to 0

davidor:
This PR adds a rake task to delete the stats keys set to 0. Those keys are useless and occupy Redis memory unnecessarily. They were generated by the bug that was fixed in #247

### DIFF
--- a/lib/3scale/backend/stats/cleaner.rb
+++ b/lib/3scale/backend/stats/cleaner.rb
@@ -45,6 +45,9 @@ module ThreeScale
         STATS_KEY_PREFIX = 'stats/'.freeze
         private_constant :STATS_KEY_PREFIX
 
+        REDIS_CONN_ERRORS = [Redis::BaseConnectionError, Errno::ECONNREFUSED, Errno::EPIPE].freeze
+        private_constant :REDIS_CONN_ERRORS
+
         class << self
           include Logging
           def mark_service_to_be_deleted(service_id)
@@ -81,22 +84,9 @@ module ThreeScale
               redis_conns.each do |redis_conn|
                 begin
                   delete_keys(redis_conn, services, log_deleted_keys)
-                # If it's a connection error, mark as failed and continue
-                # cleaning other shards. If it's another kind of error, it
-                # could be a bug, so better re-raise.
-                rescue Redis::BaseConnectionError, Errno::ECONNREFUSED, Errno::EPIPE => e
-                  logger.error("Error while deleting stats of server #{redis_conn}: #{e}")
+                rescue => e
+                  handle_redis_exception(e, redis_conn)
                   delete_successful = false
-                rescue Redis::CommandError => e
-                  # Redis::CommandError from redis-rb can be raised for multiple
-                  # reasons, so we need to check the error message to distinguish
-                  # connection errors from the rest.
-                  if e.message == 'ERR Connection timed out'.freeze
-                    logger.error("Error while deleting stats of server #{redis_conn}: #{e}")
-                    delete_successful = false
-                  else
-                    raise e
-                  end
                 end
               end
 
@@ -107,6 +97,25 @@ module ThreeScale
           end
 
           private
+
+          def handle_redis_exception(exception, redis_conn)
+            # If it's a connection error, simply log so we can continue with
+            # other shards. If it's another kind of error, it could be caused by
+            # a bug, so better re-raise.
+
+            case exception
+            when *REDIS_CONN_ERRORS
+              logger.error("Error while deleting stats of server #{redis_conn}: #{exception}")
+            when Redis::CommandError
+              if exception.message == 'ERR Connection timed out'.freeze
+                logger.error("Error while deleting stats of server #{redis_conn}: #{exception}")
+              else
+                raise exception
+              end
+            else
+              raise exception
+            end
+          end
 
           # Returns a set with the services included in the
           # SET_WITH_SERVICES_MARKED_FOR_DELETION Redis set.
--- a/lib/3scale/backend/stats/cleaner.rb
+++ b/lib/3scale/backend/stats/cleaner.rb
@@ -96,6 +96,30 @@ module ThreeScale
             logger.info("Finished deleting the stats keys for these services: #{services.to_a}")
           end
 
+          # Deletes all the stats keys set to 0.
+          #
+          # Stats keys set to 0 are useless and occupy Redis memory
+          # unnecessarily. They were generated due to a bug in previous versions
+          # of Apisonator.
+          # Ref: https://github.com/3scale/apisonator/pull/247
+          #
+          # As the .delete function, this one also receives a collection of
+          # instantiated Redis clients and those need to connect to Redis
+          # servers directly.
+          #
+          # @param [Array] redis_conns Instantiated Redis clients.
+          # @param [IO] log_deleted_keys IO where to write the logs. Defaults to
+          #             nil (logs nothing).
+          def delete_stats_keys_set_to_0(redis_conns, log_deleted_keys: nil)
+            redis_conns.each do |redis_conn|
+              begin
+                delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
+              rescue => e
+                handle_redis_exception(e, redis_conn)
+              end
+            end
+          end
+
           private
 
           def handle_redis_exception(exception, redis_conn)
@@ -196,6 +220,30 @@ module ThreeScale
             # always empty. That format has not been used in a long time. We'll
             # simply ignore those keys.
             nil
+          end
+
+          def delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
+            cursor = 0
+
+            loop do
+              cursor, keys = redis_conn.scan(cursor, count: SCAN_SLICE)
+
+              stats_keys = keys.select { |k| is_stats_key?(k) }
+
+              unless stats_keys.empty?
+                values = redis_conn.mget(*stats_keys)
+                to_delete = stats_keys.zip(values).select { |_, v| v == '0'.freeze }.map(&:first)
+
+                unless to_delete.empty?
+                  redis_conn.del(to_delete)
+                  to_delete.each { |k| log_deleted_keys.puts k } if log_deleted_keys
+                end
+              end
+
+              break if cursor.to_i == 0
+
+              sleep(SLEEP_BETWEEN_SCANS)
+            end
           end
         end
       end
--- a/lib/3scale/backend/stats/cleaner.rb
+++ b/lib/3scale/backend/stats/cleaner.rb
@@ -83,17 +83,21 @@ module ThreeScale
             logger.info("Going to delete the stats keys for these services: #{services.to_a}")
 
             unless services.empty?
-              delete_successful = true
-              redis_conns.each do |redis_conn|
+              _ok, failed = redis_conns.partition do |redis_conn|
                 begin
                   delete_keys(redis_conn, services, log_deleted_keys)
+                  true
                 rescue => e
                   handle_redis_exception(e, redis_conn)
-                  delete_successful = false
+                  false
                 end
               end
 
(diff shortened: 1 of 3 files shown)


lib/3scale/backend/stats/cleaner.rb
```diff
@@ -108,6 +98,25 @@ def delete!(redis_conns, log_deleted_keys: nil)
 
           private
 
+          def handle_redis_exception(exception, redis_conn)
+            # If it's a connection error, simply log so we can continue with
+            # other shards. If it's another kind of error, it could be caused by
+            # a bug, so better re-raise.
+
+            is_connection_err = REDIS_CONN_ERRORS.any? do |conn_error|
+              exception.is_a?(conn_error)
```
unleashed, line 131: `is_a?` is not what `rescue` uses to discriminate exceptions. Instead it uses `===` (_threequals_).

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -108,6 +98,25 @@ def delete!(redis_conns, log_deleted_keys: nil)
...
+            end
+
+            is_timeout = exception.is_a?(Redis::CommandError) &&
+              exception.message == 'ERR Connection timed out'.freeze
```
unleashed, line 135: Can an exception be both a connection error and a timeout?

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -108,6 +98,25 @@ def delete!(redis_conns, log_deleted_keys: nil)
...
+
+            if is_connection_err || is_timeout
+              logger.error("Error while deleting stats of server #{redis_conn}: #{exception}")
+            else
+              raise exception
+            end
```
unleashed, line 141: What you want for this whole method is a `case` statement replicating the original `rescue` (it will also automatically apply `===` just like rescue).

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -197,6 +221,30 @@ def service_from_stats_key(stats_key)
...
+          def delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
+            cursor = 0
+
+            loop do
+              cursor, keys = redis_conn.scan(cursor, count: SCAN_SLICE)
+
+              stats_keys = keys.select { |k| is_stats_key?(k) }
+
+              unless stats_keys.empty?
+                values = redis_conn.mget(*(stats_keys.to_a))
```
unleashed, line 234: Isn't `.to_a` superfluous because `stats_keys` is just a simple array with the keys? I think this would be `redis_conn.mget(*stats_keys)`.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -197,6 +221,30 @@ def service_from_stats_key(stats_key)
...
+                to_delete = Hash[stats_keys.zip(values)].select { |_k, v| v == '0'.freeze }
```
unleashed, line 235: You don't really need to create a `Hash` (or 2) here. How about `stats_keys.zip(values).select { |_, v| v == '0'.freeze }.map(&:first)`? You could even do these in place in the zipped array.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -197,6 +221,30 @@ def service_from_stats_key(stats_key)
...
+
+                unless to_delete.empty?
+                  redis_conn.del(to_delete.keys)
```
unleashed, line 238: What if the `scan`, the `mget` or the `del` fails? You would break and not recover the original scan data from the point it broke. Given how much time these can take, would you want to try to recover a couple of times before giving up?

unleashed (changes requested):
Check out the comments for some minor changes needed - also s/func/method/g in commit messages :)

lib/3scale/backend/stats/cleaner.rb
davidor: No: https://github.com/redis/redis-rb/blob/v4.1.3/lib/redis/errors.rb

lib/3scale/backend/stats/cleaner.rb
davidor: :man_facepalming:

lib/3scale/backend/stats/cleaner.rb
davidor: You're right, I forgot about this. The values are needed in the other method because it prints keys and values, but there's no need to print the values in this one because they're always 0.

lib/3scale/backend/stats/cleaner.rb
davidor: Fixed.

lib/3scale/backend/stats/cleaner.rb
davidor: Fixed :+1:

lib/3scale/backend/stats/cleaner.rb
davidor: Changed. Although there's a little bit more duplication with the `case`, I think it's better because it avoids calling `===` directly.

lib/3scale/backend/stats/cleaner.rb
davidor: No need to fix because https://github.com/3scale/apisonator/pull/250#discussion_r564722826

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -106,8 +96,51 @@ def delete!(redis_conns, log_deleted_keys: nil)
...
+          #
+          # As the .delete function, this one also receives a collection of
+          # instantiated Redis clients and those need to connect to Redis
+          # servers directly.
+          #
+          # @param [Array] redis_conns Instantiated Redis clients.
+          # @param [IO] log_deleted_keys IO where to write the logs. Defaults to
+          #             nil (logs nothing).
+          def delete_stats_keys_set_to_0(redis_conns, log_deleted_keys: nil)
+            redis_conns.each do |redis_conn|
```
unleashed, line 114: We could use something like `partition` to collect successul and unsuccessful connections and output the details (including URIs) of each category. Wdyt?

davidor: `logger.error` already tells you the details of the connection that failed

unleashed: Good - what about reporting the URIs that failed after the rake task completes?

davidor: Why wait until the end?

unleashed: To make it easy for the user to know which ones should be retried - in particular if there is a screenful of errors.

davidor: How can there be a screenful of errors? How many shards do you think users have for this? :scream:

unleashed: I'm not thinking about shards. I'm thinking about backtraces when any one of those `raise exception` lines is executed, say, after yet any other shard failed. Or having a shard fail while logging the deleted keys to stdout, and then logging even more deleted keys, losing the message in the middle of a million screens and having to go hunt for it manually if your terminal still has it in its buffer.

davidor: If the code gets to the `raise exception` line it doesn't continue with other shards. It stops the execution. It's done that way because if it gets to that line, something unexpected (possible bug?) happened. It's not for things that we might expect like connection errors, timeouts, etc.

There can be at most 1 error message per shard. It could get mixed with the keys as you said, but keep in mind that the keys are written to STDOUT and the errors to STDERR.

unleashed: I'm aware, I am thinking that the usual case of people launching this task will be with both stderr and stdout going to the same terminal output. This is a suggestion that I think will improve the usability and be appreciated by the guy launching the task at a low cost in code changes, but if you think it's not worth it then let's resolve this thread.

davidor: Wait, this made me realize that what we are doing in `handle_redis_exception` is not correct. When there's an error, we should rescue, retry, log the error if all the attempts failed, and continue with the next shard. Right now the code continues with the same one.

lib/3scale/backend/stats/cleaner.rb
davidor: The existing `delete!` method didn't have retries. I added the for both: https://github.com/3scale/apisonator/pull/250/commits/5d8a9e1a0d7ded5cf9c1d2bc028cf75b7251aa75

lib/3scale/backend/stats/cleaner.rb
davidor: Added a commit implementing your suggestion about logging only at the end: https://github.com/3scale/apisonator/pull/250/commits/c836a3dfe5b66d0310bc998c6842b9c040f716f1

davidor:
@unleashed please check the 3 new commits.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -245,6 +252,16 @@ def delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
               sleep(SLEEP_BETWEEN_SCANS)
             end
           end
+
+          def with_retries
+            begin
+              retries ||= 0
+              return yield
+            rescue Exception => e
+              retry if (retries += 1) < MAX_RETRIES_REDIS_ERRORS
```
unleashed, line 267: :astonished: This line is a bit surprising. How is this following your usual readability policy? Why don't you just extract the increment to a statement before the `retry` check?

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -245,6 +252,16 @@ def delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
               sleep(SLEEP_BETWEEN_SCANS)
             end
           end
+
+          def with_retries
+            begin
+              retries ||= 0
+              return yield
```
unleashed, line 265: No need to add `return` here.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -245,6 +252,16 @@ def delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
               sleep(SLEEP_BETWEEN_SCANS)
             end
           end
+
+          def with_retries
+            begin
+              retries ||= 0
```
unleashed, line 264: `retries = 0` could be used before `begin` and you wouldn't need to check again nor initialize with `||=`.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -245,6 +252,16 @@ def delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
               sleep(SLEEP_BETWEEN_SCANS)
             end
           end
+
+          def with_retries
```
unleashed, line 262: Not important, but would be nice for this or other more important places: `def with_retries(max = MAX_RETRIES_REDIS_ERRORS)`

unleashed (approved):
Ok, but check the few comments I made. I think there is some more room for improvement but the affected code does not merit the extra effort.

lib/3scale/backend/stats/cleaner.rb
davidor: Fair enough. Changed :+1:

lib/3scale/backend/stats/cleaner.rb
davidor: Fixed.

lib/3scale/backend/stats/cleaner.rb
davidor: Fixed.

lib/3scale/backend/stats/cleaner.rb
davidor: I don't have a strong opinion about this. You would have complained either way :trollface:

davidor:
bors r=@unleashed

davidor:
bors retry

PR #251, opened 2021-02-02
Update async-redis to 0.5.1

davidor:
Raises the minimum version to 0.5.1. It's not the latest release, but it's the latest that still supports ruby 2.4.
That version includes the PR that I contributed to add support for Redis Sentinels.


unleashed (approved):
bors r+

PR #253, opened 2021-02-03
Async worker: stop sharing redis client instance with the job fetcher thread

davidor:
This PR fixes a bug that triggers sometimes in the worker with async enabled and when a job fails. The root cause is that the main thread and the one that fetches job from the queue were sharing the same Redis instance to access the queue DB. Reusing the same instance raises errors because a fiber is shared between different threads.

This PR also includes some related fixes in the test suite.


unleashed (approved):
LGTM.

bors r+

PR #254, opened 2021-02-04
Async worker: exit cleanly when fetcher fails

davidor:
When the job fetcher thread fails, it does not make sense to keep the worker running, because it's not going to do anything useful.
This PR notifies when the job fetcher fails and exits cleanly ensuring that the shared job queue is always closed.

### DIFF
--- a/lib/3scale/backend/worker_async.rb
+++ b/lib/3scale/backend/worker_async.rb
@@ -68,6 +68,10 @@ module ThreeScale
           # unblocks when there are new jobs or when .close() is called
           job = @jobs.pop
 
+          # If job is nil, it means that the queue is closed. No more jobs are
+          # going to be pushed, so shutdown.
+          shutdown unless job
+
           break if @shutdown
 
           @reactor.async { perform(job) }
--- a/spec/unit/job_fetcher_spec.rb
+++ b/spec/unit/job_fetcher_spec.rb
@@ -245,6 +245,29 @@ module ThreeScale
                 .at_least(1).times
           end
         end
+
+        context 'when there is an error not related with Redis connectivity' do
+          let(:error) { Exception.new('Some error') }
+          let(:queue) { Queue.new }
+          let(:job_fetcher) { JobFetcher.new(redis_client: test_redis) }
+
+          before do
+            allow(test_redis).to receive(:blpop).and_raise error
+            allow(Worker.logger).to receive(:notify)
+          end
+
+          it 'closes the queue' do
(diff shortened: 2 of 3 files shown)


spec/unit/job_fetcher_spec.rb
```diff
@@ -245,6 +245,29 @@ module Backend
                 .at_least(1).times
           end
         end
+
+        context 'when there is an error not related with Redis connectivity' do
+          let(:error) { RuntimeError.new('Some error') }
```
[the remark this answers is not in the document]
davidor: Fair enough. Changed :+1:

lib/3scale/backend/worker_async.rb
```diff
@@ -68,6 +68,10 @@ def schedule_jobs
           # unblocks when there are new jobs or when .close() is called
           job = @jobs.pop
 
+          # The queue is closed, no more jobs are going to be pushed, so
+          # shutdown.
+          shutdown if @jobs.closed?
```
unleashed, line 73: Can `job` be a valid job and in this point have `@jobs.closed?` be truthy? My suspicion is that this case is not possible, but I wanted to check with you.

unleashed:
r=me after https://github.com/3scale/apisonator/pull/254#discussion_r570437330 is confirmed ok.

lib/3scale/backend/worker_async.rb
davidor: Hmm actually I think it's possible. I think `job` could be a valid one, and the other thread could call `@jobs.close` before this one calls `@jobs.closed?`

davidor: I think we need this instead:
```

unleashed: Then the alternative question is: can `job` be `nil` and `@jobs.closed` be falsey? :D

davidor: `Queue#pop` blocks until there's a new element in the queue or the queue is closed. So, if `job` is nil, the queue is closed.

unleashed: Ok, push the fix then and r=me.

unleashed (approved):
r+ waiting for https://github.com/3scale/apisonator/pull/254#discussion_r570448012

lib/3scale/backend/worker_async.rb
davidor: Fixed :rocket:

davidor:
bors r=unleashed

PR #255, opened 2021-02-05
Async worker: reconnect correctly after redis connection errors

davidor:
This PR fixes a bug: the async worker was not reconnecting correctly after redis connection errors.

The reason is that it was instantiating a client without the "resque" namespace.

### DIFF
--- a/lib/3scale/backend/job_fetcher.rb
+++ b/lib/3scale/backend/job_fetcher.rb
@@ -80,10 +80,11 @@ module ThreeScale
 
               # Re-instantiate Redis instance. This is needed to recover from
               # Errno::EPIPE, not sure if there are others.
-              @redis = ThreeScale::Backend::QueueStorage.connection(
-                  ThreeScale::Backend.environment,
-                  ThreeScale::Backend.configuration
+              @redis = Redis::Namespace.new(
+                WorkerAsync.const_get(:RESQUE_REDIS_NAMESPACE),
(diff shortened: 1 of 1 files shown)


spec/integration/worker_async_spec.rb
```diff
@@ -28,51 +36,73 @@ module Backend
...
         end
 
-        worker.shutdown
-
-        worker_thread.join
-
-        expect(Storage.instance.get(stats_key).to_i).to eq n_reports
+        ensure
+          worker.shutdown
+          worker_thread.join
```
unleashed, line 105: hm, I think you could just force MRI to kill the thread here if it did not finish in time to make sure we won't wait too much for whatever reason (ie. actually having some sort of connectivity problem, such as having the server respond very slowly), because as it stands we'll raise just fine, but we'll have to wait.

davidor: I tried it without applying the fix and it worked fine.

unleashed (approved):

davidor:
bors r=@unleashed

spec/integration/worker_async_spec.rb
davidor: Writing a correct integration test for this is complicated because of some limitations of the testing suite that require some work to be fixed.
I'll merge this now because we need the fix to unblock other work, I'll try to come back to this next week.

davidor:
bors retry

PR #257, opened 2021-02-09
Delete unused stats code

davidor:
This PR removes some code related with deleting stats that we no longer use.
We don't need to keep it, we can always revert this PR if it's needed again.


unleashed (approved):
bors r+

PR #259, opened 2021-02-10
Gemfiles: pin rack-test to 0.8.2

davidor:
When updating rack-test some tests fail because rack-test 0.8.3 includes a breaking change that doesn't work well with `rspec-apic-document`. See the github issues linked in the Gemfile.


unleashed (approved):
`¯\_(ツ)_/¯`

bors r+

PR #261, opened 2021-02-11
Update our puma fork to 4.3.7

davidor:
This PR updates our puma fork to 4.3.4. We already tried this version some time ago and know that it works.
It'd be nice to update directly to 5.x but it's not a high priority now. 4.3.4 unblocks other work, like being able to switch to UBI8.

### DIFF
--- a/Gemfile.base
+++ b/Gemfile.base
@@ -46,7 +46,7 @@ group :development, :test do
 end
 
 # Default server by platform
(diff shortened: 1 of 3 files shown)


unleashed (commented):
As discussed in the chat, let's wait to update our fork to at least 4.3.5 and use that unless it breaks us. See https://github.com/puma/puma/commit/f3b409c565d67557c04ad37c10a42dd8cad0b655#commitcomment-39478516.

davidor:
Updated to v4.3.7 which is the latest release of the 4.x series.

unleashed (approved):
:+1:

bors r+

PR #262, opened 2021-02-12
Replace deprecated `URI.encode` calls

davidor:
Ruby 2.7 throws a warning when calling `URI.encode`:
"warning: URI.escape is obsolete" (encode is an alias of escape)


unleashed (approved):
LGTM

bors r+

PR #265, opened 2021-02-15
Install ruby 2.7 in CI image

davidor:
This PR install ruby 2.7 in the CI image. It also uninstalls 2.6 because we no longer need to ensure that apisonator works with that version.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -97,14 +97,14 @@ RUN git clone --recurse-submodules --depth 1 --no-checkout --progress \
 RUN git clone --recurse-submodules --depth 1 --no-checkout --progress \
     https://github.com/rbenv/ruby-build.git "${RBENV_RUBYBUILD_ROOT}"
 
-ARG RBENV_TAG=v1.1.1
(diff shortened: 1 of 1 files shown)


Dockerfile.ci
```diff
@@ -248,6 +248,7 @@ RUN mkdir -p ~/.local/bin \
        "  for ruby_version in \$(rbenv whence ruby); do\n" \
        "    echo \"Switching to \${ruby_version}\"\n" \
        "    rbenv shell \"\${ruby_version}\"\n" \
+       "    gem update --system\n" \
```
unleashed, line 251: Have you needed this line for it to work? I remember having had issues with some broken gem updates in the past, so I went with the gem version shipping which each ruby version.

unleashed (approved):
LGTM, but check the comment before merging.

Dockerfile.ci
davidor: Yes, this is needed for ruby 2.7. With the version installed by default, it crashes.

davidor:
bors r=@unleashed

PR #267, opened 2021-02-16
Update pry to 0.14.0 and pry-doc to 1.1.0

davidor:
The pry-doc version that was being used gives a warning on Ruby 2.7. I've updated both pry and pry-doc to the latest version.

```bash
[ruby@apisonator-dev apisonator]$ bundle exec pry
ruby/2.7.2 isn't supported by this pry-doc version
```


Gemfile.base
```diff
@@ -36,8 +36,8 @@ end
 group :development do
   gem 'sshkit'
   gem 'source2swagger', git: 'https://github.com/3scale/source2swagger', branch: 'backend'
-  gem 'pry',      '~> 0.11.3'
-  gem 'pry-doc',  '~> 0.11.1'
+  gem 'pry',      '~> 0.14'
+  gem 'pry-doc',  '~> 1.1'
```
unleashed, line 40: `~> 1` should be enough here if pry-doc follows semver (that is not the case with `0.x` versions).

unleashed (approved):

Gemfile.base
davidor: 1.0.0 doesn't support 2.7, so the minimum is 1.1.

davidor:
bors r=@unleashed

PR #268, opened 2021-02-16
Update docker image to use UBI8 with ruby 2.7

davidor:
I've marked this PR as "draft" because we'll also need to add the 2 Dockerfiles used for the release. Those 2 are based on the Dockerfile included in this PR, so I prefer to get this one reviewed first.

Regarding the dockerfile, it's heavily based on the existing one plus the one used in https://github.com/3scale/echo-api.


unleashed (approved):

unleashed (approved):
Looks good - I think the only step left is moving the dev environment.

davidor:
Yes, but no need to do that now.

davidor:
bors r=@unleashed

PR #270, opened 2021-02-23
Remove redundant prometheus config param

davidor:
This PR removes a redundant config param used to enable Prometheus metrics in the listener.

We already have an ENV to do that and we need it because the metric server is started in Puma/Falcon "before_fork" and the configuration is not loaded at that point.

### DIFF
--- a/lib/3scale/backend/rack.rb
+++ b/lib/3scale/backend/rack.rb
@@ -17,7 +17,10 @@ module ThreeScale
 
           Backend::Logging::External.setup_rack self
 
-          if Backend.configuration.listener_prometheus_metrics.enabled
+          # Notice that this cannot be specified via config, it needs to be an
+          # ENV because the metric server is started in Puma/Falcon
+          # "before_fork" and the configuration is not loaded at that point.
+          if ENV['CONFIG_LISTENER_PROMETHEUS_METRICS_ENABLED'].to_s.downcase.freeze == 'true'.freeze
             use Rack::Prometheus
           end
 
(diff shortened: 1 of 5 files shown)


lib/3scale/backend/rack.rb
```diff
@@ -17,7 +17,10 @@ def self.run(rack)
 
           Backend::Logging::External.setup_rack self
 
-          if Backend.configuration.listener_prometheus_metrics.enabled
+          # Notice that this cannot be specified via config, it needs to be an
+          # ENV because the metric server is started in Puma/Falcon
+          # "before_fork" and the configuration is not loaded at that point.
+          if ENV['CONFIG_LISTENER_PROMETHEUS_METRICS_ENABLED'].to_s.downcase == 'true'
```
unleashed, line 23: `.freeze`?

lib/3scale/prometheus_server.rb
```diff
@@ -4,7 +4,7 @@
 require_relative '../3scale/backend/listener_metrics'
 
 # Config is not loaded at this point, so read ENV instead.
-if ENV['CONFIG_LISTENER_PROMETHEUS_METRICS_ENABLED'].to_s == 'true'
+if ENV['CONFIG_LISTENER_PROMETHEUS_METRICS_ENABLED'].to_s.downcase == 'true'
```
unleashed, line 7: `.freeze`?

unleashed (approved):
`freeze` and merge?

davidor:
bors r=@unleashed

PR #272, opened 2021-03-05
Stop maintaining unused "current_max" key in Alerts

davidor:
We are maintaining a "current_max" key that's not used anywhere. I guess it was used at some point and we forgot to delete it when we removed the functionality that checked that key.


unleashed (approved):
bors r+

PR #275, opened 2021-03-08
Perf optimization: Check if alerts can be raised before calculating utilization

davidor:
The `update_alerts` call in `Aggregator` can be costly. It loads the usages limits and the current usages to check whether an alert should be raised. This PR introduces a perf optimization. It adds a check to see if there can still be alerts raised for an app (there can only be one every 24h per alert level) before calculating all the usages.

### DIFF
--- a/test/unit/alerts_test.rb
+++ b/test/unit/alerts_test.rb
@@ -2,10 +2,12 @@ require File.expand_path(File.dirname(__FILE__) + '/../test_helper')
 require '3scale/backend/alert_limit'
 
 class AlertsTest < Test::Unit::TestCase
-  def setup
-    @service_id = 10
+  include TestHelpers::Sequences
 
-    Alerts::ALERT_BINS.each { |val| AlertLimit.save(@service_id, val) }
+  def setup
+    @storage = Storage.instance(true)
+    @service_id = next_id
+    @application_id = next_id
   end
 
   test 'check proper use of bins' do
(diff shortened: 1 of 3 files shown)


test/unit/alerts_test.rb
```diff
@@ -15,4 +17,33 @@ def setup
...
+  test 'can_raise_more_alerts? returns false when there are no allowed bins' do
+    assert_false(Alerts.can_raise_more_alerts?(@service_id, @application_id))
+  end
+
+  test 'can_raise_more_alerts? returns false when there is already an alert for the highest bin' do
+    allowed_bins = [50, 100]
+    allowed_bins.each { |bin| AlertLimit.save(@service_id, bin) }
+
+    key_highest_already_notified = Alerts.send(
+      :key_already_notified, @service_id, @application_id, 100
```
unleashed, line 30: s/100/allowed_bins.sort.last/

test/unit/alerts_test.rb
```diff
@@ -15,4 +17,33 @@ def setup
...
+    assert_false(Alerts.can_raise_more_alerts?(@service_id, @application_id))
+  end
+
+  test 'can_raise_more_alerts? returns true when there are no alerts for the highest bin' do
+    allowed_bins = [50, 100, 200]
+    allowed_bins.each { |bin| AlertLimit.save(@service_id, bin) }
+
+    # Set notified for a level that's not the highest
+    key_already_notified = Alerts.send(
+      :key_already_notified, @service_id, @application_id, 100
```
unleashed, line 43: s/100/allowed_bins.sort[-2]/

unleashed (approved):
Looks good, but fix comments.

test/unit/alerts_test.rb
davidor: Changed :+1:

test/unit/alerts_test.rb
davidor: Changed :+1:

davidor:
bors r=@unleashed

PR #277, opened 2021-04-08
Delete unused Service attrs related with end-users functionality

davidor:
This PR deletes 3 unused attributes of `Service`. Those attributes are related with the end-users functionality that was removed some time ago, so they're no longer needed.


unleashed (approved):
bors r+

PR #279, opened 2021-05-14
ignore .env and tools files

akostadinov:


.gitignore
```diff
@@ -8,6 +8,10 @@ coverage
 
 .*.swp
 *~
+.env
+.ruby-gemset
```
davidor, line 12: `.ruby-gemset` and `.ruby-version` are committed in the repo. They are very useful when using tools like rbenv or rvm to choose the appropriate ruby version and set of gems.

akostadinov: @davidor , in such case I added `.tools-versions` in #305, would you take a look at it?

PR #280, opened 2021-06-02
Use OAuth service rules when using OAuth services in authorize and authrep endpoints

unleashed:
We recently received tickets from users that were having issues getting OIDC to work with the 3scale Istio Adapter.

The reason is that Porta stores `client_secret`s as `app_key`s of applications. This is not needed (and arguably should not be done at all), but the effect is that any service configured as "oauth" calling the `authrep.xml` or `authorize.xml` endpoints must provide the client secret as the app_key parameter.

This does not make sense at all, since the premise is that the caller provides a signed OIDC token (ie. via the Authorization header with a bearer token authz type) and this is validated by the proxy / SM API client _before_ calling into Apisonator. There's no gain in providing yet another secret to Apisonator itself.

This change considers whether a service is configured to use OIDC when receiving auth*.xml calls (while leaving the oauth_auth*.xml calls alone), and changes the requirement that a valid `app_key` must be provided so that if none such key is provided the call will succeed, but if it is provided then success will depend on providing a correct one.

Additionally the pre-existing behaviour of being able to use `user_key`s with these kinds of services is kept, even if it does not make sense, to avoid any potential breakage (this is why only oauth_auth*.xml endpoints will require the `app_id` parameter).

This PR also paves the way to deprecate the oauth_auth*.xml endpoints, as after this lands there will be nothing in oauth_auth*.xml that cannot be done already via authorize and authrep.

### DIFF
--- a/lib/3scale/backend/transactor.rb
+++ b/lib/3scale/backend/transactor.rb
@@ -61,6 +61,8 @@ module ThreeScale
 
       def validate(oauth, provider_key, report_usage, params, request_info)
         service = Service.load_with_provider_key!(params[:service_id], provider_key)
+        oidc_service = !oauth && service.backend_version == 'oauth'.freeze
+
         # service_id cannot be taken from params since it might be missing there
         service_id = service.id
 
@@ -70,12 +72,18 @@ module ThreeScale
         # significant.
         params[:app_id] = nil if app_id && app_id.empty?
 
-        if oauth
-          raise ApplicationNotFound.new nil if app_id.nil?
-          validators = Validators::OAUTH_VALIDATORS
-        else
-          validators = Validators::VALIDATORS
-        end
+        # While OIDC without an app_id makes little sense, we would break existing
+        # behaviour when calling non oauth_auth*.xml endpoints if we returned an
+        # error here, so only do this for oauth_auth*.xml endpoints.
+        raise ApplicationNotFound.new nil if oauth && app_id.nil?
+
+        validators = if oidc_service
+                       Validators::OIDC_VALIDATORS
+                     elsif oauth
+                       Validators::OAUTH_VALIDATORS
+                     else
+                       Validators::VALIDATORS
+                     end
 
         params[:user_key] = nil if params[:user_key] && params[:user_key].empty?
         application = Application.load_by_id_or_user_key!(service_id,
--- a/lib/3scale/backend/validators.rb
+++ b/lib/3scale/backend/validators.rb
@@ -21,6 +21,13 @@ module ThreeScale
       OAUTH_VALIDATORS = ([Validators::OauthSetting,
                            Validators::OauthKey,
                            Validators::RedirectURI] + COMMON_VALIDATORS).freeze
+
+      # OIDC specific validators will only check app keys when app_key is given.
+      #
+      # No need to add OauthSetting, since we need to check that to tell
+      # OIDC apart from the rest when calling authrep.xml (note lack of
+      # the oauth_ prefix).
+      OIDC_VALIDATORS = ([Validators::OauthKey] + COMMON_VALIDATORS).freeze
     end
   end
 end
--- a/test/integration/authrep/application_keys_test.rb
+++ b/test/integration/authrep/application_keys_test.rb
@@ -12,80 +12,115 @@ class AuthrepApplicationKeysTest < Test::Unit::TestCase
 
     Memoizer.reset!
 
-    setup_oauth_provider_fixtures
+    setup_provider_fixtures
+    setup_oauth_provider_fixtures_noclobber
 
     @application = Application.save(:service_id => @service.id,
                                     :id         => next_id,
                                     :state      => :active,
                                     :plan_id    => @plan_id,
                                     :plan_name  => @plan_name)
+    @application_oauth = Application.save(:service_id => @service_oauth.id,
+                                          :id         => next_id,
+                                          :state      => :active,
+                                          :plan_id    => @plan_id,
+                                          :plan_name  => @plan_name)
   end
 
   test_authrep 'succeeds if no application key is defined nor passed' do |e|
     get e, :provider_key => @provider_key,
-           :app_id       => @application.id
+           :service_id   => @service_oauth.id,
+           :app_id       => @application_oauth.id
 
     assert_authorized
   end
 
   test_authrep 'succeeds if one application key is defined and the same one is passed' do |e|
-    application_key = @application.create_key
+    application_key = @application_oauth.create_key
 
     get e, :provider_key => @provider_key,
-           :app_id       => @application.id,
+           :service_id   => @service_oauth.id,
+           :app_id       => @application_oauth.id,
            :app_key      => application_key
 
     assert_authorized
   end
 
   test_authrep 'succeeds if multiple application keys are defined and one of them is passed' do |e|
-    application_key_one = @application.create_key
-    _application_key_two = @application.create_key
+    application_key_one = @application_oauth.create_key
+    _application_key_two = @application_oauth.create_key
 
     get e, :provider_key => @provider_key,
-           :app_id       => @application.id,
+           :service_id   => @service_oauth.id,
+           :app_id       => @application_oauth.id,
            :app_key      => application_key_one
 
     assert_authorized
   end
 
-  test_authrep 'does not authorize if application key is defined but not passed',
+  test_authrep 'does not authorize if application key is defined but not passed for non oauth services',
                except: :oauth_authrep do |e|
     @application.create_key
 
     get e, :provider_key => @provider_key,
+           :service_id   => @service.id,
            :app_id       => @application.id
 
     assert_not_authorized 'application key is missing'
   end
 
-  test_authrep 'does not authorize if application key is defined but wrong one is passed' do |e|
-    @application.create_key('foo')
+  test_authrep 'authorizes if application key is defined but not passed if the service is oauth' do |e|
+    @application_oauth.create_key
 
     get e, :provider_key => @provider_key,
-           :app_id       => @application.id,
-           :app_key      => 'bar'
+           :service_id   => @service_oauth.id,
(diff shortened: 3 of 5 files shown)


lib/3scale/backend/transactor.rb
```diff
@@ -61,6 +61,8 @@ def utilization(service_id, application_id)
 
       def validate(oauth, provider_key, report_usage, params, request_info)
         service = Service.load_with_provider_key!(params[:service_id], provider_key)
+        oidc_service = !oauth && service.backend_version == 'oauth'.freeze
```
davidor, line 64: Related to my other comment. I think the distinction between oidc and oauth is confusing.
I think we need a method that tells us if we need to apply the oauth rules (oauth endpoint or backend_version == 'oauth') or something similar.

test/integration/authrep/application_keys_test.rb
```diff
@@ -12,80 +12,126 @@ def setup
...
+
+    get e, :provider_key => @provider_key,
+           :service_id   => @service_oauth.id,
+           :app_id       => @application_oauth.id,
+           :app_key      => application_key
+
+    assert_authorized
+  end
+
+  test_authrep 'does not authorize if application key is defined and passed even if the service is oauth',
```
davidor, line 93: I think this description needs to mention that the key passed is invalid.

davidor: Also, I don't understand the `except: :oauth_authrep` part :thinking:

test/integration/authrep/application_keys_test.rb
```diff
@@ -12,80 +12,126 @@ def setup
...
+           :service_id   => @service_oauth.id,
+           :app_id       => @application_oauth.id,
+           :app_key      => 'invalid_key'
+
+    assert_not_authorized 'application key "invalid_key" is invalid'
+  end
+
   test_authrep 'does not authorize if application key is defined but wrong one is passed' do |e|
-    @application.create_key('foo')
+    @application_oauth.create_key 'foo'
```
davidor, line 106: bye parens :crying_cat_face:

test/integration/backend_version_test.rb
```diff
@@ -97,4 +97,68 @@ def setup
       assert_not_authorized 'application key is missing'
     end
   end
+
+  test 'when backend_version is 2 and we switch to oauth, there is no complaint about missing app_keys when calling auth endpoints' do
```
davidor, line 101: These tests make me wonder if Porta should delete the existing app keys when a service switches to oauth :thinking:

lib/3scale/backend/transactor.rb
```diff
@@ -70,12 +72,18 @@ def validate(oauth, provider_key, report_usage, params, request_info)
         # significant.
         params[:app_id] = nil if app_id && app_id.empty?
 
-        if oauth
-          raise ApplicationNotFound.new nil if app_id.nil?
-          validators = Validators::OAUTH_VALIDATORS
-        else
-          validators = Validators::VALIDATORS
-        end
+        # While OIDC without an app_id makes little sense, we would break existing
```
davidor, line 75: I'm confused by this description. I think what it's saying does not match what the line below does.

test/unit/transactor_test.rb
```diff
@@ -445,14 +454,14 @@ def setup
   test_authrep 'does not include usages of 0 when it generates a report job' do |_ , method|
     current_time = Time.now
     metric_name = 'some_metric'
-    Metric.save(service_id: @service_id, id: next_id, name: metric_name)
+    Metric.save(service_id: @service_oauth_id, id: next_id, name: metric_name)
```
davidor, line 457: Why did you decide to use an oauth service in these 2 tests specifically?

lib/3scale/backend/validators.rb
```diff
@@ -21,6 +21,13 @@ module Validators
       OAUTH_VALIDATORS = ([Validators::OauthSetting,
                            Validators::OauthKey,
                            Validators::RedirectURI] + COMMON_VALIDATORS).freeze
+
+      # OIDC specific validators will only check app keys when app_key is given.
+      #
+      # No need to add OauthSetting, since we need to check that to tell
+      # OIDC apart from the rest when calling authrep.xml (note lack of
+      # the oauth_ prefix).
+      OIDC_VALIDATORS = ([Validators::OauthKey] + COMMON_VALIDATORS).freeze
```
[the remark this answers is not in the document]
davidor: Related to https://github.com/3scale/apisonator/pull/280#discussion_r645386500

lib/3scale/backend/transactor.rb
davidor: Related to https://github.com/3scale/apisonator/pull/280#discussion_r644708714

lib/3scale/backend/validators.rb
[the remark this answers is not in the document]
unleashed: Apisonator does not know anything about OIDC: notice that OIDC does not have any additional information like referrer uris in this phase nor does it (it should not) check any secrets - it all happens before 3scale sees it. With OAuth (and more so when using the removed token infrastructure) you had some workflows in which you needed checks at this level (ie. 3scale was acting as a poor-man's identity provider and verifier). With OIDC 3scale cannot do that, and it does not get to see any extra details like whatever is encoded in the JWTs or other administrative settings in the application, which belong in the IdP.

So the `OauthKey` validator above is not really needed, it should be gone, but it is kept so that the current behavior regarding the checks when calling authorize and authrep is maintained. That is, the change is limited to avoid checking application keys and not removing questionable previous behavior to avoid potentially breaking (wrong) usages.

Another thing is removing the OAuth specific validators that no longer make sense if 3scale is only handling user_key (btw, the docs say user_key is deprecated...), app_id/app_key and OIDC (app_id).

lib/3scale/backend/transactor.rb
unleashed: Well, this is that method except it is not extracted out. The long term change would be to drop completely oauth_auth*.xml and the OAuth validators and treat everything in authorize and authrep as `user_key` and`app_id` (OIDC) on one hand, which don't need to check any app_keys, and `app_id` _with_ `app_key` on the other hand, which does require it.

The actual distinction may be made by checking the `backend_version` of the service or by deciding whether the lack of an `app_key` parameter is significative (ie. if the client, which is trusted, does not specify an `app_key` by omitting the parameter, then it can be interpreted as _not wanting_ the check on app keys).

test/integration/authrep/application_keys_test.rb
unleashed: :muscle:

test/integration/authrep/application_keys_test.rb
unleashed: oauth_authrep only works with oauth services (ok) and never checks app_keys (nope, the check is about an invalid app_key but it will just authorize if calling oauth_authrep), so it's pointless to test oauth_authrep.

unleashed: re: the description, yes, should mention it is invalid.

test/integration/backend_version_test.rb
unleashed: What if it switched by accident and it is immediately followed by another switch back to the original?

unleashed: So this ties in with what I said before: OAuth services (OIDC, really) don't need (nor want) app_keys. And they certainly should not store client secrets. :/

test/unit/transactor_test.rb
unleashed: The original tests used oauth services (ie. `@service_id` was indeed a `@service_oauth_id`), so did not want to change existing and unrelated tests.

lib/3scale/backend/validators.rb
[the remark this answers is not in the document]
davidor: All of what you said is true, what I meant is that now, OIDC is the only oauth flow supported, so having OIDC_VALIDATORS and OAUTH_VALIDATORS is confusing.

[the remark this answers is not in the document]
unleashed: Yes, but then we still might want to not break anything - do you know whether there is anyone still using this old/wrong behavior?

test/integration/backend_version_test.rb
davidor: Yeah, that's the issue.

test/unit/transactor_test.rb
davidor: But there are a bunch of tests here and you only changed the last two. I don't see why the tests in this file should use an oauth service.

unleashed: This may be a result of me initially breaking the old/wrong behavior and then trying to keep it to avoid a round of testing whether changing behavior broke any real usage... the RTT might be measured in years.

unleashed: (that is, we can probably drop this diff, will undo this)

lib/3scale/backend/validators.rb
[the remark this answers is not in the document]
davidor: Right, I forgot about that `RedirectURI` validator :thinking: 
I don't know if that's still used. It should not, but we should check carefully before getting rid of it.

davidor:
@unleashed the PR looks good to me, but it doesn't solve all the issues mentioned in the PR description.

First, I think services configured with OIDC should have been using oauth_* endpoints and not the other ones. I guess we didn't document and communicate that well enough.

Also, if Porta is storing app keys for OIDC services, there's something wrong and we should open an issue to discuss that. As you said, that does not make much sense to me.

Lastly, we should investigate whether we can delete the Redirect validator. I suspect we can, but we should check first. Can you open an issue for that? :pray: I think that would allow us to simplify some code.

davidor (approved):

unleashed:
Rebased on top of current master to pick up the test helper to have an OAuth service fixture.

bors r=@davidor

PR #283, opened 2021-06-07
Avoid duplicating definition of the Redis key that contains the alerts configured

davidor:
This PR gets rid of a duplication. The Redis key that contains the alerts configured was defined both in `AlertLimit` and `Alert`.


unleashed (approved):
bors r+

PR #284, opened 2021-06-08
List application keys extension

### LINKED ISSUE #282: Add extension to list application keys
This extension would be useful for caches, much like `hierarchy`, to avoid contacting 3scale when they find an application key they don't know anything about.

Without this extension caches need to contact 3scale even if they know an app is within limits and otherwise it's been authorized before with a different app key, because this new application key might actually exist in the database _or not_ and then the request should be rejected. If a user kept calling a cache with different app keys, a cache would be forced to keep contacting 3scale.

With this extension a cache can take the opportunity in which they learn about metric hierarchy to also list the set of accepted application keys. Like the hierarchy, this information would then be periodically retrieved to pick up any updates.

Security-wise there is no privilege boundary crossed, since a cache already has full access to a 3scale account via the Porta APIs.

unleashed:
This implements the `list_app_keys` extension, which outputs the `app_keys` XML section:

```xml
<app_keys app="app_id" svc="service_id">
  <key id="key_1"/>
  <key id="key_2"/>
  <key id="key_3"/>
</app_keys>
```

The fact that the `app_id` and `service_id` are also output helps with another request from the caching side to help the implementation a bit by not forcing a client to remember about the app_id/service_id the response refers to (ie. in async apps this means two less strings need to be saved across IO contexts).

/cc @rahulanand16nov @NomadXD

Closes #282.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -35,6 +35,33 @@ checking to ensure data is consistent, ie. usage data with `n` incremented in
 5 necessarily means `n` parent, if existing, should be specified with a minimum
 of 5, not counting any other potential children.
 
+### list_app_keys (integer)
+
+This extension just outputs an XML `app_keys` section in authorization responses
+so that application keys registered to it will be listed there with the following
+format:
+
+> \<app_keys app="app_id" svc="service_id"\>
+>   \<key id="app_key1"/\>
+>    ...
+>   \<key id="app_keyN"/\>
+> \</app_keys>
+
+Applications that don't have associated keys will just show an empty `app_keys`
+section. Also do note that at the time of writing there is a maximum of 256
+such keys listed, and there is no guarantee that two successive calls will end
+up returning the same 256 keys for larger sets, nor the same order in any case.
+
+#### Accepted values
+
+- 0: disable the extension (default)
+- 1: enable the extension, always output the app_keys section regardless of
+     whether an application actually has any associated keys.
+
+Values over 1 are currently undefined, but in the future they might become
+significant. For example they could provide the maximum number of desired
+application keys returned.
+
 ### no_body (boolean)
 
 This instructs the software to avoid generating response bodies for certain endpoints.
--- a/lib/3scale/backend/transactor.rb
+++ b/lib/3scale/backend/transactor.rb
@@ -98,8 +98,9 @@ module ThreeScale
           # hierarchy parameter adds information in the response needed
           # to derive which limits affect directly or indirectly the
           # metrics for which authorization is requested.
-          hierarchy:       extensions[:hierarchy] == '1',
-          flat_usage:      extensions[:flat_usage] == '1'
+          hierarchy:       extensions[:hierarchy] == '1'.freeze,
+          flat_usage:      extensions[:flat_usage] == '1'.freeze,
+          list_app_keys:   extensions[:list_app_keys] == '1'.freeze
         }
 
         application.load_metric_names
--- a/lib/3scale/backend/transactor/status.rb
+++ b/lib/3scale/backend/transactor/status.rb
@@ -8,17 +8,23 @@ module ThreeScale
         # We only use 'redirect_uri' if a request sent such a param. See #397.
         REDIRECT_URI_FIELD = 'redirect_url'.freeze
         private_constant :REDIRECT_URI_FIELD
+        # Maximum number of keys to list when using the list_app_keys extension
+        # At the time of writing System/Porta has a limit of 5 different app_keys
+        # at any given moment, but this could change anytime.
+        LIST_APP_KEYS_MAX = 256
+        private_constant :LIST_APP_KEYS_MAX
 
         def initialize(attributes)
-          @service_id      = attributes[:service_id]
-          @application     = attributes[:application]
-          @oauth           = attributes[:oauth]
-          @usage           = attributes[:usage]
-          @predicted_usage = attributes[:predicted_usage]
-          @values          = filter_values(attributes[:values] || {})
-          @timestamp       = attributes[:timestamp] || Time.now.getutc
-          @hierarchy_ext   = attributes[:hierarchy]
-          @flat_usage_ext  = attributes[:flat_usage]
+          @service_id        = attributes[:service_id]
+          @application       = attributes[:application]
+          @oauth             = attributes[:oauth]
+          @usage             = attributes[:usage]
+          @predicted_usage   = attributes[:predicted_usage]
+          @values            = filter_values(attributes[:values] || {})
+          @timestamp         = attributes[:timestamp] || Time.now.getutc
+          @hierarchy_ext     = attributes[:hierarchy]
+          @flat_usage_ext    = attributes[:flat_usage]
+          @list_app_keys_ext = attributes[:list_app_keys]
 
           raise 'service_id not specified' if @service_id.nil?
           raise ':application is required' if @application.nil?
@@ -106,6 +112,7 @@ module ThreeScale
             add_plan_section(xml, 'plan'.freeze, plan_name)
             add_reports_section(xml, application_usage_reports)
             hierarchy_reports.concat application_usage_reports if hierarchy_reports
+            add_app_keys_section xml if @list_app_keys_ext
           end
 
           if hierarchy_reports
@@ -159,6 +166,17 @@ module ThreeScale
             xml << (children ? children.join(' '.freeze) : '') << '"/>'.freeze
           end
           xml << '</hierarchy>'.freeze
+        end
+
+        def add_app_keys_section(xml)
+          xml << '<app_keys app="'.freeze
+          xml << @application.id << '" svc="'.freeze
+          xml << @service_id << '">'.freeze
+          @application.keys.take(LIST_APP_KEYS_MAX).each do |key|
+            xml << '<key id="'.freeze
+            xml << key << '"/>'.freeze
+          end
+          xml << '</app_keys>'.freeze
         end
 
         # helper to iterate over reports and get relevant hierarchy info
--- a/test/integration/authorize/list_app_keys_test.rb
+++ b/test/integration/authorize/list_app_keys_test.rb
@@ -1,0 +1,220 @@
+require File.expand_path(File.dirname(__FILE__) + '/../../test_helper')
+
+class AuthorizeListAppKeysExtensionTest < Test::Unit::TestCase
+  include TestHelpers::Fixtures
+  include TestHelpers::Integration
+  include TestHelpers::Extensions
+
+  def setup
+    @storage = Storage.instance(true)
+    @storage.flushdb
+
+    Resque.reset!
+    Memoizer.reset!
+
+    setup_provider_fixtures
+    setup_oauth_provider_fixtures_noclobber
+
+    @application = Application.save(service_id: @service.id,
+                                    id: next_id,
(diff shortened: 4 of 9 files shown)


lib/3scale/backend/transactor/status.rb
```diff
@@ -8,17 +8,21 @@ class Status
         # We only use 'redirect_uri' if a request sent such a param. See #397.
         REDIRECT_URI_FIELD = 'redirect_url'.freeze
         private_constant :REDIRECT_URI_FIELD
+        # Maximum number of keys to list when using the list_app_keys extension
```
davidor, line 11: To avoid confusion, I think we should also clarify that system only allows to create 5.

lib/3scale/backend/transactor.rb
```diff
@@ -99,7 +99,8 @@ def validate(oauth, provider_key, report_usage, params, request_info)
           # to derive which limits affect directly or indirectly the
           # metrics for which authorization is requested.
           hierarchy:       extensions[:hierarchy] == '1',
-          flat_usage:      extensions[:flat_usage] == '1'
+          flat_usage:      extensions[:flat_usage] == '1',
```
davidor, line 102: freeze?

docs/extensions.md
```diff
@@ -35,6 +35,33 @@ checking to ensure data is consistent, ie. usage data with `n` incremented in
 5 necessarily means `n` parent, if existing, should be specified with a minimum
 of 5, not counting any other potential children.
 
+### list_app_keys (integer)
+
+This extension just outputs an XML `app_keys` section in authorization responses
+so that application keys registered to it they will be listed there with the
```
davidor, line 41: registered to it ~they~ will

docs/extensions.md
```diff
@@ -35,6 +35,33 @@ checking to ensure data is consistent, ie. usage data with `n` incremented in
...
+following format:
+
+> \<app_keys app="app_id" svc="service_id"\>
+>   \<key id="app_key1"/\>
```
davidor, line 45: Why the `id` field ?

test/unit/transactor/status_test.rb
```diff
@@ -249,5 +249,21 @@ def setup
...
+                                      values: usage,
+                                      oauth: true)
+
+      doc = Nokogiri::XML(status.to_xml)
+
+      keys = doc.at 'app_keys'
+      assert keys
+      assert_equal @application.id, keys['app']
+      assert_equal @service_id, keys['svc']
+    end
```
davidor, line 267: Would be good to test that the keys are there.

test/integration/authorize/list_app_keys_test.rb
```diff
@@ -0,0 +1,132 @@
...
+                                           plan_id: @plan_id,
+                                           plan_name: @plan_name)
+  end
+
+  test 'calling authorize without the list_app_keys extension does not output the application keys section' do
+    get '/transactions/authorize.xml',
+        { provider_key: @provider_key, app_id: @application.id }
+
+    xml_resp = Nokogiri::XML(last_response.body)
+    assert xml_resp.at('app_keys').nil?
```
davidor, line 41: `assert_nil` ?

test/integration/authorize/list_app_keys_test.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    xml_resp = Nokogiri::XML(last_response.body)
+    app_keys = xml_resp.at 'app_keys'
+    assert !app_keys.nil?
+    assert_equal @application.id, app_keys['app']
+    assert_equal @service.id, app_keys['svc']
+
+    keys = xml_resp.search("app_keys key").map { |n| n[:id] }
+
+    assert_equal @keys, keys.sort
```
davidor, line 73: sort `@keys` too, so we don't have to look at the beginning of the file to check whether it's sorted or not.

test/integration/authorize/list_app_keys_test.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    xml_resp = Nokogiri::XML(last_response.body)
+    app_keys = xml_resp.at 'app_keys'
+    assert !app_keys.nil?
+    assert_equal @application_nokeys.id, app_keys['app']
+    assert_equal @service.id, app_keys['svc']
+
+    keys = xml_resp.search("app_keys key").map { |n| n[:id] }
+
+    assert keys.empty?
```
davidor, line 57: `assert_empty` ?

test/integration/authorize/list_app_keys_test.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    xml_resp = Nokogiri::XML(last_response.body)
+    app_keys = xml_resp.at 'app_keys'
+    assert !app_keys.nil?
+    assert_equal @application.id, app_keys['app']
+    assert_equal @service.id, app_keys['svc']
+
+    keys = xml_resp.search("app_keys key").map { |n| n[:id] }
+
+    assert_equal @keys, keys.sort
```
davidor, line 90: ditto

test/integration/authorize/list_app_keys_test.rb
```diff
@@ -0,0 +1,132 @@
...
```
davidor, line 107: ditto

test/integration/authorize/list_app_keys_test.rb
davidor: There are other asserts where `assert_nil` could be used.

test/integration/authorize/list_app_keys_test.rb
```diff
@@ -0,0 +1,132 @@
...
+    app_keys = xml_resp.at 'app_keys'
+    assert !app_keys.nil?
+    assert_equal @application.id, app_keys['app']
+    assert_equal @service.id, app_keys['svc']
+
+    keys = xml_resp.search("app_keys key").map { |n| n[:id] }
+
+    assert_equal max_keys, keys.size
+  end
+end
```
davidor, line 132: Should we test that this does not return anything when service and app_id are valid but the provider key is not?

test/integration/authrep/list_app_keys_test.rb
```diff
@@ -0,0 +1,131 @@
+require File.expand_path(File.dirname(__FILE__) + '/../../test_helper')
```
davidor, line 1: This is basically the same as the other file, check if my comments apply to this one too.

test/integration/authorize/list_app_keys_test.rb
davidor, line 132: It'd be good to add tests for services configured with user_key and oidc to verify that enabling the extension does not break anything and does not return anything.

lib/3scale/backend/transactor.rb
unleashed: yes, we should be freezing the literals :)

docs/extensions.md
unleashed: :+1:

test/integration/authorize/list_app_keys_test.rb
unleashed: Ok, will take a look!

test/integration/authorize/list_app_keys_test.rb
unleashed: Hm, this one is quite subjective: this can only consistently succeed if `@keys` is sorted, that's kind of the premise, and sorting something that can only be sorted to succeed in the general case means more repetitions and more runtime for tests :).

That said... if you still think it's worth it to repeat it... just insist on it here (technically: _repeat_ it to me ;). :neckbeard:

test/integration/authorize/list_app_keys_test.rb
unleashed: I don't think so - they are unrelated circumstances that should have no connection (and the related ones should be tested already). If we went that route then we could start a explosion of test combinations all over the place that would arguably provide little value.

test/integration/authrep/list_app_keys_test.rb
unleashed: I think they do.

test/integration/authorize/list_app_keys_test.rb
unleashed: The idea is that the extension always returns something - at the very least the app and service the auth refers to, even if the app is user_key or oauth. We can test that at least once, but I suspect we already do somewhere else with an oauth application. Hmm.

docs/extensions.md
davidor: My question was more about needing the attribute. I guess you did it like this because it allows to add more attributes in the future, although I don't think that's going to happen.
It's fine :+1:

lib/3scale/backend/transactor/status.rb
davidor: Yes, it could change anytime. We can start the comment with "At the time of writing this... " :writing_hand: 
But I think it's important to clarify.

test/integration/authorize/list_app_keys_test.rb
davidor: You decided that `@keys` is sorted, but for your tests you could have chosen any other thing (adapting the assert).

My point is, when comparing arrays I think it's better for readability to sort both because that works always. If it's very clear that they're sorted this might not be needed.

When you get to this line of the assert and you see that `@keys` is not sorted, now you have to go back to the line that defines it to see if it is or not :neckbeard:

test/integration/authorize/list_app_keys_test.rb
davidor: It does not make much sense to enable the extension for services configured with user_key, but I think that we should test that anyway and check that the list is empty and no exceptions are raised.

For OIDC is a bit more complicated. Services with OIDC should not have any app keys defined. However, because of what we discussed in a previous PR, it seems that Porta is storing them in apisonator. Should we return them? I don't think we should because it's not useful for caching anyway. The cache should only rely on the app ID for services that use OIDC. Whatever we decide, we should add a test for it.

lib/3scale/backend/transactor.rb
davidor: :cold_face:

docs/extensions.md
unleashed: Well, not exactly - the attributes indeed help extensibility. But I used it to be able to have a self-closing tag rather than:

> \<key\>some_app_key\</key\>

This usually helps parsing a bit and is less verbose. In this case attributes can still be added later, so that's not a factor.

test/unit/transactor/status_test.rb
unleashed: In this case there are no keys, but I've changed to test that... there are no keys :). I will create another test with keys.

docs/extensions.md
davidor: ok :+1:

test/integration/authorize/list_app_keys_test.rb
unleashed: With user_key you can still make use of the extension to get the `app` and `svc` attributes of the response (`user_key` will map to an `app_id`), but admittedly a cache would just use directly the `user_key` value as a sort of `app_id` for caching.

For OIDC we are returning them and there is now a test for it.

unleashed:
@davidor this should now be a slam dunk :neckbeard::basketball: :basket:

davidor (approved):

PR #287, opened 2021-06-14
Optimize alerts detection

davidor:
Currently, the code is loading all the usage limits of a plan and the current usages for those limits on every report job.

This PR introduces a new class (`UsageChecked`) that allows us to do just once and then, no longer do it for the same application until:
1) An alert expires (24h passed since it was triggered) and can be raised again.
2) A new alert bin is enabled for the service.

With this, on every report job we just need to check the limits and usages of the metrics increased in the job. This should have a noticeable impact for users with many usage limits defined.


lib/3scale/backend/alerts.rb
```diff
@@ -45,6 +50,69 @@ def key_current_id
...
+        def self.need_to_check_all?(service_id, app_id)
+          !storage.exists(key_usage_already_checked(service_id, app_id))
+        end
+        memoize :need_to_check_all?
+
+        def self.mark_all_checked(service_id, app_id)
+          ttl = ALERT_BINS.map do |bin|
+            ttl = storage.ttl(key_already_notified(service_id, app_id, bin))
+
+            # Redis returns -2 is key does not exist, and -1 if it exists without
```
unleashed, line 77: typo `s/is/if`

test/unit/alerts_test.rb
```diff
@@ -46,4 +49,69 @@ def setup
...
+    key_bin_80 = Alerts.send(:key_already_notified, @service_id, @application_id, 80)
+    @storage.setex(key_bin_80, min_ttl, '1')
+
+    Alerts::UsagesChecked.mark_all_checked(@service_id, @application_id)
+
+    assert_false Alerts::UsagesChecked.need_to_check_all?(@service_id, @application_id)
+
+    # The 80% bin has a lower TTL. We cannot guarantee the exact TTL because
+    # some time might pass between the moment we create the key and check its
+    # TTL, but we know it needs to be <= min_ttl and close to it.
```
unleashed, line 70: So maybe just ensure the offset is something bigger than a time difference since before you create the key and right before you check its TTL to reduce the odds of spurious failures?

test/unit/alerts_test.rb
```diff
@@ -46,4 +49,69 @@ def setup
...
+  end
+
+  test 'when marking as checked, sets the key with TTL = 1 day if none of the bins are set as notified' do
+    Alerts::UsagesChecked.mark_all_checked(@service_id, @application_id)
+
+    assert_false Alerts::UsagesChecked.need_to_check_all?(@service_id, @application_id)
+
+    # We can't guarantee that the TTL will be exactly one day because some
+    # seconds might pass between the moment we create the key and check the TTL,
+    # but it should be pretty close.
```
unleashed, line 82: Ditto.

lib/3scale/backend/stats/aggregator.rb
```diff
@@ -141,14 +141,23 @@ def update_alerts(transactions)
...
+              # loading usage limits and current usages. That's why before that,
+              # we check if there are any alerts that can be raised.
               next unless Alerts.can_raise_more_alerts?(service_id, app_id)
 
               begin
-                max_utilization = Utilization.max_in_all_metrics(service_id, app_id)
+                max_utilization = if Alerts::UsagesChecked.need_to_check_all?(service_id, app_id)
+                                    max = Utilization.max_in_all_metrics(service_id, app_id)
+                                    Alerts::UsagesChecked.mark_all_checked(service_id, app_id)
+                                    max
```
unleashed, line 153: Great opportunity to `tap` a Ruby idiom:

```ruby
Utilization.max_in_all_metrics(service_id, app_id).tap do
  Alerts::UsagesChecked.mark_all_checked service_id, app_id
end
```

unleashed (approved):
LGTM but check comments!

test/unit/alerts_test.rb
davidor: I think that's difficult to get exactly right, that's why I used a range instead.

lib/3scale/backend/alerts.rb
davidor: fixed :+1:

lib/3scale/backend/stats/aggregator.rb
davidor: Changed :+1:

lib/3scale/backend/stats/aggregator.rb
```diff
@@ -141,14 +141,23 @@ def update_alerts(transactions)
...
-                max_utilization = Utilization.max_in_all_metrics(service_id, app_id)
+                max_utilization = if Alerts::UsagesChecked.need_to_check_all?(service_id, app_id)
+                                    Utilization.max_in_all_metrics(service_id, app_id).tap do
+                                      Alerts::UsagesChecked.mark_all_checked(service_id, app_id)
+                                    end
+                                  else
+                                    # metrics_ids here includes the metrics
+                                    # explicitly reported plus their parents in
+                                    # the hierarchy.
+                                    metric_ids = txs.map { |tx| tx.usage.keys }.flatten.uniq
```
unleashed, line 158: Hm, didn't a test fail because of lacking `flatten` here?

davidor: No, we need more tests in this part. I'm working on it.

davidor:
@unleashed I added a new commit with the tests that were missing. It's the last commit.
I also added a `Memoizer.clear` call here: https://github.com/3scale/apisonator/pull/287/files#diff-b176f67b39cc7bd8b50fe54aa2a4de9cf404809e691aab4a2199d6aa8dc10011R84

lib/3scale/backend/alerts.rb
```diff
@@ -45,43 +50,82 @@ def key_current_id
...
-          if item.max_value > 0
-            utilization = item.current_value / item.max_value.to_f
-
-            if utilization > max_utilization
-              max_record = item
-              max_utilization = utilization
-            end
-          end
+          storage.setex(key_usage_already_checked(service_id, app_id), ttl, '1'.freeze)
+          Memoizer.clear(Memoizer.build_key(self, :need_to_check_all?, service_id, app_id))
```
unleashed, line 84: Rather than clearing you could just set the new memoized value for `:need_to_check_all?` (would avoid one command per memorization cycle, so no big deal).

unleashed (approved):

test/unit/alerts_test.rb
davidor: https://github.com/3scale/apisonator/pull/287#discussion_r650953996

unleashed:
bors delegate+

lib/3scale/backend/alerts.rb
davidor: That would increase complexity a bit in my opinion.
I don't think there's a clear winner in this case, bud I'd rather leave it as it is now.

PR #290, opened 2021-06-17
Handle TTL = 0 in UsagesChecked.mark_all_checked

davidor:
The Redis setex command fails when ttl = 0 and we were not handling that case.
The test included in this PR fails in master.


unleashed (approved):
Let's Go To Mexico (this time around at least one of us did!!)

bors r+

PR #292, opened 2021-06-21
Fix bug that causes Apisonator to stop pinging System to fetch events

davidor:
The current code fails if the incr call is made effective in Redis but the client returns an error because of a timeout. When that happens, which should be pretty rare, apisonator stops sending events because from that moment, `ping_key_value.to_i == 1` in https://github.com/3scale/apisonator/blob/677dfbf16346bf1efebdf4fec5b5e0a9d72e8ce1/lib/3scale/backend/event_storage.rb#L99 will always be false.

### DIFF
--- a/spec/unit/event_storage_spec.rb
+++ b/spec/unit/event_storage_spec.rb
@@ -1,3 +1,4 @@
+require 'pry'
 module ThreeScale
   module Backend
     describe EventStorage do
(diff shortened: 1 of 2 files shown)


spec/unit/event_storage_spec.rb
```diff
@@ -189,8 +190,10 @@ module Backend
           context 'when a ping was executed previously' do
             context 'and ping TTL is expired' do
               before do
-                stub_const("ThreeScale::Backend::EventStorage::PING_TTL", 0)
```
davidor, line 192: @unleashed This is a bit unexpected, but as I mention in the commit description, the `expire` command accepts 0 as the TTL, but `set` with TTL does not.

unleashed: :+1:

unleashed (approved):
bors r+

PR #297, opened 2021-07-02
Revert #295

davidor:
Reverts #295 
The reason is that the version of bundler is not installed in some of our images. We need to update them before updating bundler.


unleashed (approved):
bors r+

PR #302, opened 2021-07-12
Async storage: read all responses in pipelines even if some of them are errors

davidor:
Redis returns an answer for each of the commands sent in the pipeline, but we were not reading all of them. We stopped at the first error. This PR fixes the issue.


unleashed (approved):
bors r+

PR #304, opened 2021-07-13
Fix Openshift HPA setup instructions

davidor:
We had some examples with CPU target usage >= 100% which only works if there are enough resources in the cluster.


unleashed (approved):
bors r+

PR #305, opened 2021-07-21
add .tools-versions and ignore .env file

akostadinov:
follow-up to #279

in case Apisonator project prefers tools versions to be in source,
then also adding `.tools-versions` which is used by the [asdf](https://asdf-vm.com/) version manager.

Ignoring `.env` as my understanding is that it is for local purposes and may differ from setup to setup. So avoiding to be comitted by mistake.


unleashed:
@akostadinov I can't figure out what is it that this PR is fixing, so let me ask you a few questions below so that I can better understand it.

The information this new `.tool-versions` file is conveying already exists for the very same purpose in `.ruby-version`, which is the one we work with and standardized for many years across different version managers. If you work with `asdf`, the real fix would be for that manager to read `.ruby-version` and honor its values. Quite frankly, if this `asdf` manager is popular I'd be surprised it does not respect this standard file. Can you make it work as-is with `.ruby-version`? In addition, ideally, you would never run anything locally but in the development environment, which is a container that is already set up for you with the right Ruby version.

Regarding the `.env` file I don't understand your use case. Where are you using it that requires sourcing it? I've never had this need for the project before. Perhaps those needs are addresses once you run tests in the `make dev` development container?

akostadinov:
@unleashed , thank you for looking at this.

asdf-vm is great for versioning many tools, not only ruby. `.env` is used by different tools for local environment setup when you work in a particular directory. I'm running apicast in container, importing project in IDE though is local.

The ultimate issue I'm trying to fix is to avoid git showing these files as new locally. If I add the two of them to `.gitignore` would that be accepted?

hallelujah:
As for 3scale/porta we do not add in git the .tool-versions in the contrary it is ignored https://github.com/3scale/porta/blob/c78c49870c5c421d3a7b522fa3445e1a039f3d3a/.gitignore#L30

Though we have a sample in porta, it is not relevant in apicast.

hallelujah:
@unleashed yes apisonator... to many IPA
Thanks for the local ignore tip!

akostadinov:
I see that asdf-vm supports that with an additional setting inside `~/.asdfrc`. Nice to know. Also nice tip about gitignore. Thanks.

This works for me so I'm closing PR.

PR #307, opened 2021-08-30
bors.toml: remove codeclimate from pr_status

unleashed:
It is problematic when reverting PR's that improve the CC rating.


unleashed:
bors retry

PR #317, opened 2021-11-18
Add s390x support to ci-build

DaniilRomanov:
When running `make ci-build` it uses `Dockerfile.ci` based on CentOS image, but there is no CentOS for s390x architecture. New `Dockerfile.ci.s390x` is made from original one and based on Ubuntu. One can run `make ci-build-s390x`, it successfully builds an image for s390x and this image passes all tests by `make DOCKER_OPTS="-e TEST_ALL_RUBIES=1" test`.

### DIFF
--- a/Dockerfile.ci.s390x
+++ b/Dockerfile.ci.s390x
@@ -1,0 +1,297 @@
+FROM ubuntu
+MAINTAINER Alejandro Martinez Ruiz <amr@redhat.com>
+
+ARG CIRCLECI_TOOLS="git openssh-client tar gzip ca-certificates locales locales-all"
+# set localtime to UTC and install basic build tools
+RUN (ln -s /usr/share/zoneinfo/UTC /etc/localtime 2> /dev/null || true) \
+    && apt update -y \
+    && apt install -y sudo git make gcc autoconf automake libtool gettext file \
+       ${CIRCLECI_TOOLS} \
+    && apt clean all \
+    && localedef -c -f UTF-8 -i en_US en_US.UTF-8
+
+ARG USER_NAME=ruby
+RUN addgroup wheel
+RUN useradd -m -d "/home/${USER_NAME}" -l -G wheel ${USER_NAME} \
+    && echo "${USER_NAME} ALL=(ALL:ALL) NOPASSWD: ALL" > \
+    "/etc/sudoers.d/${USER_NAME}" \
+    && chmod 0440 "/etc/sudoers.d/${USER_NAME}"
+
+USER "${USER_NAME}"
+
+# PostgreSQL installation
+ARG POSTGRES_BUILD_DEPS="bison flex zlib1g-dev libreadline-dev libssl-dev"
+RUN test "x${POSTGRES_BUILD_DEPS}" = "x" \
+    || sudo apt install -y ${POSTGRES_BUILD_DEPS}
+
+ARG POSTGRES_VERSION=REL_10_17
+RUN curl -sSLf "https://github.com/postgres/postgres/archive/${POSTGRES_VERSION}.tar.gz" \
+    | tar -C "/home/${USER_NAME}" -xz
+
+ARG POSTGRES_PREFIX=/usr/local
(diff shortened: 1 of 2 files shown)


krishvoor:
Hi @eguzki  - request for your review.

eguzki:
I will leave to @unleashed

Dockerfile.ci.s390x
```diff
@@ -0,0 +1,297 @@
...
+
+ARG MRI_DEPS="bzip2 zlib1g-dev libreadline-dev libssl-dev libffi-dev libgdbm-dev libncurses-dev"
+RUN test "x${MRI_DEPS}" = "x" || sudo apt install -y ${MRI_DEPS}
+
+RUN echo 'gem: --no-document' >> ~/.gemrc \
+    && echo --color > ~/.rspec
+
+ARG CONFIGURE_OPTS=--disable-install-doc
+RUN mkdir -p ~/.local/bin \
+    && echo  "#!/bin/bash\n\n" \
```
unleashed, line 133: These `echo` commands originally used the `-e` switch - is it not working correctly for you?

I just ran (on x86_64) the following, which shows `-e` to work well and removing it to quote the `\n`s rather than interpreting them:
```shell
apisonator$ podman run --rm -ti ubuntu /bin/bash -c 'echo -e "hello\n"; echo "hello\n"'
hello

hello\n
apisonator$
```

Dockerfile.ci.s390x
```diff
@@ -0,0 +1,297 @@
...
+       > ~/.local/bin/bundle_install_rubies \
+    && chmod +x ~/.local/bin/bundle_install_rubies
+
+COPY .ruby-* /tmp/apisonator/
+RUN sudo chown -R "${USER_NAME}": /tmp/apisonator \
+    && cd /tmp/apisonator \
+    && rbenv_update_env
+
+# specify all versions to be installed, partial versions also understood
+ARG RUBY_VERSIONS="2.7.0"
```
unleashed, line 278: The original `Dockerfile.ci` file uses `MINOR.MAJOR` here to test against the latest patch release of MINOR.MAJOR, but this will pin the Ruby version to 2.7.0 - ie. it won't take 2.7.3, which is the original intention. Did you find any issues with this or otherwise preferred to pin the version to exactly 2.7.0?

Dockerfile.ci.s390x
```diff
@@ -0,0 +1,297 @@
...
+    "${POSTGRES_PREFIX}/bin/postgres -D '${POSTGRES_DATA_PREFIX}/data'" \
+      > /tmp/postgres.log 2>&1 & sleep 5 \
+    && sudo runuser -l postgres -c "${POSTGRES_PREFIX}/bin/createdb test" \
+    && sudo runuser -l postgres -c "${POSTGRES_PREFIX}/bin/psql test"
+
+# Twemproxy installation
+ARG TWEMPROXY_VERSION=0.5.0
+RUN curl -sSLf \
+    "https://github.com/twitter/twemproxy/releases/download/0.5.0/twemproxy-${TWEMPROXY_VERSION}.tar.gz" \
+    | tar -C "/home/${USER_NAME}" -xz
```
unleashed, line 62: I know this has been carried over from `Dockerfile.ci`, but it makes little sense to be able to select the `TWEMPROXY_VERSION` via an argument only to encode the version in the path anyway. That change slipped somehow into the other file, but should be fixed, either by reusing `TWEMPROXY_VERSION` in its stead, or by finding another download link.

unleashed (commented):
Hi @DaniilRomanov, I've commented on the changes I don't understand, please take a look at them - regarding adding another file, I would like to understand what broke for you from the current `Dockerfile.ci`? I _assume_ it must be a lack of support for `s390x` in CentOS container images, but would like to have confirmation.

OTOH, CentOS Stream 9 has/will have support for s390x, so perhaps we could explore the option of using pre-release images to avoid having different Dockerfiles depending on the arch, although I've never tested this and not sure container images are available for Stream.

DaniilRomanov:
I accept the notices and close current PR to re-do it.

PR #319, opened 2022-02-09
Update Dockerfile.ci to centos:stream9

DaniilRomanov:
Changes were made to make ci-build run on s390x platform.
The root change is using centos:stream9 image, because centos:8 doesn't support s390x, other changes are just consequential.


krishvoor:
FYR @eguzki @unleashed ^^^

eguzki (approved):

PR #321, opened 2022-03-29
Created .travis.yml

DaniilRomanov:
Created `.travis.yml` file for testing with Travis-CI, currently for s390x architecture only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+os: linux
+dist: 
+  - focal
+arch:
+  - s390x
+language: 
+  - shell
+install: 
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+  - echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+  - sudo apt-get update
+  - sudo apt-get install docker-ce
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  - focal
+arch:
+  - s390x
+language: 
+  - shell
+install: 
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+  - echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+  - sudo apt-get update
+  - sudo apt-get install docker-ce
```
krishvoor, line 12: Can we use this instead
```
services:
  - docker
```

krishvoor (commented):
Minor nit

.travis.yml
DaniilRomanov: Unfortunately no. It is already defined for Apisonator to build separate docker image for CI (`Dockerfile.ci`), in this image it compiles specific version of PostgreSQL, and there is a problem with PostgreSQL `./configure` inside an image building and default version of Docker on z at Travis. When I used Docker from Travis environment (version 18), `./configure` stops with an error that it doesn't recognize the shell and can't proceed. When I switched to installing Docker (current version is 20) by official Docker instruction - `./configure` works without errors.

krishvoor (approved):
lgtm

.travis.yml
krishvoor: Thanks for the explanation!

PR #322, opened 2022-10-10
update  deps

eguzki:
Update deps:
* redis
* resque


gsaslis:
@eguzki i think we also need to update the commit sha for resque, after https://github.com/3scale/resque/pull/1

PR #326, opened 2022-12-13
delete user_required attr from app

### LINKED ISSUE #299: Delete "user_required" attr from Application
I don't think it's used anywhere. We probably forgot to delete it when we removed the "end-users" feature.

eguzki:
Fixes #299

### DIFF
--- a/lib/3scale/backend/application.rb
+++ b/lib/3scale/backend/application.rb
@@ -4,8 +4,7 @@ module ThreeScale
       include Storable
 
       # list of attributes to be fetched from storage
-      ATTRIBUTES = [:state, :plan_id, :plan_name, :redirect_url,
-                    :user_required].freeze
+      ATTRIBUTES = [:state, :plan_id, :plan_name, :redirect_url].freeze
       private_constant :ATTRIBUTES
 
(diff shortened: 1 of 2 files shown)


eguzki:
review @thomasmaas ??

gsaslis (approved):
I am not sure of the implications of this change, but CI is green and the changes look relatively straightforward.

lib/3scale/backend/application.rb
```diff
@@ -170,10 +163,6 @@ def with_app_id_from_params(service_id, app_id, user_key)
         end
       end
 
-      def user_required?
```
gsaslis, line 173: My ruby understanding isn't enough to tell, so I'll just ask: 

is this a change to the public API ? i.e. is this method invokable from outside this class, meaning that it could be used by some other program (that I don't have the sources for) ?

eguzki: This method is public, yes. But no one was calling that anymore.

eguzki: well, public in the context of the class. This isn't part of any REST API or public API that can be used outside this project.

PR #325, opened 2022-12-13
Remove saas mode

eguzki:

PR #332, opened 2022-12-19
fix ci image

eguzki:
### what 

CI image was based on `quay.io/centos/centos:stream9` which reached EOL and cannot be used. When trying to build dev image from the CI image, the installation of the packages failed.

This PR depends on #325 because that PR updates the Gemfile, simplifying it. The new Gemfile removes some dependencies, one of them being `pg` which was used to access PostgreSQL. Since `pg` is no longer needed, postgreSQL is not needed in the CI image either. Thus, this PR removes postgreSQL from the CI image. 

The new image is based on UBI 9 `registry.access.redhat.com/ubi9:9.1.0`. 

Dev notes: there were some issues installing postgreSQL dependencies such as `bison`, `flex` and `readline-devel`. Since postgreSQL is no longer needed, the issues were gone.

Once merged, the new CI image must be pushed to `quay.io/3scale/apisonator-ci` to be used in the dev image.

### Verification steps

* build new CI image

```
make ci-build
```

* run dev env

```
make dev
```

* additionally, you can run tests from the container

```
script/ci
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM quay.io/3scale/apisonator-ci
-MAINTAINER Alejandro Martinez Ruiz <amr@redhat.com>
+MAINTAINER Eguzki Astiz Lezaun <eastizle@redhat.com>
 
 ARG APP_HOME
 RUN mkdir "${APP_HOME}"
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,5 +1,5 @@
-FROM quay.io/centos/centos:stream9
-MAINTAINER Alejandro Martinez Ruiz <amr@redhat.com>
+FROM registry.access.redhat.com/ubi9:9.1.0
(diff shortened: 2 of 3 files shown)


gsaslis:
If I Understand Correctly (IIUC), the "fix" here is just the EOL base image, right ?

Otherwise, this seems to just be removing some unneeded dependencies, which looks good to me.

eguzki:
> If I Understand Correctly (IIUC), the "fix" here is just the EOL base image, right ?

That was the motivation of the PR

> Otherwise, this seems to just be removing some unneeded dependencies, which looks good to me.

I had issues installing (with the existing installation method) of postgreSQL. I found out that it is a leftover and no longer used feature, so in the CI image there is no need to have postgresql installed. It was not being used.

eguzki:
> Thanks for the explanation @eguzki !
> 
> Let's Get This Merged (LGTM)

Thanks to you. This information is relevant for later reference and I should have added that before.

PR #337, opened 2022-12-21
Cherry pick saas mode removal into 2.13 stable stream

gsaslis:


eguzki:
no need, I have merged `3scale-2.13-stable` with `master`

gsaslis:
@eguzki doesn't that also add the async-pool update to the 2.13.1 release, meaning one more issue for QE to verify ?

eguzki:
in on-prem, async mode is not enabled. It does not affect in any way.

PR #324, opened 2022-10-17
THREESCALE-7864 async-pool v0.3.12

eguzki:
### what
 Upgrade async deps
*  async 1.24.2 -> 1.26.2
* async-pool 0.2.0 -> 0.3.12

Fixes https://issues.redhat.com/browse/THREESCALE-7864
Fixes https://github.com/3scale/apisonator/issues/308

Performed tests:

| image | async-pool version | Mem leak fixed | [connection dropping issue](https://issues.redhat.com/browse/THREESCALE-7864?focusedCommentId=19394087&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19394087) fixed | 
| --- | --- | --- | --- |
| [latest 2.12](registry.redhat.io/3scale-amp2/backend-rhel8@sha256:437ad353f164871d741dd662015aec221406564c4de68f27c56d3e67c0cf177c) | *0.2.0* |  :x: | :heavy_check_mark: |
| [master-c1889f94-async-pool-0.3.9](quay.io/3scale/apisonator:master-c1889f94-async-pool-0.3.9) | *0.3.9* | :heavy_check_mark: | :x: |
| [master-c1889f94-async-pool-0.3.12](quay.io/3scale/apisonator:master-c1889f94-async-pool-0.3.12) | *0.3.12* |  :heavy_check_mark: | :heavy_check_mark: |

### How check mem leak

* Deploy 3scale with the operator
* Remove redis virtual databases. Change backend-redis secret, and both storage and queue have the same value: "redis://backend-redis:6379"
```
oc rollout latest deploymentconfig/backend-listener
```
* Apply backend worker async
```
oc set env deploymentconfig/backend-worker CONFIG_REDIS_ASYNC=1
```
* 3scale provisioning using buddhi
```
bundle exec perftest-toolkit-buddhi --portal https://TOKEN@3scale-admin.example.com --profile backend --public-base-url "" --private-base-url "http://toystore.my-namespace.svc:80" -o /dev/stdout
```
* Traffic generator (siege)
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: siege
spec:
  replicas: 1
  selector:
    matchLabels:
      app: siege
  template:
    metadata:
      labels:
        app: siege
    spec:
      containers:
        - args:
            - --time=10H
            - --concurrent=16
            - --quiet
[...]

Issue #135, opened 2019-10-15
Async mode not compatible with Redis logical databases


davidor:
The async-redis client does not support Redis logical databases. That means that it does not work properly on environments that have for example, the main db in `redis://redis-backend:6379/0` and the queues in `redis://redis-backend:6379/1`: https://github.com/3scale/3scale-operator/blob/master/pkg/3scale/amp/auto-generated-templates/amp/amp.yml

Ref: https://github.com/socketry/async-redis/issues/15

akostadinov:
There is an example protocol wrapper upstream with TLS and logical databases support.

https://github.com/socketry/async-redis/blob/main/examples/auth/wrapper.rb

Works like a charm for me, like this:
```diff
diff --git a/lib/3scale/backend/storage_async/client.rb b/lib/3scale/backend/storage_async/client.rb
index fc0c0a4a..ebdcb108 100644
--- a/lib/3scale/backend/storage_async/client.rb
+++ b/lib/3scale/backend/storage_async/client.rb
@@ -205,8 +205,7 @@ module ThreeScale
           host = uri.host || DEFAULT_HOST
           port = uri.port || DEFAULT_PORT
 
-          endpoint = Async::IO::Endpoint.tcp(host, port)
-          Async::Redis::Client.new(endpoint, limit: opts[:max_connections])
+          AsyncRedisClientWrapper.connect(uri, limit: opts[:max_connections])
         end
 
         def init_sentinels_client(opts)
@@ -217,6 +216,73 @@ module ThreeScale
           Async::Redis::SentinelsClient.new(name, opts[:sentinels], role)
         end
       end
+
+      class AsyncRedisClientWrapper
+        class << self
+          # @param url [String] Redis URL connection string
+          # @param ssl_params [Hash] passed to OpenSSL::SSL::SSLContext
+          # @param options [Hash] passed to Async::Redis::Client.new
+          # @return [Async::Redis::Client]
+          def call(url = 'redis://localhost:6379', ssl_params: nil, **options)
+            uri = URI(url)
+
+            endpoint = prepare_endpoint(uri, ssl_params)
+
+            credentials = []
+            credentials.push(uri.user) if uri.user && !uri.user.empty?
+            credentials.push(uri.password) if uri.password && !uri.password.empty?
+
+            db = uri.path[1..-1].to_i if uri.path
+
+            protocol = AsyncRedisProtocolWrapper.new(db: db, credentials: credentials)
+
[...]

PR #350, opened 2023-09-28
THREESCALE-8404: Add ACL and TLS support for Redis

jlledom:
This is a work in progress to allow Apisonator to connect to Redis using ACL and TLS credentials.

### Jira Issue:
[THREESCALE-8404](https://issues.redhat.com/browse/THREESCALE-8404)

### Summary of changes:
- Update gems (https://github.com/3scale/apisonator/pull/350/commits/89fe2ef53ce6bc0099646dd1c47d4e3d8e152a86)
  - To add support to TLS
  - Stop using `redis` 4.3.1 fork and use upstream `redis` 5.0.7
- Adapt code to the new boolifyed Redis API (https://github.com/3scale/apisonator/pull/350/commits/bde5ea15fee380ce701ab9d0742bdaa269092cbf)
  - Some methods like `:sadd` returned a boolean in `redis` 4.3.1, but now they return what they get from the Redis server.
  - `redis` 5.0.7 includes boolified versions of those methods, like `:sadd?`, which behave like the old `:sadd`
  - Also, now the redis client only accepts string as parameters
- Adapt the code to the new pipelined Redis API (https://github.com/3scale/apisonator/pull/350/commits/06776385f872b008d1982d3598b79156a7a67150)
  - `redis` 4.6.0 introduced some changes in the `pipelined` method, (Check [Changelog](https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#460)).
  - Now it doesn't run Redis commands from the original client when under a pipelined scope. Commands can only be sent from the pipelined context.
- Update the async client interface to match the new redis API (https://github.com/3scale/apisonator/pull/350/commits/08e5a3b47761d47c0d7ea588ce2d310ab6029753)
  - We want the business logic code to be equally valid in both sync and async modes. But `redis`, and `async-redis` gems provide different interfaces. We have an async client that reproduces the `redis` interface and translates it to `async-redis` calls.
  - Our async client interface has been updated to include calls to new boolyfied methods and use the correct pipelined context.
  - **EDIT**: I updated the pipeline logic and thread-isolated the `async-redis` client (https://github.com/3scale/apisonator/pull/350/commits/c94bd275d67261aff1cd27ed42bf12e3a3bff7ef)
- Fix tests (https://github.com/3scale/apisonator/pull/350/commits/be9c1e6bc780c37d78fb85baf25eeceee6376b19)
  - Updated the tests to adapt to the changes
- Add new accepted params and environment variables (https://github.com/3scale/apisonator/pull/350/commits/b99e3075ca98f84c1aa0b68b556e27f45c688ccc)
  - Add parameters to configure SSL Context
  - Provide environment variables to set those params
  - Update documentation
- **EDIT**: Add support for ACL and TLS in async mode (https://github.com/3scale/apisonator/pull/350/commits/8d8c4af5753c7bea0779143f616afe122a6d6e69)
  - Instead of the `redis` gem, the async mode uses the `async-redis` gem. `async-redis` hasn't been updated, but I modified our client to add support for TLS and ACL.

Connecting to Redis using TLS and ACL is done. Running Apisonator in sync mode works fine and all tests pass in sync and async mode.

I'm blocked due to a deadlock in the async mode which couldn't fix yet. To reproduce it, just run the tests with `make test`. The test failing is [spec/integration/worker_async_spec.rb](https://github.com/3scale/apisonator/blob/b99e3075ca98f84c1aa0b68b556e27f45c688ccc/spec/integration/worker_async_spec.rb)

### Pending tasks:
- [x] Fix the async deadlock.
- [x] Implement TLS in the async mode (check [docs](https://github.com/socketry/async-redis#connecting-to-redis-ssl-endpoint))
- [x] Update `test/unit/storage_sync_test.rb` to add tests for TLS connections.
- [x] Ensure all tests pass for sync and async modes.
- [x] Actually set the ENV variables and check it works, so far I only tried editing the `.3scale_backend.conf` file.
- [x] Add a new parameter and env. variable to set the `ca_path` ssl param
- [x] Investigate the consequences of having stopped using a `redis` and `resque` forks and started using the upstream gems.
- [x] Fix `ERR AUTH` error

About the forks mentioned above, the problem is we were using a forked version of the `redis` gem before: https://github.com/3scale/redis-rb

That fork included 3 PRs which are not upstream:

- https://github.com/3scale/redis-rb/pull/1
- https://github.com/3scale/redis-rb/pull/2
- https://github.com/3scale/redis-rb/pull/3

Since the fork is stuck in `redis-rb 4.1.3` and we need `>= 5`. Those three PRs are lost in this PR, so we need to investigate what were they for and whether are they still required.

The same happens with `resque`. We were using a fork before: https://github.com/3scale/resque

The fork included one PR and a few commits not in upstream:

- https://github.com/3scale/resque/pull/1
- https://github.com/3scale/resque/commit/88839e71756ea9b6edfc9426a0af71e94109c135

### Configure server and client

This requires three steps:

1. Generate TLS certificates and keys for Redis.
2. Configure Redis to enable ACL and TLS, and use the generated keys.
3. Configure Apisoator.
   3.1. TLS Client
   3.2. Mutual TLS Client

#### Generate keys and certificates:

The first thing we need is a certification authority. For that, we need to create its key and certificate.

- key: `openssl genrsa -out ca-root-key.pem 4096`
- cert: `openssl req -new -x509 -days 365 -key ca-root-key.pem -out ca-root-cert.pem`

Then we'll have to create a key and certificate for the server, and sign it with our CA. Ensure setting `localhost` as `CN`, if that's the domain you're going to use to connect to Redis:

- key: `openssl genrsa -out redis.key 4096`
- cert request: `openssl req -new -key redis.key -out redis.csr`
- signed cert: `openssl x509 -req -days 365 -in redis.csr -CA ca-root-cert.pem -CAkey ca-root-key.pem -CAcreateserial -out redis.crt`

#### Configure Redis:

First, we need a Redis configuration file which enables ACL and TLS, defines the users and sets the certificate the server will use and the CA the server will trust. This is the minimal `redis.conf` for that purpose:

```
port 6379
tls-port 26379
tls-cert-file /etc/redis.crt
tls-key-file /etc/redis.key
tls-ca-cert-file /etc/ca-root-cert.pem
tls-auth-clients optional
user default off
user porta on >sup3rS3cre1! ~* &* +@all
user apisonator on >secret#Passw0rd ~* &* +@all
```

Next step is to launch a container that has access to the certificates, keys and Redis config files created above. We create a volume for each file and modify the container start command to reference the configuration file we created. This is how I configured the redis pod in my `docker-compose.yml`. But the container can be launched by other ways:

```yml
  redis:
    image: redis:6.2-alpine
    container_name: redis-compose-ssl
    ports:
      - "26379:26379"
    volumes:
      - /home/jlledom/redis.conf:/etc/redis.conf:Z
      - /home/jlledom/redis.crt:/etc/redis.crt:z
      - /home/jlledom/redis.key:/etc/redis.key:z
      - /home/jlledom/ca-root-cert.pem:/etc/ca-root-cert.pem:z
    command: [redis-server, /etc/redis.conf]
```

The `:z` at the end of the volume definitions is required when using Selinux, for instance in Fedora.

#### Configure Apisonator:

Apisonator can be configured as a regular TLS client, when only the server needs a certificate, or a Mutual TLS Client, where both client and server must provide a certificate.

##### As TLS Client:

What I did is create a `~/.3scale-backend.conf`, based on [openshift/3scale_backend.conf](https://github.com/jlledom/apisonator/blob/962563b2429e9309f5061b2f29c8414e48dbdfd8/openshift/3scale_backend.conf) and set these properties:

```ruby
config.queues.master_name = "rediss://localhost:26379/5"
config.queues.username = "apisonator"
config.queues.password = "secret#Passw0rd"
config.queues.ssl_params = {
  ca_file: "/home/jlledom/ca-root-cert.pem"
}
config.redis.proxy = "rediss://localhost:26379/6"
config.redis.username = "apisonator"
config.redis.password = "secret#Passw0rd"
config.redis.ssl_params = {
  ca_file: "/home/jlledom/ca-root-cert.pem"
}
```

Note how the `ssl_params` sections are not always required. We are using them here because the server is using a certificate signed by an unknown authority we've just created, so we need to explicitly tell Porta to trust that authority. If the server were using a certificate signed by any of the well-known CAs, the `ssl_params` sections could be omitted.

Another possible situation is when the server is using a self-signed certificate. When this happens, there's no trusted CA to add to `ssl_params`, so the only way to go is to skip  certificate validation. Unfortunately we don't support this yet because the `redis-client` gem still haven't added support for verify modes (see: https://github.com/redis-rb/redis-client/issues/133) so we can't set `SSL_VERIFY_NONE` to the client. Anyway this is only useful for development purposes, we'll have to use our own CA for development for the moment.

##### As Mutual TLS Client:

Generate a key and a signed certificate for Apisonator:

- key: `openssl genrsa -out apisonator.key 4096`
- cert request: `openssl req -new -key apisonator.key -out apisonator.csr`
- signed cert: `openssl x509 -req -days 365 -in apisonator.csr -CA ca-root-cert.pem -CAkey ca-root-key.pem -CAcreateserial -out apisonator.crt`

Then update the config file to set the next properties:

```ruby
config.queues.master_name = "rediss://localhost:26379/5"
config.queues.username = "apisonator"
config.queues.password = "secret#Passw0rd"
config.queues.ssl_params = {
  ca_file: "/home/jlledom/ca-root-cert.pem",
  cert: "/home/jlledom/apisonator.crt",
  key: "/home/jlledom/apisonator.key"
}
config.redis.proxy = "rediss://localhost:26379/6"
config.redis.username = "apisonator"
config.redis.password = "secret#Passw0rd"
config.redis.ssl_params = {
  ca_file: "/home/jlledom/ca-root-cert.pem",
  cert: "/home/jlledom/apisonator.crt",
  key: "/home/jlledom/apisonator.key"
}
```

As mentioned above, the `ca_file` fields are only required when the server isn't using a certificate signed by one of the well-known CAs.

It could be also configured using the new ENV variables (see [docs/configuration.md](https://github.com/3scale/apisonator/compare/master...jlledom:THREESCALE-8404-redis-acl-tls?expand=1#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180))


lib/3scale/backend/configuration.rb
```diff
@@ -38,11 +38,10 @@ def parse_int(value, default)
...
 
       # Add configuration sections
-      config.add_section(:queues, :master_name, :sentinels, :role,
+      config.add_section(:queues, :master_name, :username, :password, :ssl_params, :sentinels, :role,
                          :connect_timeout, :read_timeout, :write_timeout, :max_connections)
-      config.add_section(:redis, :url, :proxy, :sentinels, :role,
-                         :connect_timeout, :read_timeout, :write_timeout, :max_connections,
-                         :async)
+      config.add_section(:redis, :url, :proxy, :username, :password, :ssl_params, :sentinels, :role,
+                         :connect_timeout, :read_timeout, :write_timeout, :max_connections,:async)
```
mayorova, line 44:
```suggestion
                         :connect_timeout, :read_timeout, :write_timeout, :max_connections, :async)
```

docs/configuration.md
```diff
@@ -26,6 +26,41 @@ variables.
...
+###  CONFIG_REDIS_PASSWORD
+
+- Redis ACL password
+- Optional. Defaults to empty.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+###  CONFIG_REDIS_CA_FILE
+
+- Certification authority certificate Redis should trust to accept TLS connections
```
akostadinov, line 45:
```suggestion
- Certification authority to validate Redis server TLS connections with
```

btw usually software also supports `CA_DIR` option but maybe for the use case we won't need. I haven't checked how we ensure trust with HTTPS connections that we use in 3scale.

docs/configuration.md
```diff
@@ -26,6 +26,41 @@ variables.
...
+- Format: path to file as string.
+
+###  CONFIG_REDIS_CERT
+
+- User certificate to connect to Redis through TLS
+- Optional. Defaults to empty.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
+
+###  CONFIG_REDIS_KEY
```
akostadinov, line 57:
```suggestion
###  CONFIG_REDIS_PRIVATE_KEY
```

I think private key is a more familiar term to describe what we have here. i.e. TLS certificate and a private key

docs/configuration.md
```diff
@@ -80,6 +115,41 @@ sentinels.
...
+- Format: path to file as string.
+
+###  CONFIG_QUEUES_CERT
+
+- User certificate to connect to Redis through TLS
+- Optional. Defaults to empty.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
+
+###  CONFIG_QUEUES_KEY
```
akostadinov, line 146: I wonder if all of these make more sense to have standard file locations. And users would be able to mount secrets at these locations with the file contents. Because these files will still need some way to be mounted into the containers anyway. So it would be double the work. Not sure, just a thought.

docs/configuration.md
```diff
@@ -80,6 +115,41 @@ sentinels.
 - Applies to: listener, worker, cron.
 - Format: string.
 
+###  CONFIG_QUEUES_USERNAME
```
akostadinov, line 118: This I guess is for the `resque` background jobs. So I suggest naming it `CONFIG_RESQUE_USERNAME` or `CONFIG_JOBS_USERNAME`. This `QUEUES` word is a little ambiguous.

btw how nice that we have yet another background jobs implementation to care about...

docs/configuration.md
jlledom: Yes, that's also supported by `redis-client`, I might implement it in the future... maybe...

akostadinov: The more important point of my comment is to figure out how the feature is to be used with the operator on-prem and make sure it suits :)

It's not presently clear to me, how are users expected to supplement this.

docs/configuration.md
jlledom: I think it's good to have the variables, since they can be very useful for development: the standard file locations you mention would probably be out of the project folder. What we could do is adding a default in the case they are not provided.

docs/configuration.md
jlledom: The `CONFIG_QUEUES_*` prefix was already there before, I prefer to not touch it.

docs/configuration.md
jlledom: Yes, it's good to provide a CA_DIR env var.

docs/configuration.md
akostadinov: hmm, I wondered how did you come up with this name.. I don't like it but what you say makes sense.

docs/configuration.md
akostadinov: The dir thing may or may not be practical depending on the way users are supposed to supply that to the pod.

docs/configuration.md
akostadinov: This is what I meant.

eguzki (dismissed):
review still in progress, first comments

docs/configuration.md
```diff
@@ -26,6 +26,41 @@ variables.
...
+###  CONFIG_REDIS_CERT
+
+- User certificate to connect to Redis through TLS
+- Optional. Defaults to empty.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
+
+###  CONFIG_REDIS_PRIVATE_KEY
+
+- User key to connect to Redis through TLS
```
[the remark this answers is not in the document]
akostadinov: >The path to the client SSL private key

eguzki:
For some reason the call to `Storage.instance.get(stats_key)` in https://github.com/jlledom/apisonator/blob/THREESCALE-8404-redis-acl-tls/spec/integration/worker_async_spec.rb#L63 never returns. It happens after a few calls to it. Usually after second or third call.

It needs to be fixed

Gemfile
```diff
@@ -55,21 +55,17 @@ gem 'daemons', '= 1.2.4'
 # Production gems
 gem 'rake', '~> 13.0'
 gem 'builder', '= 3.2.3'
-# Use a patched resque to allow reusing their Airbrake Failure class
-gem 'resque', git: 'https://github.com/3scale/resque', branch: '3scale'
+gem 'redis', '~> 5'
```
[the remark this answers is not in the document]
akostadinov: I was also surprised by this but it is documented as equivalent.

https://bundler.io/guides/gemfile.html#gemfiles

> ~> 0 is identical to >= 0.0 and < 1.0

Who would have thought..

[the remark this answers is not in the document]
eguzki: Then, shut up and take my money :)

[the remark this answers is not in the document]
jlledom: Since it's equivalent, I'll just use `5.0` which is more clear

lib/3scale/backend/storage_async/client.rb
```diff
@@ -46,124 +48,13 @@ def initialize(opts)
           port ||= DEFAULT_PORT
 
           endpoint = Async::IO::Endpoint.tcp(host, port)
-          @redis_async = Async::Redis::Client.new(
+          @redis_async = Concurrent::ThreadLocalVar.new{ Async::Redis::Client.new(
```
akostadinov, line 51: I'm ok with such an approach if using proper async is not trivial. But it appears to me that `async-redis` was built to handle concurrency. So mayeb gove a little more explanation why do we need a separate client per thread?

jlledom:
> For some reason the call to `Storage.instance.get(stats_key)` in https://github.com/jlledom/apisonator/blob/THREESCALE-8404-redis-acl-tls/spec/integration/worker_async_spec.rb#L63 never returns. It happens after a few calls to it. Usually after second or third call.
> 
> It needs to be fixed

This fixed it: https://github.com/3scale/apisonator/pull/350/commits/c94bd275d67261aff1cd27ed42bf12e3a3bff7ef

What do you think?

eguzki:
You are saying that multiple threads (were) are using the same async redis client. Which threads are those?

lib/3scale/backend/storage_async/client.rb
jlledom: `async-redis` relies on Fibers to implement concurrency, but concurrency doesn't mean parallelism. What `async-redis` does, AFAIK, is to switch between fibers on IO operations, e.g. when one fiber is waiting for answer from Redis. But all Fibers run in the same Thread. In Our case, we are calling `Storage.instance` from different Threads, `@redis_async` was being shared between threads and that caused an error `Fiber called across threads` which blocked the test.

@eguzki Does this make sense to you?

jlledom:
@eguzki 

> You are saying that multiple threads (were) are using the same async redis client. Which threads are those?

For the failing test, there are three threads accessing the storage in parallel:

1. The test itself: https://github.com/3scale/apisonator/blob/master/spec/integration/worker_async_spec.rb#L63
2. The Thread that runs the worker: https://github.com/3scale/apisonator/blob/master/spec/integration/worker_async_spec.rb#L52
3. The thread consuming Resque jobs: https://github.com/3scale/apisonator/blob/master/lib/3scale/backend/worker_async.rb#L90

eguzki:
> here are three threads accessing the storage in par

What changed in your PR to be broken? It makes all the sense that an async client is not shared between threads. Is it shared in (current) `master` branch? I understand that in the worker, one thread is reading the queue and another one is processing the tasks. It works with one async client shared among threads. If it is not supported, why is it working?

If the multithreading is added in the tests, then we should change the tests, not the base code of apisonator.

jlledom:
> What changed in your PR to be broken?

I don't know.

> Is it shared in (current) `master` branch?

Yes

> If it is not supported, how is it working?

I don't know, maybe I'm wrong. Ideas?

> If the multithreading is added in the tests, then we should change the tests, not the base code of apisonator.

The tests are the same as before, if they passed before, they should pass now to ensure the behavior is the same.

jlledom:
@akostadinov I created this script to prove `async_redis` doesn't allow sharing the connection between threads:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem 'concurrent-ruby', '~> 1.1.6'
  gem 'hiredis', '~> 0.6.1'
  gem 'async-redis', '~> 0.7.0'
  gem 'redis', git: 'https://github.com/3scale/redis-rb', branch: 'apisonator'
end

require 'concurrent'
require 'async/io'
require 'async/redis/client'
require 'hiredis'
require 'redis'

def async_thread_isolate
  client = Concurrent::ThreadLocalVar.new do
    endpoint = Async::IO::Endpoint.tcp('localhost', 22121)
    Async::Redis::Client.new(endpoint, limit: 5)
  end

  Async { client.value.set("random_key", "a_key") }.wait

  result = []
  5.times.map { Thread.new {
    Async { result.push client.value.get("random_key") }
  }}.each(&:join)

  puts result
end

def async_thread_share
  endpoint = Async::IO::Endpoint.tcp('localhost', 22121)
  client = Async::Redis::Client.new(endpoint, limit: 5)

  Async { client.set("random_key", "a_key") }.wait

  result = []
  5.times.map { Thread.new {
    Async { result.push client.get("random_key") }
  }}.each(&:join)

  puts result
end

def sync_thread_share
  client = Redis.new({url: 'redis:localhost:22121'})

  client.set("random_key", "a_key")

  result = []
  5.times.map { Thread.new {
    result.push client.get("random_key")
  }}.each(&:join)

  puts result
end

method = ARGV.first || "async_thread_isolate"
send(method)
```
I run it from inside the apisonator container, like this (I saved the script as `tmp/async_tests.rb` under the apisonator source folder):

```shell
jlledom@fedora apisonator:master=]$ make dev

[ruby@apisonator-dev apisonator]$ bundle exec script/services start

ruby@apisonator-dev apisonator]$ ruby tmp/async_tests.rb sync_thread_share
Fetching https://github.com/3scale/redis-rb
Fetching gem metadata from https://rubygems.org/.....
Resolving dependencies...
Using nio4r 2.6.1 (was 2.5.9)
Using protocol-redis 0.6.1
Using bundler 2.3.5
Using hiredis 0.6.3
Using redis 4.1.3 from https://github.com/3scale/redis-rb (at apisonator@7210a9d)
Using fiber-annotation 0.2.0
Using fiber-local 1.0.0
Using timers 4.3.5
Using concurrent-ruby 1.1.10 (was 1.1.6)
Using console 1.23.2 (was 1.16.2)
Using async 1.31.0
Using async-pool 0.4.0 (was 0.3.12)
Using async-io 1.38.0 (was 1.34.3)
Using async-redis 0.7.0
a_key
a_key
a_key
a_key
a_key

[ruby@apisonator-dev apisonator]$ ruby tmp/async_tests.rb async_thread_isolate
Fetching https://github.com/3scale/redis-rb
Fetching gem metadata from https://rubygems.org/.....
Resolving dependencies...
Using fiber-annotation 0.2.0
Using fiber-local 1.0.0
Using timers 4.3.5
Using bundler 2.3.5
Using redis 4.1.3 from https://github.com/3scale/redis-rb (at apisonator@7210a9d)
Using nio4r 2.6.1 (was 2.5.9)
Using protocol-redis 0.6.1
Using concurrent-ruby 1.1.10 (was 1.1.6)
Using hiredis 0.6.3
Using console 1.23.2 (was 1.16.2)
Using async 1.31.0
Using async-io 1.38.0 (was 1.34.3)
Using async-pool 0.4.0 (was 0.3.12)
Using async-redis 0.7.0
a_key
a_key
a_key
a_key
a_key

[ruby@apisonator-dev apisonator]$ ruby tmp/async_tests.rb async_thread_share
Fetching https://github.com/3scale/redis-rb
Fetching gem metadata from https://rubygems.org/.....
Resolving dependencies...
Using bundler 2.3.5
Using fiber-local 1.0.0
Using nio4r 2.6.1 (was 2.5.9)
Using timers 4.3.5
Using protocol-redis 0.6.1
Using hiredis 0.6.3
Using redis 4.1.3 from https://github.com/3scale/redis-rb (at apisonator@7210a9d)
Using concurrent-ruby 1.1.10 (was 1.1.6)
Using fiber-annotation 0.2.0
Using console 1.23.2 (was 1.16.2)
Using async 1.31.0
Using async-pool 0.4.0 (was 0.3.12)
Using async-io 1.38.0 (was 1.34.3)
Using async-redis 0.7.0
  0.7s     warn: Async::Pool::Controller [oid=0xab4] [ec=0xac8] [pid=263] [2023-11-28 16:19:03 +0000]
               | Closing resource while still in use!
               | {"resource":"#<Async::Redis::Protocol::RESP2::Connection:0x00007fd8d0072ac8>","usage":1}
  0.7s     warn: Async::Task [oid=0xadc] [ec=0xaf0] [pid=263] [2023-11-28 16:19:03 +0000]
               | Task may have ended with unhandled exception.
               |   NoMethodError: undefined method `zero?' for nil:NilClass
               |   → /home/ruby/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/async-pool-0.4.0/lib/async/pool/controller.rb:236 in `reuse'
               |     /home/ruby/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/async-pool-0.4.0/lib/async/pool/controller.rb:112 in `release'
               |     /home/ruby/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/async-pool-0.4.0/lib/async/pool/controller.rb:102 in `ensure in acquire'
               |     /home/ruby/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/async-pool-0.4.0/lib/async/pool/controller.rb:102 in `acquire'
               |     /home/ruby/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/async-redis-0.7.0/lib/async/redis/client.rb:119 in `call'
               |     /home/ruby/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/protocol-redis-0.6.1/lib/protocol/redis/methods/strings.rb:62 in `get'
               |     tmp/async_tests.rb:46 in `block (3 levels) in async_thread_share'
               |     /home/ruby/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/async-1.31.0/lib/async/task.rb:261 in `block in make_fiber'
a_key
a_key
a_key
a_key
```

`$ruby tmp/async_tests.rb async_thread_share` returns a different problem each time you run it.

I haven't found any mentions to thread-safety in the `async-redis` [repo](https://github.com/socketry/async-redis), but `async-redis` is built on top of [async](https://github.com/socketry/async), which according to this [issue](https://github.com/socketry/async/issues/175) doesn't allow sharing resources between threads.

It seems logical to me: an `Async::Redis::Client` instance will have it's own `@pool` (https://github.com/socketry/async-redis/blob/v0.8.0/lib/async/redis/client.rb#L38), which we are sharing between threads like in the issue I mentioned above.

I still don't have an answer to why does it work in production, where `Async::Redis::Client` is also shared among threads.

akostadinov:
I'm inclined to think that the reason not to fail in production is either that we don't really use threaded workers there, or load is not high enough or pure luck. I agree that `async` framework is not thread safe but with some changes to your test bed. Please bear with my long post.

At this part there is a small issue with the test bed.
```rb
  5.times.map { Thread.new {
    Async { result.push client.get("random_key") }
  }}.each(&:join)

  puts result
```

We wait for each thread but don't wait for each Async task. So Threads can die and the program may exit before all async tasks have completed. And we may (or may not) see the error message you posted above.

If we add a `#wait` call to the async task, then things always work. I created a gist so I'm able to link diffs and code lines. Here's the modification I'm talking about:
https://gist.github.com/akostadinov/d7a0b00fabb48ca74f08fb54090993e7/revisions?diff=unified&w=#diff-240e2d0b41b99e3ebf81ae17215a432402bfe29406f7e71df271abee4b1b544fR48

But a concurrency of 5 is not very stressing. So I changed it a little to create 55 async tasks in 55 threads, then wait for all threads and then for all async tasks before printing the results. Like this:

```rb
  55.times.map { Thread.new {
    Async { result.push client.get("random_key") }
  }}.map(&:value).map(&:wait)

  puts result
```
See https://gist.github.com/akostadinov/d7a0b00fabb48ca74f08fb54090993e7/revisions#diff-240e2d0b41b99e3ebf81ae17215a432402bfe29406f7e71df271abee4b1b544f

And the result was quiet telling. First of all I saw in console several messages:
```
               | Task may have ended with unhandled exception.
               |   FiberError: fiber called across threads
```

And additionally the process has hanged indefinitely until I interrupted it with `Ctrl+C`. And btw if I run the same code with 5 threads/tasks, then it completes alright.

Needless to say, the isolated client per thread method worked without issue, see https://gist.github.com/akostadinov/d7a0b00fabb48ca74f08fb54090993e7/revisions#diff-240e2d0b41b99e3ebf81ae17215a432402bfe29406f7e71df271abee4b1b544f

Unfortunately I'm looking at the `async` framework for the first time but it appears to me that it is a known fact that these async tasks are not thread safe. And tasks and pools shouldn't be shared across threads. While looking at your PR I spotted [an async project issue](https://github.com/socketry/async/issues/175) where the resolution was:

> the pool created in one thread may have been memoized and used from another thread. I removed the memoization and it fixed the issue

This reads to me as this is a known design limitation. So overall the taken approach by you - create an async client per thread and avoid sneaky modifications of instance variables, are **the right (if not the only) way to go forward**. Thank you for doing a great detective job and for the nice test bed for thread safety.

An interesting question still is, why test in the `master` branch didn't fail. So I think it is worth spending some more time checking whether the updated code is fully sound. Or is it just a difference of the generated instructions? We can try increasing the concurrency of the original test in `master` and see if it's gonna fail at some point.


P.S. I'm running the script locally, just using Ruby 3.0.4, it is reproducible like that just fine at least for me.

jlledom:
@akostadinov Thanks for your research. In fact, calling `.wait` for tasks mkes all tests always pass with 5 threads, but only the `async_thread_shared` test fail with 55 threads.

About why it doesn't fail in master, I have a couple of theories.

In `worker_async_spec.rb` (the test failing) we have the worker running in a separated thread, but that doesn't represent production. In production the worker runs in its own process. That process has two threads, one to read from the queue ([code](https://github.com/3scale/apisonator/blob/master/lib/3scale/backend/worker_async.rb#L90)), another one to process the jobs ([code](https://github.com/3scale/apisonator/blob/master/lib/3scale/backend/worker_async.rb#L42-L46)). The thread processing the jobs connects to redis through the shared instance ([code](https://github.com/3scale/apisonator/blob/fc7935d0f9051d91d8277f97e1f579f8c94845b3/lib/3scale/backend/stats/aggregator.rb#L66)). But the thread reading from the queue access redis using the `JobFetcher` redis instance ([code](https://github.com/3scale/apisonator/blob/4f73a1c2aa483b52a84fa295913c805d137b0ed3/lib/3scale/backend/job_fetcher.rb#L110)) which is initialized once during the startup ([code](https://github.com/3scale/apisonator/blob/5cfa13099edbdd479122076c3733eb133efd626f/lib/3scale/backend/worker_async.rb#L101)), and it creates a new instance ([code](https://github.com/3scale/apisonator/blob/04e28f36fddde408c41afd960aa0240b0ccb87ea/lib/3scale/backend/queue_storage.rb#L12)) instead of taking the shared one. Observe the comment over `redis_client` ([code](https://github.com/3scale/apisonator/blob/5cfa13099edbdd479122076c3733eb133efd626f/lib/3scale/backend/worker_async.rb#L95-L100)). So basically there are not shared client among threads in the worker, the fetcher has it's own instance and the processor it's taking the shared one and not sharing it with anybody else because there are no more threads.

About the listener process, I'm not sure how multithreaded is it really, I guess Puma creates a thread for every request, but as proven in the test script we wrote, it can cope fine with a certain amount of parallelism without failing, maybe we never reach that workload in production. On the other hand, there's that problem about all threads sending their commands to the pipeline whenever any thread opens any pipeline anywhere, and those commands are eventually sent to redis one after another from a single thread when the pipeline closes. That would contribute to reduce the number of situations where there are multiple threads actually accessing `async_redis` at the same time.

akostadinov:
Additionally running `worker_async_spec.rb` with `n_reports = 150` instead of the original `10` show the following error in console as well deadlocks the whole process
```
10.14s     warn: Async::Pool::Controller [oid=0x7cec] [ec=0x2184] [pid=1623621] [2023-11-30 00:47:58 +0200]
               | {"resource":"#<Async::Redis::Protocol::RESP2::Connection:0x00000000035c2fd0>","usage":1}
               | Closing resource while still in use!
```

To me this shows that the original version also exhibited thread unsafety (as expected). For whatever reason it manifested somehow not as prominent. But still unacceptable.

I'd like to explore 2 areas of further development here. Validate better the new async implementation. Or possibly remove async altogether if we can observe that the sync implementation has similar performance to async.

akostadinov:
> If the multithreading is added in the tests, then we should change the tests, not the base code of apisonator.

Probably at runtime we actually avoid sharing the client between threads [1]. So the test seems inconsistent with runtime environment in that regard. I'm concerned though whether a client may leak between threads in future development or runtime configuration changes and I'm inclined to err on the side of safety. Especially while this project is new to us.

@eguzki , do you have confidence that we never access same storage client from multiple threads?

[1] https://github.com/3scale/apisonator/blob/56ad75236dba3dbfd1d30f5c48716048e2176cab/lib/3scale/backend/worker_async.rb#L95-L100

jlledom:
Today I've been investigating something I don't understand. When running apisonator locally with `CONFIG_REDIS_ASYNC=true`, I always get a `No async task available!` error when receiving the first request from porta.

According to some [issues](https://github.com/socketry/cloudflare/issues/58#issuecomment-572372549) I found, I assume all calls to `@async_redis` must be done under and `Async { ... }` block. Async-redis [docs](https://github.com/socketry/async-redis/blob/main/readme.md) seem to confirm that, as all examples are wrapping client calls under an `Async` block.

That probably the reason to this block that wraps every test:

https://github.com/3scale/apisonator/blob/72977e834c311eb592d478aa74d2f44428724a30/test/test_helper.rb#L41-L43

Or this one for specs (nice comment, btw):

https://github.com/3scale/apisonator/blob/72977e834c311eb592d478aa74d2f44428724a30/spec/spec_helper.rb#L62-L69

It all makes sense, but now I'd like to now why we are not seeing this `No async task available!` error on production... ¯\_(ツ)_/¯

I'll let you have fun with this @akostadinov

jlledom:
@slopezz thanks for your comments.

>     * AFAIK async mode has some limitations like only permitting 1 single worker  https://github.com/3scale/apisonator/blob/master/docs/async.md#limitations

Shit, I would have saved some time if I have known about this doc before 😔.

>     * However there are so many advantages having just 1 worker that makes having a single worker worth it:

I assume by "worker" you mean one process. A process can have many threads in parallel and each thread can have many concurrent fibers. Right now we stick to 1 process -> 1 Thread -> Many fibers. So I think there's room for improvement on performance, and still keeping the single process so all your points below still meet.

Anyway this is all just ideas, in this PR I'll do the less changes as possible to implement TSL.

>         * we handle up to 10,000 requests/second with 95th percentile latency under 80ms

Did you use the sync mode before? did you notice a performance boost in migrating from sync to async mode?

>         * A year ago a library update of  `async-pool` introduced a mem leak, and also made performance went down like 40%, and the only reason was a change on redis connection pool management on the library (that apparently was an innocent change), fixed I think at  [THREESCALE-7864 async-pool v0.3.12 #324](https://github.com/3scale/apisonator/pull/324)

I upgraded that gem in that PR because "Why not?", but I'm removing the upgrade, it's not needed for this PR.

lib/3scale/backend/storage_async/client.rb
```diff
@@ -41,131 +34,17 @@ def instance(reset = false)
...
-          :zremrangebyscore,
-          :zrevrange
-        ].freeze
-        private_constant :METHODS_TO_BE_CALLED_DIRECTLY
-
-        METHODS_TO_BE_CALLED_DIRECTLY.each do |method|
-          define_method(method) do |*args|
-            @redis_async.call(method, *args.flatten)
-          end
+          @redis_async = Concurrent::ThreadLocalVar.new{ initialize_client(opts) }
```
eguzki, line 37: WE do not have a good reason to do this, do we? When this is running in production, it is not needed.

jlledom: I think so, we've confirmed that production will run in single-thread mode so this won't be needed. I'd like to remove it but first I need a way to fix `spec/integration/worker_async_spec.rb`. The test is wrong and blocks the  test suite if I remove the `Concurrent::ThreadLocalVar` trick.

eguzki: I also guess that there is some issue in the tests.

jlledom:
> Today I've been investigating something I don't understand. When running apisonator locally with CONFIG_REDIS_ASYNC=true, I always get a `No async task available!` error when receiving the first request from porta.

I answer to myself here. In production, the async mode is launched with [Falcon](https://github.com/socketry/falcon) instead of Puma. Falcon is ready for use with the async libraries we use in the async mode, like `async`, `async-io` and `async-redis`.

lib/3scale/backend/storage_async/client.rb
akostadinov: I think we should keep this unless it proofs to be a performance issue. This makes the class safer and it may help with future improvements of the deployment.

eguzki: IMO we should remove  that. We want to make sure the system fails if, for whatever reason, that `redis_async` instance is shared between threads.

akostadinov: Does it fail fast presently? I thought only in presence of traffic it starts complaining.

This change maybe resolves the thread safety issue. And we need to validate changes of deployment and code with good amount of traffic either way. So I don't think this would cause any additional risk.

jlledom: I pushed some changes (https://github.com/3scale/apisonator/pull/350/commits/8c281104a65f4cdf3115e4d3ed67355cf0c9a90a) to remove this. While `Concurrent::ThreadLocalVar` has been very useful to understand what was happening, right now it's just unused temporary code. I prefer to get rid of it.

Also, I updated (https://github.com/3scale/apisonator/pull/350/commits/cdc25bb790695ea7e6f0aa03fd76282f161fc19f) the test to be sure it doesn't share the Redis client between threads. I have to say I'm still unsure about this test, specially because in the past I made similar changes to it and it failed, so I don't know why it worked today. It could be just because in the past I had less background and missed something.

akostadinov: >  it should never be multi-threaded.
 
 If it yields performance gains, I see no reason not to become multi-threaded some day. For a class that is used globally (through `::instance`) I find it a limitation not to be thread safe. Alternatively `::instance` may return a different one per thread which might be even better.
 
Either way, I'm fine with whatever Joan decided to do.
 
 >  I'm still unsure about this test
 
 did you try increasing `n_reports`? Otherwise it is a hit and miss

eguzki: > If it yields performance gains, I see no reason not to become multi-threaded some day.

That would require, for N threads, N reactors and N redis-async clients. Currently the model is 1 pod -> 1 process -> 2 threads (1 for reading queue jobs, 1 for processing jobs)

Indeed, in the future, we can explore this. But currently the scaling approach is per pod. 1 pod consuming 1 CPU at max.

jlledom: > did you try increasing n_reports? Otherwise it is a hit and miss

I didn't try, I forgot the basics! 😥. But I have tried just now and it works fine. The tests pass with n_reports = 50, 500, 5000 and 50000. With no `Concurrent::ThreadLocalVar`. Always passes.

> That would require, for N threads, N reactors and N redis-async clients.

"If some day I have the time" I'd like to explore this, I think the main problem to implement multi-threading in apisonator is the cache memory implemented by memoization, which isn't thread-safe.

eguzki: > "If some day I have the time" I'd like to explore this, I think the main problem to implement multi-threading in apisonator is the cache memory implemented by memoization, which isn't thread-safe.

For bare metal with N cpus, this would make sense. In openshift with scaling based on pods, having one pod - 1 cpu is good enough to scale based on pods. We are currently working in adding [HPA](https://docs.openshift.com/container-platform/4.14/nodes/pods/nodes-pods-autoscaling.html) to the backend workers in https://issues.redhat.com//browse/THREESCALE-10500

lib/3scale/backend/storage_async/client.rb
```diff
@@ -41,131 +33,17 @@ def instance(reset = false)
...
-          host, port = opts[:url].match(HOST_PORT_REGEX).captures if opts[:url]
-          host ||= DEFAULT_HOST
-          port ||= DEFAULT_PORT
-
-          endpoint = Async::IO::Endpoint.tcp(host, port)
-          @redis_async = Async::Redis::Client.new(
-            endpoint, limit: opts[:max_connections]
-          )
-          @building_pipeline = false
+          @redis_async = initialize_client(opts)
```
akostadinov, line 36: Actually I disagree. Either this needs to initialize a client per thread (making it thread safe), or `::instance` should return a different objects per thread (preferable), or `::instance` should fail if it is ever called from another thread, effectively "failing fast". For example memoize thread that initially called it and err in case another thread calls it ever.

akostadinov: Presently we can never be sure that that this would not be used by multiple threads, relying on deployment options without any check is unreliable and prone to failure.

jlledom:
I added some tests for TLS and ACL connections. However they can't run inside the container because the image (quay.io/3scale/apisonator-ci:20230418-02) installs Redis 5, which doesn't support TLS not ACL.

I create an [issue](https://issues.redhat.com/browse/THREESCALE-10606) for that, as I'm not sure upgrading the CI image is under the scope of this PR.

docs/configuration.md
jlledom: Done in https://github.com/3scale/apisonator/pull/350/commits/a81a4e9d2b45b49c6d8be418963de9dc2c03664e

akostadinov: will we have these variables ever empty? Just wondering how it will work. As if empty, I assume empty value may confuse us. Or not?

jlledom: I pushed some changes to ensure empty variables doesn't reach Redis: https://github.com/3scale/apisonator/pull/350/commits/c429e716213fe2ff7c9a66d0d838188f51c13ec4.

lib/3scale/backend/storage_helpers.rb
```diff
@@ -183,6 +183,12 @@ def ensure_url_param(options)
             options
           end
 
+          def cfg_compact(options)
+            compact = ->(_k,v) { v.to_s.strip.empty? }
+            options[:ssl_params].reject!(&compact) if options.key? :ssl_params
+            options.reject!(&compact)
```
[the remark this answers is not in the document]
jlledom: I applied the suggestion

lib/3scale/backend/stats/aggregators/base.rb
```diff
@@ -11,7 +11,8 @@ module Base
           # @param [Time] timestamp
           # @param [Array] keys array of {(service|application|user) => "key"}
           # @param [Symbol] cmd
-          def aggregate_values(value, timestamp, keys, cmd)
+          # # @param [Redis] client
```
mayorova, line 14:
```suggestion
          # @param [Redis] client
```

?

akostadinov:
IDK how was this implemented but I see an upstream [example protocol wrapper](https://github.com/socketry/async-redis/blob/main/examples/auth/wrapper.rb) for using TLS and logical databases that easily worked for me. Not sure how easy same can apply to sentinels though

see my lazy PoC https://github.com/3scale/apisonator/issues/135#issuecomment-2020485196

lib/3scale/backend/storage_async/client.rb
```diff
@@ -216,6 +101,52 @@ def init_sentinels_client(opts)
...
+        # Authenticated RESP2 if credentials are provided, RESP2 otherwise
+        def make_redis_protocol(opts)
+          uri = URI(opts[:url] || "")
+          credentials = [ uri.user || opts[:username], uri.password || opts[:password]]
+
+          if credentials.any?
+            AuthenticatedRESP2.new(credentials)
+          else
+            Async::Redis::Protocol::RESP2
+          end
```
akostadinov, line 114: I think it is better to have a protocol like this and always use it instead of conditionally select a protocol here.

https://github.com/socketry/async-redis/blob/f11a36655aa000fe54c82899631ead0753e6cb37/examples/auth/wrapper.rb#L52-L74

script/config/ca-root-cert.pem
akostadinov, line 1: Are these certs only for testing? maybe better create such dynamically to avoid issues when they expire or idk what. Not sure this would be a concern.

Anyway, here's a gist to create certs https://gist.github.com/nickyp/886884

lib/3scale/backend/storage_async/client.rb
jlledom: Right, I'll try to do this.

script/config/ca-root-cert.pem
jlledom: Those certificates are used from the async client test suite. Since the client is so simple, we need to implement SSL support ourselves ([here](https://github.com/jlledom/apisonator/blob/THREESCALE-8404-redis-acl-tls/lib/3scale/backend/storage_async/client.rb#L117-L149)). Since it's our code, we must add tests for it, so I added some tests to ensure the only three key algorithms supported work properly ([here](https://github.com/jlledom/apisonator/blob/048904b428e89df7ecdba27d1cfc4141ca1e671c/test/unit/storage_async_test.rb#L201-L238)).

The certificates expire at 2051. The remains of a collapsed society will fight for food and water after World War 3, and the few survivors will die anyway due to global warming and nuclear radiation, so I don't think we have to worry about certificates expiration.

Anyway, I'll explore the dynamic certs approach.

mayorova: > The certificates expire at 2051. The remains of a collapsed society will fight for food and water after World War 3, and the few survivors will die anyway due to global warming and nuclear radiation, so I don't think we have to worry about certificates expiration.

Amen :pray:

akostadinov: Ok, ok, if those would be the consequences, I'm fine to keep it like it is now.

jlledom: I made some changes: https://github.com/3scale/apisonator/pull/350/commits/4dc7cab14273a98d662c2c0f292b89d67895abae

lib/3scale/backend/storage_async/client.rb
jlledom: Done https://github.com/3scale/apisonator/pull/350/commits/a126a9b781b02f040d7743f50c3788319cc23de0

The new protocol also accepts logical dbs: https://github.com/3scale/apisonator/pull/350/commits/821f08cabf8825dd7aa853cefa86d127157d1f21

lib/3scale/backend/storage_async/client.rb
```diff
@@ -202,20 +73,61 @@ def initialize_client(opts)
...
           uri = URI(opts[:url] || '')
           name = uri.host
           role = opts[:role] || :master
+          protocol = make_redis_protocol(opts)
+
+          Async::Redis::SentinelsClientACLTLS.new(name, opts[:sentinels], role, protocol, opts)
+        end
+
+        # Authenticated RESP2 if credentials are provided, RESP2 otherwise
+        def make_redis_protocol(opts)
```
mayorova, line 91: Is this code just new custom development, or copy from upstream?

lib/3scale/backend/storage_async/methods.rb
```diff
@@ -0,0 +1,127 @@
...
+          :exists?,
+          :sismember,
+          :sadd?,
+          :srem?,
+          :zadd
+        ].freeze
+        private_constant :METHODS_TO_BOOLIFY
+
+        METHODS_TO_BOOLIFY.each do |method|
+          command = method.to_s.delete('?')
```
mayorova, line 80: What's the story behind all the methods that got an extra `?` ?

lib/3scale/backend/storage_async/methods.rb
```diff
@@ -0,0 +1,127 @@
...
+module ThreeScale
+  module Backend
+    module StorageAsync
+      module Methods
+        # Now we are going to define the methods to run redis commands
+        # following the interface of the redis-rb lib.
+        #
+        # These are the different cases:
+        #   1) Methods that can be called directly. For example SET:
+        #      @redis_async.call('SET', some_key)
```
mayorova, line 12:
```suggestion
        #      call('SET', some_key)
```
Not sure if that's correct, but assuming that there is no mention of `@redis_async` in this file, and the calls are "delegated" to `@redis_async` anyway...

lib/3scale/backend/storage_async/client.rb
jlledom: It's custom. The `async-redis` client is so simple we need to implement the most basic things like sending the credentials to the redis server when it requires authentication. We also have to handle the SSL connections ourselves and. All `private` methods in the `Client` class are there to deal with these things.

The code it's base in the examples from the repo:

https://github.com/socketry/async-redis/blob/main/examples/auth/protocol.rb
https://github.com/socketry/async-redis/tree/main/guides/getting-started

mayorova: Wow, that's sad that it doesn't even do the basic stuff...
Thanks for the info!

lib/3scale/backend/storage_async/methods.rb
jlledom: Aisonator uses the API from the `redis-rb` gem, which we use in sync mode. In order to work in both sync and async modes, our custom async mode replicates the redis-rb API. The `Methods` class implement this.

The API changed in redis-rb 5, now they addded the methods `:exists?` `:sadd?` and `:srem?` https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#500. We have to adapt our API to reflect that.

For the commands coming with an ending `?` we need to remove it to get the actual command to send to Redis.

lib/3scale/backend/storage_async/methods.rb
jlledom: You're right, I copied that comment from the old file and didn't check it.

lib/3scale/backend/service.rb
```diff
@@ -277,7 +277,7 @@ def persist_attribute(attribute, value, ignore_nils = false)
       def persist_sets
         storage.sadd storage_key_by_provider(:ids), id
         storage.sadd encode_key("services_set"), id
-        storage.sadd encode_key("provider_keys_set"), provider_key
+        storage.sadd encode_key("provider_keys_set"), provider_key unless provider_key.nil?
```
akostadinov, line 280: Why can provider key be nil? Should we conditionally add or should we raise an error?

lib/3scale/backend/job_fetcher.rb
```diff
@@ -85,6 +85,8 @@ def try_pop_from_queue(max=nil)
       def wait_pop_from_queue
         queue, encoded = @redis.blpop(*@queues, timeout: @fetch_timeout)
         [queue, [encoded]] if encoded
+      rescue RedisClient::ReadTimeoutError => _e
+        # Do nothing, we expect frequent timeouts from the sync client when the queue is empty
```
akostadinov, line 89: IIRC, when `@fetch_timeout` is reached, the method will return `nil`, not raise. A read timeout on the other hand is a low-level TCP timeout that shouldn't happen regularly.

On the other hand, we don't need to be very much bothered by a connection timeout, if we can just reconnect. It's good to be resilient in case of network glitches.

I would suggest though to log the error, can be without a trace, just dome text like:
> Que redis (ip of server if possible to get) connection timeout at #{__FILE__}:#{__LINE__}

In this way we can have an indication of network stability and see whether particular servers or locations are more problematic. This will help debugging and improving the cluster setup.

lib/3scale/backend/service.rb
jlledom: Replied at https://github.com/3scale/apisonator/pull/384#issuecomment-2109914809

lib/3scale/backend/job_fetcher.rb
jlledom: Replied at https://github.com/3scale/apisonator/pull/384#issuecomment-2109715939

jlledom:
Closing this PR. I created https://github.com/3scale/apisonator/pull/391 to make this easier to review:

PR #361, opened 2024-01-25
Async worker: Use Async tools to limit concurrency

jlledom:
The asynchronous worker was using a custom method to control the amount of concurrent tasks in the reactor. However, Async provides some tools for this. Better use them rather than inventing the wheel.

Check: https://socketry.github.io/async/guides/best-practices/index

**How to verify:**

1. Set `CONFIG_ASYNC_WORKER_MAX_CONCURRENT_JOBS` to the desired number.
2. Use [hey](https://github.com/rakyll/hey) to generate fake load
  ```
./hey_linux_amd64 -z 60s -c 60 -q 20 "http://localhost:3001/transactions/authrep.xml?service_token=TOKEN&service_id=2&usage%5Bhits%5D=1&user_key=APP_USER_KEY"
```
3. Check with `puts @semaphore.count` that CONFIG_ASYNC_WORKER_MAX_CONCURRENT_JOBS is never exceeded

### DIFF
--- a/lib/3scale/backend/worker_async.rb
+++ b/lib/3scale/backend/worker_async.rb
@@ -1,4 +1,6 @@
 require 'async'
+require 'async/barrier'
+require 'async/semaphore'
 require 'redis-namespace'
 require '3scale/backend/job_fetcher'
 
@@ -23,10 +25,11 @@ module ThreeScale
 
         @job_fetcher = options[:job_fetcher] || JobFetcher.new(redis_client: redis_client)
 
-        @max_concurrent_jobs = configuration.async_worker.max_concurrent_jobs ||
+        max_concurrent_jobs = configuration.async_worker.max_concurrent_jobs ||
             DEFAULT_MAX_CONCURRENT_JOBS
 
-        @reactor = Async::Reactor.new
+        @barrier = Async::Barrier.new
+        @semaphore = Async::Semaphore.new(max_concurrent_jobs, parent: @barrier)
       end
 
       def work
@@ -40,31 +43,6 @@ module ThreeScale
         fetch_jobs_thread = start_thread_to_fetch_jobs
 
         loop do
-          break if @shutdown
-          schedule_jobs
-          @reactor.run
-        end
-
-        fetch_jobs_thread.join
-
-        # Ensure that we do not leave any jobs in memory
-        @reactor.async { perform(@jobs.pop) } until @jobs.empty?
-        @reactor.run
-
-        Async { unregister_worker }
-      end
-
-      def shutdown
-        @job_fetcher.shutdown
-        @shutdown = true
-      end
-
-      private
-
-      def schedule_jobs
-        scheduled = 0
-
-        loop do
           # unblocks when there are new jobs or when .close() is called
           job = @jobs.pop
 
@@ -74,12 +52,26 @@ module ThreeScale
 
           break if @shutdown
 
-          @reactor.async { perform(job) }
-          scheduled += 1
-
-          break if @jobs.empty? or scheduled >= @max_concurrent_jobs
+          @semaphore.async { perform(job) }
         end
+
+        fetch_jobs_thread.join
+
+        # Ensure that we do not leave any jobs in memory
+        @semaphore.async { perform(@jobs.pop) } until @jobs.empty?
+        @barrier.wait
+
+        Async { unregister_worker }
+      ensure
+        @barrier.stop
       end
+
+      def shutdown
+        @job_fetcher.shutdown
+        @shutdown = true
+      end
+
+      private
 
       def process_one
         job = @job_fetcher.fetch

### CODE REVIEW

File: lib/3scale/backend/worker_async.rb [hunk 1]
```diff
@@ -39,17 +39,12 @@ def work
...
-          @reactor.run
-        end
+        Async { process_all }
 
         fetch_jobs_thread.join
 
         # Ensure that we do not leave any jobs in memory
-        @reactor.async { perform(@jobs.pop) } until @jobs.empty?
-        @reactor.run
+        Async { clear_queue }
```
akostadinov, line 47: What is the difference between `process_all` and `clear_queue`?

Also shouldn't we wrap the whole thing inside a `Sync` call as described in the best practices link you posted?

jlledom: `process_all` is the main loop that waits forever for new jobs in the queue. When a new job is added to the queue, `@job_fetcher` pops it and adds it to `@jobs`. When interrupted, it could happen that between [worker_async.rb:76](https://github.com/3scale/apisonator/blob/a26ef8e07f052dfb1b6792fe73ebc440bbec541c/lib/3scale/backend/worker_async.rb#L76) and the shutdown of the fetcher at [worker_async.rb:44](https://github.com/3scale/apisonator/blob/a26ef8e07f052dfb1b6792fe73ebc440bbec541c/lib/3scale/backend/worker_async.rb#L44) some new jobs are added to `@jobs`. Those are already removed from the queue in Redis, so they must be processed. That's what `clear_queue` does.

jlledom: > Also shouldn't we wrap the whole thing inside a Sync call as described in the best practices link you posted?

Are you expecting me to read the links I post?

Well, using `Sync` would alter the behavior, because AFAIK falcon creates a root-level reactor, so all calls to `Async` here add tasks to the parent reactor. However, if we use `Sync`, each call would create it's own reactor that schedules the tasks under it, isolated from other tasks.

Anyway, for this particular case, I think nobody would notice any difference.

jlledom: OK, I think you mean this: https://socketry.github.io/async/guides/best-practices/index#use-a-top-level-sync-to-denote-the-root-of-your-program

Placing a `Sync` to denote the root of the program. Where do you suggest to place it? Would you wrap all the `#work` method under it?

jlledom:
@akostadinov @mayorova I made some changes about what your comments in the call:

- https://github.com/3scale/apisonator/pull/361/commits/f6c14e51c2c4a80200c39980c2f7b6be8f3633ac: Ensure stopping all tasks in case of exception.
- https://github.com/3scale/apisonator/pull/361/commits/4b257d8378e7be90972fbe0fcf612a6e88b2d252: Wrap entry points of the two threads under a `Sync` call which creates a new reactor, so it's clear which is the root reactor and where it's defined.
- https://github.com/3scale/apisonator/pull/361/commits/5c16cca51cf3d2b2ae88fa374ba27a6787f13493: Remove all calls to `Async` from the processor. This way they run sequentally because they are all in the same task under the parent reactor. All tasks under `process_all` and `clear_queue` are added to the same parent reactor, so the main task could get executed between tasks. This is prevented by the barrier, which makes the main task to wait until all children tasks finish.

lib/3scale/backend/worker_async.rb
```diff
@@ -25,33 +27,28 @@ def initialize(options = {})
...
 
       def work
-        if one_off?
-          Async { process_one }
-          return
-        end
+        Sync do
+          if one_off?
+            process_one
+            return
```
mayorova, line 36: This line fails in tests, but also when run locally...
```
LocalJumpError: unexpected return
```

I'm not sure though what's the proper way to return from that method :thinking:

mayorova: Avoiding `return` as below apparently works, but I think there should be some other way... 
```
          if one_off?
            process_one
          else
            ...
            process_all
            ...

            unregister_worker
          end
```

jlledom: I refactored it to remove the global reactor and the barriers, and replace it all by different reactors for each step. This seems to have solved the problem.

Also, FYI @mayorova @akostadinov, I forgot to mention that all tests are actually executed twice: once in sync mode and once in async mode. See the output to find where each iteration starts:

- Sync: https://app.circleci.com/pipelines/github/3scale/apisonator/964/workflows/9dcf62e5-f121-4557-b83e-c89bba3605d0/jobs/1832/parallel-runs/0/steps/0-107?invite=true#step-107-97_41
- Async: https://app.circleci.com/pipelines/github/3scale/apisonator/964/workflows/9dcf62e5-f121-4557-b83e-c89bba3605d0/jobs/1832/parallel-runs/0/steps/0-107?invite=true#step-107-24703_42

lib/3scale/backend/worker_async.rb
```diff
@@ -1,4 +1,6 @@
 require 'async'
+require 'async/barrier'
```
mayorova, line 2: I guess now we don't need to require 'async/barrier'

jlledom: Right!. Done.

spec/spec_helpers/worker_helper.rb
```diff
@@ -4,8 +4,8 @@ def process_one_job
       # The async worker creates a reactor and all the tests are run within a
       # reactor (see the config.around(:each) call on spec_helper.rb). To avoid
       # the worker creating a nested reactor inside the one already running, we
-      # run Worker.work in its own thread.
-      Thread.new { ThreeScale::Backend::Worker.work(one_off: true) }.join
+      # run Worker.work under a new reactor
```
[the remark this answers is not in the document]
jlledom: You're right, it doesn't create a new reactor as I thought. It just waits until the given task finishes.

> So it is not a new reactor. In fact it is within the existing reactor. And that's why > it would fail in a new thread. Do I get it right

Before the change, it failed because it was creating a new thread that scheduled new tasks, under the same [global reactor](https://github.com/3scale/apisonator/blob/72977e834c311eb592d478aa74d2f44428724a30/spec/spec_helper.rb#L63) being used by the main thread that runs the test. So different threads were adding tasks to the same reactor, which switches between those tasks, therefor sharing tasks across threads.

It works now because there's no new thread. The `Sync` call ensures it waits, to keep the old behavior, when we also waited for the thread to finish (`.join`).

CC @mayorova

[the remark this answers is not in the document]
jlledom: Fixed

lib/3scale/backend/worker_async.rb
```diff
@@ -74,21 +72,23 @@ def schedule_jobs
...
+        barrier.wait
       end
 
-      def process_one
-        job = @job_fetcher.fetch
-        perform(job) if job
+      def clear_queue
+        barrier = Async::Barrier.new
+        semaphore = Async::Semaphore.new(@max_concurrent_jobs, parent: barrier)
+        semaphore.async { perform(@jobs.pop) } until @jobs.empty?
```
akostadinov, line 84: Maybe use this, lest we produce a `ThreadError: queue empty`
```suggestion
        @jobs.length.times { semaphore.async { perform(@jobs.pop) } }
```

Example:
`@jobs` Queue contains 5 jobs and 5 async `perform` blocks have been registered. `@jobs.pop` runs in the background so 3 `performs` have been completed and 2 are to be completed. But at that time `@jobs.empty?` is still false so we schedule one or more `perform` blocks. That would leave us with for example 7 blocks resitered in total with 5 jobs in total. And 2 of the blocks will raise the `ThreadError: queue empty`.

I hope this makes sense.

akostadinov: Also I think we need a test for the `clear_queue` method to make sure it cleans up all jobs from the queue as such algorithms are prone to errors. Although I'm more concerned a situation where for some reason we schedule more jobs in the `@jobs` Queue while `#clear_queue` is running. Although `fetch_jobs_thread.join` should take care of this, still a little scared from such a situation.

jlledom: > But at that time @jobs.empty? is still false so we schedule one or more perform blocks

After examining the code, I think that situation is theoretically possible.

>  Although I'm more concerned a situation where for some reason we schedule more jobs in the @jobs Queue while #clear_queue is running

We already fixed all inconsistencies we found so far that could lead to that situation: `process_all` runs under a `Sync` wrapper, and the children tasks are waited for with a barrier.

But we can't be 100% sure. If that happened then this wouldn't work:

`@jobs.length.times { semaphore.async { perform(@jobs.pop) } }`

Because we would create as much tasks as `@jobs.length` at the moment it was called, so if new jobs were added while `clear_queue` is running, those jobs would be lost.

I think the source of the problem is having `@jobs.pop` called from inside children tasks. So we can fix your both concerns by extracting the `pop` call to the parent task. I suggest something like this:

```ruby
def clear_queue
  barrier = Async::Barrier.new
  semaphore = Async::Semaphore.new(@max_concurrent_jobs, parent: barrier)
  
  until job = @jobs.pop
    semaphore.async { perform(job) }
  end
ensure
  barrier.wait
end
```

That way:

1. Tasks are only scheduled when there's a job for them
2. `@jobs.pop` will wait forever until the queue is closed, so theoretically no jobs could be lost.

WDTY?

jlledom: > Also I think we need a test for the clear_queue

That would take some time, I'll do it in the next few days.

akostadinov: Yeah, I thought about it. I though implementing an algorithm longer than one line is not the ruby way and could be deprecated in the future. But lets do it anyway. I think it has to be
```
while job = @jobs.pop
```

Other than that - LGTM.

jlledom: I pushed some changes. I added a couple of tests for `#clear_queue` which if I'm right should cover all your worries. However, this is tricky and I could have missed something.

spec/unit/worker_async_spec.rb
```diff
@@ -43,6 +44,54 @@ module Backend
...
         end
+
+        describe 'when there are pending jobs in the memory queue' do
+          subject do
+            WorkerAsync.new
+          end
+
+          let(:test_job) { instance_double('BackgroundJob') }
+          let(:n_reports) { 10 }
+          let(:queue) { Queue.new.tap { |q| n_reports.times.map { q.push test_job } } }
```
akostadinov, line 55: Just a nitpick

```suggestion
          let(:queue) { Queue.new.tap { |q| n_reports.times { q.push test_job } } }
```

spec/unit/worker_async_spec.rb
```diff
@@ -43,6 +44,54 @@ module Backend
...
+              n_new_reports.times { barrier.async { queue.push test_job } }
+              barrier.wait # We don't want to close the queue while there are still tasks pushing jobs to it
+            end
+            queue.close # Unlock the thread
+
+            # Interrupt the test if the thread is locked for more than 10 seconds
+            # We assume something went wrong
+            t_start = Time.now
+            while thread.alive?
+              if Time.now - t_start > 10
```
akostadinov, line 85: another nitpick, to use monotonic clock for such purposes

```suggestion
            t_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
            while thread.alive?
              if Process.clock_gettime(Process::CLOCK_MONOTONIC) - t_start > 10
```

spec/unit/worker_async_spec.rb
```diff
@@ -43,6 +44,54 @@ module Backend
...
+                raise 'The worker is taking too much to process the jobs'
```
akostadinov, line 86: maybe kill the thread first?

spec/unit/worker_async_spec.rb
```diff
@@ -43,6 +44,54 @@ module Backend
...
+          end
+
+          it 'clears the queue'do
+            expect(subject).to receive(:perform).exactly(n_reports).times.with(test_job)
+            queue.close
+
+            subject.send(:clear_queue)
+          end
+
+          it 'clears the queue also when new jobs are added during the execution' do
```
akostadinov, line 68: I think this one deserves a comment along the lines of:
```
  # this is not an expected real usage scenario but we test it anyway
```

Just as a courtesy to future readers (maybe ourselves) that this may not be a hard requirement (subject to additional investigation).

akostadinov (approved):

jlledom:
@akostadinov I implemented your suggestions

akostadinov (approved):
Thanks! The implementation looks awesome. I semaphore will have a beneficial performance impact.

PR #364, opened 2024-02-14
Bump json to 2.5.1

lvillen:
**What this PR does**
Bumps `json` gem to 2.5.1.

**Why we need it** 
Ruby 3.0.X has `json 2.5.1` as the [default gem](https://stdgems.org/json/), however we're currently using `2.3.1` so this PR updates the Gemfile to align with that. It will also solve our current problem with RubyMine debugger, since it will resolve the actual conflict with the default `json`.


mayorova (approved):

jlledom (approved):
This solved the json issue for me.

Issue #357, opened 2023-11-17
Review existing documentation.


eguzki:

jlledom:
Closed by https://github.com/3scale/apisonator/pull/370

Issue #356, opened 2023-11-17
Clean up outdated files


eguzki:

jlledom:
Closed by https://github.com/3scale/apisonator/pull/370

Issue #354, opened 2023-11-17
Development environment based on docker compose. Remove CI image


eguzki:
- Support only one ruby version
- Separated containers for tweemproxy and sentinels
- Separated contatiners for redis. New redis version. Maybe 6.2??

eguzki:
ideally yes. More aligned with the company tech stack. That being said, facing the real world, if you can make it happen with podman, be my guest. We tried in the past in the APIcast repo https://github.com/3scale/APIcast/pull/1353 and failed miserably.

jlledom:
I use podman for docker-compose files locally and it works fine. Not sure if that will work for people using Mac, though.

eguzki:
what is Mac?  :trollface:

jlledom:
Closed by https://github.com/3scale/apisonator/pull/370

PR #370, opened 2024-02-29
Remove CI image

jlledom:
This closes several issues under https://github.com/3scale/apisonator/issues/352

- https://github.com/3scale/apisonator/issues/354
  - Adds a podman-compose.yml file to quickly launch the Redis services locally (https://github.com/3scale/apisonator/pull/370/commits/acb047b1717bb07094a52471f6893eae6b7999af)
- https://github.com/3scale/apisonator/issues/355 
  - Allows having different `.env` files for `development` and `test` (https://github.com/3scale/apisonator/pull/370/commits/7ec4d823c6f7e53005a360396a53d54ada6015fc)
  - Updates tests so they can run out of the CI container (https://github.com/3scale/apisonator/pull/370/commits/83b518108c033b127f6ac0ae7afe3b8d982ef680)
  - Updates the CircleCI workflow to not require a CI image anymore (https://github.com/3scale/apisonator/pull/370/commits/9ab0f9856df44e890560d069ce9177120fe656e6)
  - Implement a few fixes that were making the tests fail (https://github.com/3scale/apisonator/pull/370/commits/addbbcc6f5ec7ea02ce0451637dfbd405690b2ae), (https://github.com/3scale/apisonator/pull/370/commits/be00cd9f314237e58661bf0950372ac8bceee91e) and (https://github.com/3scale/apisonator/pull/370/commits/d1fa8fc7537946c5564d7986f0e67ff339afb808).
    - I suspect tests were actually always running in sync mode.
- https://github.com/3scale/apisonator/issues/357
  - Update documentation (https://github.com/3scale/apisonator/pull/370/commits/8588fb6277cb1b4574f89d6b6ddc1bbf95d338a4)
- https://github.com/3scale/apisonator/issues/356
  - Clean up unused scripts and configs (https://github.com/3scale/apisonator/pull/370/commits/3c4291d64f0c28049639a060cab6ee5d37b68198) and simplify the two scripts we still would want to use (https://github.com/3scale/apisonator/pull/370/commits/480b76113aa3b5ab8be83bfd036c5fe5f2ba186b)


.circleci/config.yml
```diff
@@ -36,40 +15,28 @@ jobs:
...
-      #    name: Run tests on all Rubies
-      #    command: |
-      #      TEST_ALL_RUBIES=1 script/ci
+          name: Run tests
+          command: bundle exec script/test
+workflows:
+  version: 2
+  workflow:
+    jobs:
+    - build
```
mayorova, line 42: :eyes:

script/test_external
```diff
@@ -1,9 +1,3 @@
 #!/bin/bash
 
-SCRIPT_DIR=$(dirname "$(readlink -f $0)")
-. ${SCRIPT_DIR}/lib/functions
-
-start_services
-bundle_exec 3scale_backend start -p 3000 $@
-stop_services
-
+bundle exec 3scale_backend start -p 3000 $@
```
mayorova, line 3: Hm, it seems to me that this script doesn't have much value anymore... It used to start the required services, but it doesn't anymore, so maybe it can go? 
We have some instructions on how to start apisonator locally, don't we? (a genuine question, I haven't read through all the updated docs yet)

spec/unit/stats/cleaner_spec.rb
```diff
@@ -7,7 +8,7 @@ module Stats
         include TestHelpers::Sequences
 
         let(:storage) { Backend::Storage.instance }
-        let(:non_proxied_instances) { storage.send(:non_proxied_instances) }
+        let(:non_proxied_instances) { [storage] }
```
mayorova, line 11: Probably we can mass-rename this variable, something like `storage_instances`?

Or even we could:
- drop this variable
- use `storage` instead
- change the methods of the Cleaner class to wrap into an array if the first param is not an array.

But probably would be better to do it in another PR.

lib/3scale/dotenv.rb
```diff
@@ -0,0 +1,17 @@
+# Ensure the proper .env file is loaded
+#
+# Don't load any .env file for production
+# First, try to load `.env.#{ENV['RACK_ENV']}`
+# If doesn't exist, try to load .env.test
+# If doesn't exist, try to load .env
+begin
```
mayorova, line 7: Hm... I think I already posted this comment, but I can't find it anywhere :sweat_smile:  So here it goes again:

```
def env_file
  return file if File.exists?(file)

  File.exists?('.env.test') ? '.env.test' : '.env'
end

begin
  return if ENV['RACK_ENV'] == 'production'

  require 'dotenv'

  Dotenv.load env_file
end
```

This is to avoid checking the file existence and assignment for he same file name twice. Not necessary to have an extra method, but I thought it was easier.

DEVELOPMENT.md
```diff
@@ -14,85 +12,48 @@ The first thing you will need is cloning the project:
...
-need to run `make test`, and you can configure any additional environment
-variables for the test scripts like so:
-
-> `$ make DOCKER_OPTS="-e TEST_ALL_RUBIES=1" test`
-
-Another alternative to execute all tests from within the development environment
-where you should not need to manually start/stop the services before is to execute
-them from inside the development environment via this script:
+To execute all tests from within the development environment
+is to execute them via this script:
```
mayorova, line 44:
```suggestion
To execute all tests from within the development environment, run this script:
```

DEVELOPMENT.md
```diff
@@ -14,85 +12,48 @@ The first thing you will need is cloning the project:
...
 
 ### Running tests (advanced)
 
-In order to run the tests manually in the container-based development, the services
-needed to run them correctly must be started before with:
-
-> `$ script/services start`
-
-Then you can execute the following commands to execute all tests:
+TYou can execute the following commands to execute all tests:
```
mayorova, line 50:
```suggestion
You can execute the following commands to run all tests:
```

just to avoid double execute :)

DEVELOPMENT.md
```diff
@@ -158,20 +119,18 @@ Currently we need to follow this process:
...
+5. Run task release:changelog:link_prs to link the issue or PR numbers in the
    CHANGELOG.md file and check the diff makes sense.
-7. Review and commit "apisonator: release X.Y.Z".
-8. Verify the tests pass and fix any issue that crops up. If at all possible,
+6. Review and commit "apisonator: release X.Y.Z".
+7. Verify the tests pass and fix any issue that crops up. If at all possible,
    deploy the version as a pre-release in a staging environment for further
    testing and early detection of issues.
-9. If you did step 4, build the new CI image, rebuild the dev image based on it,
+8. If you did step 4, build the new CI image, rebuild the dev image based on it,
```
mayorova, line 130: Is this step still needed?

DEVELOPMENT.md
```diff
@@ -158,20 +119,18 @@ Currently we need to follow this process:
 1. Change the contents of version.rb with the new version.
 2. Run `bundle install` with all Gemfiles to update all lockfiles.
 3. Run `rake license_finder:report:xml > licenses.xml`.
-4. (maybe) If you will generate a new CI image with `make ci-build`, change the
-   configuration of `.circleci/config.yml` to point to the future image.
-5. Modify CHANGELOG.md filling up data about notable changes along with
+4. Modify CHANGELOG.md filling up data about notable changes along with
    associated issue or PR numbers where available.
-6. Run task release:changelog:link_prs to link the issue or PR numbers in the
+5. Run task release:changelog:link_prs to link the issue or PR numbers in the
```
mayorova, line 124:
```suggestion
5. Run `bundle exec rake release:changelog:link_prs` to link the issue or PR numbers in the
```

mayorova:
That's a huuuuge improvement! Thanks @jlledom ! I've left some comments.

spec/server/listener_metrics_spec.rb
```diff
@@ -170,12 +170,16 @@ module Backend
...
-          it_behaves_like 'listener with metrics', :puma
+        unless ThreeScale::Backend.configuration.redis.async
+          context 'running Puma' do
+            it_behaves_like 'listener with metrics', :puma
+          end
         end
 
-        context 'running Falcon' do
-          it_behaves_like 'listener with metrics', :falcon
+        if ThreeScale::Backend.configuration.redis.async
```
akostadinov, line 179: if/else maybe instead of two independent conditional blocks?

akostadinov: same applies further down

spec/unit/queue_storage_spec.rb
```diff
@@ -52,8 +45,12 @@ module Backend
...
 
       def is_sentinel?(connection)
-        connector = connection.instance_variable_get(:@inner)
-                              .instance_variable_get(:@client)
+        if ThreeScale::Backend.configuration.redis.async
+          connector = connection.instance_variable_get(:@redis_async)
+          return connector.instance_of?(Async::Redis::SentinelsClient)
+        end
+
+        connector = connection.instance_variable_get(:@client)
```
akostadinov, line 53: if/else would read better for me here as well

test/unit/storage_async_test.rb
```diff
@@ -19,11 +20,6 @@ def test_redis_url
     assert_connection(storage)
   end
 
-  def test_redis_unix
-    storage = StorageAsync::Client.send :new, url('unix:///tmp/redis_unix.6379.sock')
-    assert_connection(storage)
-  end
```
akostadinov, line 25: Why remove this? We don't care about local socket connection?

test/test_helpers/storage.rb
akostadinov, line 1: Curious, why remove the mocking?

.circleci/config.yml
```diff
@@ -36,40 +15,28 @@ jobs:
...
-      - run:
-          name: rbenv update
-          command: |
-            rbenv_update_env
-
-      - save_cache:
-          paths:
-            - ~/.rbenv
-          key: v3-rbenv-dependencies-20230418-{{ checksum ".ruby-version" }}
+          - bundle-{{ checksum "Gemfile.lock" }}
```
akostadinov, line 18:
```suggestion
          - v1-gemfile-lock-{{ checksum "Gemfile.lock" }}
```

.circleci/config.yml
```diff
@@ -36,40 +15,28 @@ jobs:
...
       - run:
           name: bundle install
           command: |
             # required for the license_finder
-            # The Gemfile.lock specifies nokogiri (1.14.3)
+            # The Gemfile.lock specifies nokogiri (1.16.2)
             # but bundler would install native gem instead if not forced
             bundle config set --local force_ruby_platform true
-            bundle_install_rubies
+            bundle config set --local path ./vendor/bundle
```
akostadinov, line 27: maybe better set deployment mode like in porta?

script/test
```diff
@@ -1,32 +1,22 @@
...
-    if [ -z "$CI" ]; then
-        stop_services
-    fi
-}
-
 function do_license_check()
 {
     if [ -n "$CI" ]; then
-        bundle_exec rake license_finder
+        bundle exec rake license_finder
```
akostadinov, line 6: I'm not sure we should run license_finder with bundler. It calls bundler internally for its purposes.

.circleci/config.yml
```diff
@@ -36,40 +15,28 @@ jobs:
...
-          name: Run tests on Ruby 3.0.2
-          command: |
-            TEST_RUBY_VERSION=3.0.2 script/ci
-
-      #- run:
-      #    name: Run tests on all Rubies
-      #    command: |
-      #      TEST_ALL_RUBIES=1 script/ci
+          name: Run tests
+          command: bundle exec script/test
```
akostadinov, line 37: should we split sync and async tests in separate workflows? Is it possible that now result are overwritten by each other?

akostadinov: Also should we have a separate sentinel/TLS test run? Seemed like rather involved change. Or a simple smoke test at least maybe?

.circleci/config.yml
jlledom: 🤔

mayorova: @jlledom new line missing at the end of the file :sweat_smile:

script/test_external
jlledom: Yeah, not really needed anymore. Docs mention this script, that's why I kept it, but it  only run one command so better to directly mention the command in the docs and get rid of more junk.

.circleci/config.yml
jlledom: OMG!!! How can this happen to me??

DEVELOPMENT.md
jlledom: I'll fix this... 🤕

DEVELOPMENT.md
jlledom: 👍

DEVELOPMENT.md
jlledom: 👍

DEVELOPMENT.md
jlledom: Not really. And I'm not sure what to do with the steps above mentioning the CHANGELOG

spec/server/listener_metrics_spec.rb
jlledom: Right

spec/unit/queue_storage_spec.rb
jlledom: OK

test/test_helpers/storage.rb
jlledom: Because they were forcing all tests to connect to redis at `127.0.0.1:7379` and `127.0.0.1:7380` through a twemproxy. That worked because we had the CI image and the services tests that arranged to run that particular environment. Besides, the mocks weren't really mocks, they were a layer in the middle to make tests work properly against a twemproxy, as twemproxy doesn't support some commands the tests run, like `flushdb`.

After removing the CI image and the scripts, this mock is a leftover.

.circleci/config.yml
jlledom: What's that?

script/test
jlledom: I don't know, it seems correct to me since it's a rake task, and we always run tasks with bundler

.circleci/config.yml
jlledom: In my opinion, we would need to run the test suite pointing to:

- Normal Redis
- TLS
- Sentinels
- Sntinels + TLS

And all of them in both sync and async mode.

But that looks like a lot of work. Out of this PR scope for sure.

.circleci/config.yml
jlledom: OK

jlledom:
@akostadinov @mayorova I think I addressed all your comments.

.circleci/config.yml
mayorova:
```
bundle config set --local deployment 'true'
```

https://bundler.io/guides/deploying.html#deploying-your-application

I think this suggestion is good actually.

DEVELOPMENT.md
```diff
@@ -40,14 +40,13 @@ the test suite point to another redis instance.
 
 ## Running tests
 
-To execute all tests from within the development environment
-is to execute them via this script:
+To execute all tests from within the development environment, run this command:
 
-> `$ bundle exec script/test`
+> `$ bundle exec script/test`º
```
mayorova, line 45: :eyes: `º`

jlledom: I dºn't see the problem

.circleci/config.yml
jlledom: It looks good, maybe we can remove the above lines about nokogiri. Let me try

jlledom: Done. I think we can't remove the `force_ruby_platform` though.

DEVELOPMENT.md
jlledom: Fixed

.circleci/config.yml
jlledom: It's done. Now we can invalidate the cache by modifying `CACHE_VERSION` from the CircleCI UI.

.circleci/config.yml
jlledom: I created a Jira for this: https://issues.redhat.com/browse/THREESCALE-10870

mayorova (approved):

script/test
akostadinov: weshould run license_finder standalone in fact. This rake thing didn't work well in porta. I don't remember all it was.

Since this is no change, it's fine to merge as is.

P.S. haven't looked at the rake task. If it calls `license_finder` as an external command, then it should be fine.

akostadinov (approved):

script/test
jlledom: Yes, it does:

https://github.com/3scale/apisonator/blob/281309948e882e0d87fd602f27053e5cb1999055/Rakefile#L219-L227

lib/3scale/dotenv.rb
```diff
@@ -0,0 +1,20 @@
...
+# If doesn't exist, try to load .env
+def env_file
+  file = ".env.#{ENV['RACK_ENV']}"
+  return file if File.exists?(file)
+
+  File.exists?('.env.test') ? '.env.test' : '.env'
+end
+
+begin
+  return if ENV['RACK_ENV'] == 'production'
```
mayorova, line 15: I noticed [this thread](https://github.com/3scale/apisonator/pull/373#discussion_r1516708286), but will reply here, because it's relevant to this line.

@akostadinov raised a good point, indeed if we use `RACK_ENV=staging` in staging, this part will actually fail, because it will try to require `dotenv`, but it's not available in staging.

On the other hand, if we start using `RACK_ENV=production` in staging (which I am not opposed to), then we need to add a new environment variable to set the Bugsnag release stage, because currently we are using the value from `RACK_ENV`.

jlledom: Too late!! Now it'll never be fixed!

PR #371, opened 2024-03-01
Remove CHANGELOG.md

jlledom:
It's outdated. We either commit to keep it up to date, or we get rid of it.


mayorova:
I am not sure about this...

I know I said before that I was OK with it, but now I am in doubt.
First, I think it's nice to have some history of when some changes were introduced, so you can go to the PRs and investigate more. On the other hand, given that apisonator actually has its own versioning, to me it seems that we could try to keep this changelog up-to-date.

@akostadinov what do you think?

lvillen:
I agree with @mayorova, having somewhere to track changes and PRs can be really helpful and quite annoying when you can't. I'd say we commit to keep it up to date, using you words @jlledom.

akostadinov:
Nice to have, will we have the capacity and will we remember to do? I don't have a hard opinion. It's fine to have if we are able to spend the effort.

jlledom:
Ok, then I'm closing this.

PR #373, opened 2024-03-06
Fix Bugsnag integration

jlledom:
Apisonator seems to be reporting to Bugsnag only in production, but not in development or staging. That's because nor `staging` or `development` are accepted release stages in Bugsnag for Apisonator, the only accepted ones are `production` and `preview` ([code](https://github.com/3scale/apisonator/blob/fd2f52c2dc16baffc6ff3a39bd9f9b888276af90/lib/3scale/backend/logging/external/impl/bugsnag.rb#L59)). However, we stopped using `preview` as `RACK_ENV` when we migrated SaaS to OCP, we now use `staging` instead, so it must be added to the list of accepted release stages.

This PR also introduces some changes I had to make for the TLS PR, and I think this is a good opportunity to move them to master and reduce that PR size. Those changes are:

- Remove the integration with Airbrake because it's dead code today.
- Update the Bugsnag gem.
- Remove the old custom integration between Bugsnag and Resque. The newer Bugsnag release already includes its own, so we can remove ours.


lib/3scale/backend/logging/external/impl/bugsnag.rb
```diff
@@ -56,7 +54,7 @@ def configure(api_key)
                   config.timeout = 3
                   config.logger = Backend.logger
                   config.meta_data_filters = []
-                  config.notify_release_stages = ['production', 'preview']
+                  config.notify_release_stages = ['production', 'staging', 'development']
```
mayorova, line 57: This is deprecated, the new alternative is `enabled_release_stages`
https://rubydoc.info/gems/bugsnag/Bugsnag/Configuration#notify_release_stages-instance_method

jlledom: Fixed!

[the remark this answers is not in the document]
mayorova: Also, I'd probably suggest something like 

```
config.enabled_release_stages = ENV['BUGSNAG_ENABLED_STAGES'] || %w[production staging]
```

And then in development we can set `BUGSNAG_ENABLED_STAGES=development` if we really need to, and avoid spamming Bugsnag from development environment when we've got some WIP that throws exceptions all the time :grimacing: 

This would also allow us to enable Bugsnag in test environment (I think we use `RACK_ENV=test`, right?)

mayorova: OK, but while you were fixing this, I left another comment :sweat_smile:  https://github.com/3scale/apisonator/pull/373#discussion_r1514314333

WDYT?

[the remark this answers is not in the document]
jlledom: I'm not sure about this. I don't think there's much value on enabling Bugsnag reporting from the test environment, so we would only use `BUGSNAG_ENABLED_STAGES` for `development`. So we would have an env variable to enable or disable bugsnag reporting, but we already have two variables for that: `CONFIG_HOPTOAD_SERVICE` and `CONFIG_HOPTOAD_API_KEY`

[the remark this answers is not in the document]
mayorova: Oh, OK... Thanks to your comment I've just figured out why setting `BUGSNAG_API_KEY` was not working for me :smiling_face_with_tear: To be honest, that's very confusing... we probably need to get rid of this over complex setup and just use code for BugSnag. But... in another PR :sweat_smile:

mayorova:
Let's remove airbrake mention from [here](https://github.com/jlledom/apisonator/blob/fix-bugsnag/docs/configuration.md?plain=1#L349).

Also, can we get rid of this patched resque, I wonder? :thinking: https://github.com/jlledom/apisonator/blob/fix-bugsnag/Gemfile#L59-L60

We can keep the same version to avoid any changes in behavior (except the airbrake integration) for now, but this way we'll be able to upgrade it later.

lib/3scale/backend/logging/external/impl/bugsnag.rb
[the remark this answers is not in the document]
jlledom: Yeah, this come from before when we had Airbrake and Bugsnag, but now we only have Bugsnag we can remove `CONFIG_HOPTOAD_SERVICE` and `CONFIG_HOPTOAD_API_KEY` can be renamed to `CONFIG_BUGSNAG_API_KEY` or something similar. Maybe I can do it in this very PR. Let me check.

jlledom:
> Let's remove airbrake mention from [here](https://github.com/jlledom/apisonator/blob/fix-bugsnag/docs/configuration.md?plain=1#L349).
> 
> Also, can we get rid of this patched resque, I wonder? 🤔 https://github.com/jlledom/apisonator/blob/fix-bugsnag/Gemfile#L59-L60
> 
> We can keep the same version to avoid any changes in behavior (except the airbrake integration) for now, but this way we'll be able to upgrade it later.

That makes sense, I'll work on this. We are already removing that fork (and upgrading the resque gem) in the TLS branch anyway

lib/3scale/backend/logging/external/impl/bugsnag.rb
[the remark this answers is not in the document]
jlledom: After taking a look, I think it's better to not rename the variables, because it's a breaking change and it doesn't fix anything or adds a new feature. Not worth it.

About getting rid of this complex setup. The problem is we have a complex system to support several error monitoring services. Though we only support Bugsnag. So IMO we should decide if we want to keep this system just in case we need to support another service in the future or we go with Bugsnag and simplify this a lot. I would just keep the current system, so we avoid breaking changes and have the option to add new services in the future, for instance if a client requests it.

jlledom:
> Let's remove airbrake mention from [here](https://github.com/jlledom/apisonator/blob/fix-bugsnag/docs/configuration.md?plain=1#L349).
> 
> Also, can we get rid of this patched resque, I wonder? 🤔 https://github.com/jlledom/apisonator/blob/fix-bugsnag/Gemfile#L59-L60
> 

Done

mayorova (approved):
Nice!

lib/3scale/backend/configuration.rb
```diff
@@ -2,7 +2,7 @@
 require '3scale/backend/environment'
 require '3scale/backend/configurable'
 require '3scale/backend/errors'
-require 'dotenv/load' if ENV['RACK_ENV'] == 'development'
+require 'dotenv/load' if ENV['RACK_ENV'] != 'production'
```
akostadinov, line 5: so we want it in `test`? Just a question in case that was missed.

lib/3scale/backend/logging/external/impl/bugsnag.rb
```diff
@@ -56,7 +54,7 @@ def configure(api_key)
                   config.timeout = 3
                   config.logger = Backend.logger
                   config.meta_data_filters = []
-                  config.notify_release_stages = ['production', 'preview']
+                  config.enabled_release_stages = %w[production staging development]
```
akostadinov, line 57: actually we should get rid of `staging` environment and use `production` in staging. Like we do with porta. Otherwise we risk running a different thing in production and in staging. This separatio of environments only made sense in the old legacy deployment based on I forgot what it already was and don't want to remember.

lib/3scale/tasks/helpers/environment.rb
```diff
@@ -7,7 +7,7 @@ module Environment
         module_function
 
         def testable?
-          !%w(preview production).include?(ENV['RACK_ENV'])
+          !%w(staging production).include?(ENV['RACK_ENV'])
```
akostadinov, line 10: I think better leave it 
```suggestion
          !%w(production).include?(ENV['RACK_ENV'])
```

This check actually doesn't make sense to me. What's the point to begin with? If we run tests, then we want tests to run, no?

Can be for another PR though.

openshift/.env.test
```diff
@@ -1,4 +1,4 @@
-RACK_ENV=preview
+RACK_ENV=staging
```
akostadinov, line 1: According to above check, `staging` would not be testable, no? I mean the file is `.env.test`..

akostadinov (approved):
Added some comments and requests for changes. But since this is a net improvement, better improve further in new PRs. Will be easier for you and easier to review.

Thank you!

lib/3scale/backend/configuration.rb
jlledom: That's fixed in https://github.com/3scale/apisonator/pull/370

lib/3scale/backend/logging/external/impl/bugsnag.rb
jlledom: If ops starts using `production` for staging, then we could remove all mentions to staging along the project.

lib/3scale/backend/logging/external/impl/bugsnag.rb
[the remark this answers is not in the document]
akostadinov: If it ain't broken, don't fix it..

lib/3scale/backend/logging/external/impl/bugsnag.rb
akostadinov: we need to request that I guess

lib/3scale/tasks/helpers/environment.rb
jlledom: This makes little or no sense indeed. It's used from the Rake file:

https://github.com/3scale/apisonator/blob/281309948e882e0d87fd602f27053e5cb1999055/Rakefile#L17

To detect when we are launching the tests. I don't see the logic behind it, but if any it should be like this:

```ruby
ENV['RACK_ENV'] == 'test'
```

openshift/.env.test
jlledom: I don't know what is this `.env.test` for but it's not used for launching tests, that's sure, because this file was there before we added support for dotenv. Must be some list of vars loaded when building the image. Since we replaced `preview` by `staging` in all places, the expected behavior should be the same as before.

akostadinov: now hopefully it won't start to be used... Will it be automaticlaly loaded now?

akostadinov: Because we add something that was not used and then we will be thinking that it is actually needed. If not needed, better get rid of it.

jlledom: I'm not sure whether it's used or not, but I think the changes in this PR doesn't affect this behavior. It only affects container image build.

akostadinov: But we also added the dotenv (for whatever unknown reason), and probably now we have files being loaded :man_facepalming:

jlledom: No because it only loads `.env.test` in the source root, it ignores this one.

PR #375, opened 2024-03-11
Fix dotenv on staging

jlledom:
We can't deploy to staging due to a bug on how we load dotenv files by priority.

This PR fixes that. Also, it simplifies the code by never trying to load the `.env` file, since it would never be loaded anyway, because `.env.test` has more priority and it's always present because it's in the repo.

So this is how it works now:

- It doesn't use dotenv at all in staging or production.
- It tries to load `.env.{RACK_ENV}`.
- If that fails, it loads `.env.test`.

### DIFF
--- a/lib/3scale/dotenv.rb
+++ b/lib/3scale/dotenv.rb
@@ -1,6 +1,6 @@
 # Ensure the proper .env file is loaded
 #
-# Don't load any .env file for production
+# Don't load any .env file for production or staging
 # First, try to load `.env.#{ENV['RACK_ENV']}`
 # If doesn't exist, try to load .env.test
 # If doesn't exist, try to load .env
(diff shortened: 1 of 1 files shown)


josemigallas (approved):

lvillen (approved):

lib/3scale/dotenv.rb
```diff
@@ -1,18 +1,17 @@
...
+# Don't load any .env file for production or staging
 # First, try to load `.env.#{ENV['RACK_ENV']}`
 # If doesn't exist, try to load .env.test
-# If doesn't exist, try to load .env
 def env_file
   file = ".env.#{ENV['RACK_ENV']}"
   return file if File.exists?(file)
 
-  File.exists?('.env.test') ? '.env.test' : '.env'
+  '.env.test'
```
mayorova, line 10: then maybe remove this line and replace the previous one with 
```
File.exists?(file) ? file : '.env.test'
```

?

jlledom: Done at https://github.com/3scale/apisonator/pull/375/commits/705a71893cf9ba2d862b1d6fbfc58913f6fad807

mayorova (approved):

PR #376, opened 2024-03-13
Fix memory leak on the async worker

jlledom:
According to some tests ran by the SRE team, the last alpha contains a memory leak. Check this comment: https://github.com/3scale/3scale-saas/issues/754#issuecomment-1984490467

After an investigation, I found the problem was produced by this commit: https://github.com/3scale/apisonator/commit/1d83cbd4728b5f4bbce0bdb1a727ecd7ddbe7437

And in particular, the leak is caused by the use of an `Async::Barrier` to force the `process_all` method to wait until all jobs are finished.

https://github.com/3scale/apisonator/blob/35731f9ba08e9cfb6d78ba14631fc851a66ff457/lib/3scale/backend/worker_async.rb#L61-L79

The `Async::Barrier` implementation is pretty simple, it's just an array `@tasks` of tasks that is being filled with a new task at every call to `async`. However, the tasks in this array aren't removed from it when the task finishes, instead, all tasks are removed at once when calling `wait`. See the code here: https://github.com/socketry/async/blob/v1.31.0/lib/async/barrier.rb

Since we only call `wait` once on the shutdown process, every one of the thousands of requests apisonator receives creates a task which is added to the barrier array and never removed, this is what produces the memory leak.

If you check the old worker code ([here](https://github.com/3scale/apisonator/blob/598e7225ac6d0af61dfb3b159961d7a0f420cd3e/lib/3scale/backend/worker_async.rb)) you can see how they tried to run the tasks in batches of `@max_concurrent_jobs` jobs. They called `schedule_jobs` to create `@max_concurrent_jobs` tasks and then called `@reactor.run` to run all of them. In my opinion, that was not working, because `@reactor.async { perform(job) }` actually runs the task, and `@reactor.run` does nothing. But I think the original idea of running tasks in batches of `@max_concurrent_jobs` tasks can work.

This PR adds one line that ensures that when there are more than `@max_concurrent_jobs` tasks in the barrier, no more tasks are scheduled and the current ones are waited to finish, thus removed from the barrier and fixing the memory leak.

This is how it worked before compared to now:

Before:
1. Up to 20 tasks are scheduled
2. One of them finishes
3. Another one takes the slot
4. Will always be 20 tasks being run concurrently

Now:
1. Up to 20 tasks are scheduled
2. Wait for all of them to finish
3. Accept 20 new tasks

In my tests this doesn't have any impact in performance, on the opposite, it seems to be faster now it doesn't have to work with that huge array.

This batching code is one possible approach to fix the leak, but I'm open to suggestions. Ideas?

### DIFF
--- a/lib/3scale/backend/worker_async.rb
+++ b/lib/3scale/backend/worker_async.rb
@@ -79,7 +79,7 @@ module ThreeScale
           barrier.wait if barrier.size > semaphore.limit
         end
       ensure
-        barrier.stop
+        barrier.wait
       end
 
       def clear_queue
@@ -91,7 +91,6 @@ module ThreeScale
         end
       ensure
         barrier.wait
-        barrier.stop
       end
 
       def start_thread_to_fetch_jobs
--- a/lib/3scale/backend/worker_async.rb
+++ b/lib/3scale/backend/worker_async.rb
@@ -73,6 +73,9 @@ module ThreeScale
           break if @shutdown
 
           semaphore.async { perform(job) }
+
+          # Wait for the last `@max_concurrent_jobs` tasks to finish and clean the barrier.
+          # Otherwise it produces a memory leak. More info: https://github.com/3scale/apisonator/pull/376
           barrier.wait if barrier.size > semaphore.limit
         end
       ensure
--- a/lib/3scale/backend/worker_async.rb
+++ b/lib/3scale/backend/worker_async.rb
@@ -73,9 +73,10 @@ module ThreeScale
           break if @shutdown
 
           semaphore.async { perform(job) }
+          barrier.wait if barrier.size > semaphore.limit
         end
       ensure
-        barrier.wait
+        barrier.stop
       end
 
       def clear_queue
@@ -87,6 +88,7 @@ module ThreeScale
         end
       ensure
         barrier.wait
+        barrier.stop
       end
 
       def start_thread_to_fetch_jobs
--- a/lib/3scale/backend/worker_async.rb
+++ b/lib/3scale/backend/worker_async.rb
@@ -74,8 +74,9 @@ module ThreeScale
 
           semaphore.async { perform(job) }
 
-          # Wait for the last `@max_concurrent_jobs` tasks to finish and clean the barrier.
-          # Otherwise it produces a memory leak. More info: https://github.com/3scale/apisonator/pull/376
+          # Clean-up tasks inside barrier regularly, otherwise they accumulate throughout the worker lifetime
+          # and never GCed, eventually exhausting the whole available memory. Moreover the array keeping
+          # track of tasks in the barrier grows indefinitely too occupying memory and reducing performance.
           barrier.wait if barrier.size > semaphore.limit
         end
       ensure


lib/3scale/backend/worker_async.rb
```diff
@@ -73,9 +73,10 @@ def process_all
           break if @shutdown
 
           semaphore.async { perform(job) }
+          barrier.wait if barrier.size > semaphore.limit
         end
       ensure
-        barrier.wait
+        barrier.stop
```
akostadinov, line 79: I think we previously discussed that we don't want `stop`because it cancels tasks and that would lead to missed tasks as they are already out of Redis. No?

https://www.rubydoc.info/gems/async/2.2.1/Async/Barrier#stop-instance_method

akostadinov:
Great discovery work! It's worth filing a feature request in Async Ruby to allow clearing finished tasks from `Barrier` without waiting.

But as a fix here, that should do the job.

I'm concerned about introducing the `stop` calls though. Also would be nice to add a comment why we do the extra waiting. It's otherwise non-obvious.

lib/3scale/backend/worker_async.rb
jlledom: Right, I removed the calls to `stop` (https://github.com/3scale/apisonator/pull/376/commits/82de3b1ed07ec26e02e02f24abb921415d63a89a).

lib/3scale/backend/worker_async.rb
```diff
@@ -73,6 +73,10 @@ def process_all
           break if @shutdown
 
           semaphore.async { perform(job) }
+
+          # Wait for the last `@max_concurrent_jobs` tasks to finish and clean the barrier.
+          # Otherwise it produces a memory leak. More info: https://github.com/3scale/apisonator/pull/376
```
akostadinov, line 78:
```suggestion
          # Clean-up tasks inside barrier regularly, otherwise they accumulate throughout the worker lifetime
          # and never GCed, eventually exhausting the whole available memory. Moreover the array keeping
          # track of tasks in the barrier grows indefinitely too occupying memory and reducing performance.
```

Something along these lines I think summarizes it. Or no?

lib/3scale/backend/worker_async.rb
```diff
@@ -73,6 +73,10 @@ def process_all
...
+          barrier.wait if barrier.size > semaphore.limit
```
akostadinov, line 79: I assume we can use `semaphore.limit * 2` or just some other bigger number if we are concerned about reuced throughput. e.g. we can consider `1000` entries a small enough number to keep in memory while not disturbing the queue parallelization as much as waiting for each new batch. The value doesn't seem to really depend on the semaphore size as long as it is a bigger number.

If we want to squeeze out every last bit of performanec, we could also keep a counter in the loop and wait every `X` iterations of the loop. But that's over the top I think..

I mean keeping as is is fine for me. Just saying that there are other options.

jlledom: Yes, the batch size doesn't need to be related to the semaphore size at all. I took the semaphore size to reduce complexity and because tasks will start waiting when the semaphore is full anyway. I think I'm going to make some performance tests with different batch sizes and check if I notice any difference. If there are no differences in performance, I'll keep the batch set to `@max_concurrent_jobs` for simplicity.

akostadinov (approved):

lib/3scale/backend/worker_async.rb
jlledom: I made some tests and I haven't observed big differences in performance between different batch sizes. However, I found that sometimes the worker just stops processing jobs even when there are new jobs being added by the listener. I'll have to investigate this.

akostadinov: hmm, could it be some kind of a dead lock, but why? Do you see it regardless of batch size?

jlledom: It looks like a dead lock. I saw it with different batch sizes

akostadinov: What `hey` parameters did you use? I can't reproduce locally with `hey -z 30m -c 60`.

akostadinov:
I'm running worker locally for hours and I see no lock-up. I'll merge this because it is a net improvement and we need to figure out whether the lock-up is some environmental issue. Can be done in another PR. 

One possible theory about lock-up I can imagine might be caused by losing connection to Redis. For some strange reason, we pass redis client from the worker to the `JobFetcher`. But then on connection error, the `JobFetcher` itself would re-create the connection.

https://github.com/jlledom/apisonator/blob/9a4c16cb8a2d266c8229946957a7e55bb23ba6e9/lib/3scale/backend/job_fetcher.rb#L83-L86https://github.com/jlledom/apisonator/blob/9a4c16cb8a2d266c8229946957a7e55bb23ba6e9/lib/3scale/backend/job_fetcher.rb#L83-L86

This is code duplication, unnecessary complexity and a risk for refactoring.

Anyway, maybe in certain redis connection errors, the `JobEnqueuer` will loop creating new connections and failing forever.

Although, locally, when I stop redis while worker is running, I see the non-descriptive `ERROR 275551 [26/Mar/2024 20:11:04 UTC] EOFError` and then worker quits. So I'm not sure what kind of connection errors the fetcher actually handles gracefully.

Apparently it would be beneficial to get this method `Worker.logger.notify` log errors better. But if you observe lock-ups, do you observe any errors reported to console? Only exception name is visible but still something is.

PR #378, opened 2024-03-29
do all in async

akostadinov:
Use a single client and do everything within the async framework.

Do not complicate things with threads and duplicating Redis client connection.

The failing tests are redundant. Probably just have to be deleted. But wanted to first get this agreed upon and then handle the details.


lib/3scale/backend/job_fetcher.rb
```diff
@@ -11,25 +11,13 @@ class JobFetcher
       REDIS_TIMEOUT = 60
       private_constant :REDIS_TIMEOUT
 
-      DEFAULT_MAX_PENDING_JOBS = 100
-      private_constant :DEFAULT_MAX_PENDING_JOBS
-
-      DEFAULT_WAIT_BEFORE_FETCHING_MORE_JOBS = 1.0/100
-      private_constant :DEFAULT_WAIT_BEFORE_FETCHING_MORE_JOBS
-
```
akostadinov, line 19: These are handled by using `SizedQueue`.

lib/3scale/backend/job_fetcher.rb
```diff
@@ -64,35 +52,9 @@ def fetch
...
-              )
-
-             # If there is a different kind of error, it's probably a
-             # programming error. Like sending an invalid blpop command to
-             # Redis. In that case, let the worker crash.
-            end
-            job_queue << job if job
-          end
+          job = fetch
+          job_queue << job if job
```
akostadinov, line 56: leave reconnection to how it is handled by the default client. I dunno if reconnection is done there. But it would be the same as the job processors. It makes no sense if job processing fails to keep running fetcher anyway.

lib/3scale/backend/worker_async.rb
```diff
@@ -13,17 +12,17 @@ class WorkerAsync
...
       def initialize(options = {})
         trap('TERM') { shutdown }
         trap('INT')  { shutdown }
 
         @one_off = options[:one_off]
-        @jobs = Queue.new # Thread-safe queue
+        @jobs = SizedQueue.new(max_pending_jobs)
 
-        @job_fetcher = options[:job_fetcher] || JobFetcher.new(redis_client: redis_client)
+        @job_fetcher = options[:job_fetcher] || JobFetcher.new
```
akostadinov, line 25: It will inherit the default redis client like it is done when called by the sync worker. But in this case the inherited client is async.

lib/3scale/backend/worker_async.rb
```diff
@@ -32,13 +31,11 @@ def initialize(options = {})
...
-        fetch_jobs_thread = start_thread_to_fetch_jobs
-
-        Sync { process_all }
-
-        fetch_jobs_thread.join
+        Sync do
+          register_worker
+          start_to_fetch_jobs
+          process_all
+        end
```
akostadinov, line 38: We don't care too much about order of these. Also fetcher now should use the same reactor as the job processing.

lib/3scale/backend/worker_async.rb
```diff
@@ -62,22 +59,24 @@ def process_all
...
+
         loop do
           # unblocks when there are new jobs or when .close() is called
           job = @jobs.pop
 
           # If job is nil, it means that the queue is closed. No more jobs are
-          # going to be pushed, so shutdown.
-          shutdown unless job
-
-          break if @shutdown
```
akostadinov, line 73: This break here was harmful previously, because at this point `job` is present but we do not `perform` it, rather it would just be lost.

Also at this point, it makes no sense to shutdown fetcher, because if queue is closed, then fetcher would be down already.

That leave us just to quit in case queue is closed. `barrier.wait` in the `ensure` block should make sure that all jobs were `performed`.

With this flow we can also remove the `clear_queue` phase. But it should not hurt to stay for now.

lib/3scale/backend/worker_async.rb
```diff
@@ -94,23 +93,14 @@ def clear_queue
         barrier.wait
       end
 
-      def start_thread_to_fetch_jobs
-        Thread.new do
-          Sync { @job_fetcher.start(@jobs) }
-        end
+      def start_to_fetch_jobs
+        Async { @job_fetcher.start(@jobs) }
```
akostadinov, line 97: instead of a new thread, we just start a new async fiber.

lib/3scale/backend/job_fetcher.rb
jlledom: What you say sounds logic. In the case of a connection error, what the client does by default is crashing, AFAIK. Then OCP starts another container to replace the crashed one.

lib/3scale/backend/worker_async.rb
jlledom: I agree we don't care about the order of `start_to_fetch_jobs` and `process_all`. But I think `register_worker` should be finished before entering this block. Even when we don't know exactly what's for.

jlledom (commented):
Apart from moving `register_worker` to it's own `Sync` block, I'd say this looks good to me. I made a lot of tests locally. Like making tests artificially slow, crashing the worker, the listener, redis... and I never saw the block and in all cases the amount of `200 OK` answers from the listener matched the number of reported jobs. Looks pretty stable. Good job.

lib/3scale/backend/job_fetcher.rb
```diff
@@ -11,49 +11,41 @@ class JobFetcher
...
       def start(job_queue)
         loop do
           break if @shutdown
-          job = fetch
-          job_queue << job if job
+
+          jobs = fetch(wait: false, max: Worker::DEFAULT_MAX_CONCURRENT_JOBS)
+
+          # if there were no jobs, we can make a blocking call instead of reinventing waiting logic
+          jobs = fetch(max: Worker::DEFAULT_MAX_CONCURRENT_JOBS) unless jobs
```
jlledom, line 46: I don't see the point on `wait:false` and this double call to `fetch`. First we call it non-waiting, if it returns nothing because there's nothing in any queue, then wait. OK, But that's what will happen anyway if we call it with `wait:true` directly, isn't it? So the first call seems redundant to me, and the `wait` parameter unnecessary.

lib/3scale/backend/job_fetcher.rb
```diff
@@ -11,49 +11,41 @@ class JobFetcher
...
+      #   note: presently ignored in waiting mode because BLMPOP was added in Redis 7
+      #   note: also we need to update redis-rb gem so that sync client can support this parameter
+      # @return [nil, Resque::Job, Array<Resque::Job>] nil when no jobs are present in the queue,
+      #                                                 a single job when max was not specified,
+      #                                                 an array of jobs otherwise
+      def fetch(wait: true, max: nil)
+        encoded_jobs = wait ? wait_pop_from_queue : try_pop_from_queue(max)
+        return if !encoded_jobs || encoded_jobs.empty?
+
+        jobs = encoded_jobs.map { decode_job _1 }
```
jlledom, line 32: This code is wrong,because `encoded_jobs` will be a single dimension array like 

`['queue:priority', '{"class": "ReporJob", ... }']` 

when only one job is returned, but a two-dimensional array when there are more jobs:

`[['queue:priority', '{"class": "ReporJob", ... }'],['queue:priority', '{"class": "ReporJob", ... }']]`. 

So `_1` here will be `queue:priority` in the first case and `['queue:priority', '{"class": "ReporJob", ... }']` in the second.

Note how the code to decode the job was like this before:

```ruby
encoded_job[0].sub!('queue:', '')
Resque::Job.new(encoded_job[0],
                          Yajl::Parser.parse(encoded_job[1], check_utf8: false))
```
The calls to `encoded_job[0]` and `encoded_job[1]` prove the decoder expected an array (the second form), but the new decoder expects aparently a `String`, which won't work.

lib/3scale/backend/job_fetcher.rb
```diff
@@ -105,23 +59,47 @@ def shutdown
...
-          # connection errors from the rest.
-          if e.message == 'ERR Connection timed out'.freeze
-            raise RedisConnectionError.new(e.message)
-          else
-            raise e
-          end
+      # @param max [Integer] maximum number of records to return
+      #   note: we need to update redis-rb gem so that sync client can also support it
+      # @return [nil, Array<String>] array of strings or `nil` if no results
+      def try_pop_from_queue(max=nil)
```
jlledom, line 65: Same as above, I don't see the point on having this method if we are going to end up calling `wait_pop_from_queue` anyway when the queue is empty, and assuming it's not empty, `wait_pop_from_queue` will return immediately as `try_pop_from_queue` would do, so `try_pop_from_queue` seems redundant to me.

akostadinov:
> I think we should revert this last commit because it complicates the code, requires upgrading redis-rb for the sync mode and apparently will only make a difference when the client uses Redis 7+.

* doesn't require Redis 7, but on 7 we can be more efficient
* doesn't require upgrading redis-rb because sync client doesn't use any of the improvements

lib/3scale/backend/job_fetcher.rb
akostadinov: I've got it already, yes

lib/3scale/backend/job_fetcher.rb
akostadinov: `try_pop_from_queue` will return multiple results. The idea is that we fetch jobs as fast as possible with `try_..` and if queues are empty, block until something comes up. That is for efficiency. Under high load we will be fetching fast, under low load we don't needlessly perform many queries.

lib/3scale/backend/job_fetcher.rb
akostadinov: See my other comment. `lpop` fetch multiple values with one call so we can fetch much more efficient.

jlledom:
>     * doesn't require Redis 7, but on 7 we can be more efficient

OK.

>     * doesn't require upgrading redis-rb because sync client doesn't use any of the improvements

But the job fetcher is shared between the sync and async clients, how does it not use the improvements?

akostadinov:
>> doesn't require upgrading redis-rb because sync client doesn't use any of the improvements
> But the job fetcher is shared between the sync and async clients, how does it not use the improvements?

the sync worker, just calls fetch without parameters and that it, suggested implementation keeps this call the same as before

lib/3scale/backend/job_fetcher.rb
akostadinov: > I don't understand this part, under low load this will always call blpop as it did before this PR, it will be the same, right?

`blpop` will be only called if redis list is empty/missing. Under low load I expect us to end-up using `blpop` far more often, but then if load is low, doing a few extra calls shouldn't bother the database or any other component.

I believe the benefits during high load far outweigh the effect during low-load.

lib/3scale/backend/worker_async.rb
akostadinov: done

spec/integration/worker_async_spec.rb
```diff
@@ -52,7 +52,7 @@ module Backend
         end
 
         worker = Worker.new(async: true, job_fetcher: JobFetcher.new(fetch_timeout: 1))
-        worker_thread = Thread.new { worker.work }
+        worker_thread = Async { worker.work }
```
akostadinov, line 55: it turns out that in CI, using a thread can cause the deadlock you have observed locally, although I still couldn't reproduce on my machine.. just to tell you it is not only you

spec/integration/worker_async_spec.rb
```diff
@@ -8,7 +8,7 @@ module Backend
 
     DEFAULT_SERVER = '127.0.0.1:6379'.freeze
 
-    context 'when there are jobs enqueued' do
+    context 'when there are jobs enqueued', if: Backend.configuration.redis.async do
```
akostadinov, line 11: exclude this test in sync mode

spec/integration/worker_async_spec.rb
jlledom: Maybe this should be called `worker_fiber` now

Rakefile
```diff
@@ -260,6 +260,7 @@ end
 desc 'open debug console'
 task :console do
   require 'irb'
+  require '3scale/backend/job_fetcher'
```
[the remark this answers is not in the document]
akostadinov: when you open console, with htis, you can just call `fetcher = JobFetcher.new` without requiring it first. Convenience.

lib/3scale/backend/worker.rb
```diff
@@ -17,6 +17,8 @@ module Worker
       include Resque::Helpers
       include Configurable
 
+      DEFAULT_MAX_CONCURRENT_JOBS = 20
```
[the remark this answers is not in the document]
akostadinov: Integer?

lib/3scale/backend/worker_async.rb
```diff
@@ -47,8 +46,8 @@ def work
       end
 
       def shutdown
+        Worker.logger.info "Shutting down fetcher.."
```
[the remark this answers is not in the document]
akostadinov: Is this a problem? I don't know why it doesn't want to log but should we be concerned with it?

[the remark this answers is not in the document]
jlledom: I think it would require some debugging to ensure it closes correctly instead on crashing. I'm not sure if this is just a warning or it crashes.

lib/3scale/backend/worker.rb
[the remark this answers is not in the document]
akostadinov: You can call `freeze` on it but what is it supposed to be its effect?

lib/3scale/backend/worker_async.rb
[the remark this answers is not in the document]
akostadinov: Good point, although I'm not sure it would make a practical difference. I'll look into it.

lib/3scale/backend/worker.rb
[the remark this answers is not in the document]
jlledom: Apparently no effect:

```ruby
$ bundle exec pry
[1] pry(main)> CONSTANT = 10
=> 10
[2] pry(main)> CONSTANT += 1
(pry):2: warning: already initialized constant CONSTANT
(pry):1: warning: previous definition of CONSTANT was here
=> 11
[3] pry(main)> CONSTANT
=> 11
[4] pry(main)> CONSTANT.freeze
=> 11
[5] pry(main)> CONSTANT += 1
(pry):5: warning: already initialized constant CONSTANT
(pry):2: warning: previous definition of CONSTANT was here
=> 12
[6] pry(main)> CONSTANT
=> 12
[7] pry(main)> 
```
So ignore this.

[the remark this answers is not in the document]
akostadinov: integers should be immutable already, `+=` doesn't actually modify an object, rather it is a syntax sugar IIRC

spec/integration/worker_async_spec.rb
akostadinov: not actually a fiber, renamed to `worker_async_job`

akostadinov: finally `worker_async_task`

lib/3scale/backend/worker_async.rb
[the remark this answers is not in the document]
akostadinov: So here's how I verified it.

First I modified the shutdown function like:
```rb
      def shutdown
        Worker.logger.info "Shutting down fetcher.."
        @job_fetcher.shutdown
        warn "asdafsdsgsdgsdsgdgds"
      end
```

Then I started worker:
```
CONFIG_REDIS_ASYNC=true CONFIG_INTERNAL_API_USER=system_app CONFIG_INTERNAL_API_PASSWORD=password CONFIG_QUEUES_MASTER_NAME=localhost:6379/5 CONFIG_REDIS_PROXY=localhost:6379/6 bundle exec bin/3scale_backend_worker --debug
```

Then `Ctrl+C` and I saw:
```
INFO 673308 [05/Apr/2024 16:47:51 UTC] Start processing all..
^Clog writing failed. can't be called from trap context
asdafsdsgsdgsdsgdgds
```

I think this is a pretty good proof that the shutdown is properly executed, just logger refuses to work "in the trap context".

Also if runner was not properly shutdown, the process wouldn't exit that quickly but would linger forever. This I also tried by commenting out the `@job_fetcher.shutdown` call. Then started worker in the same way. All `Ctrl+C` resulted in `^Clog writing failed. can't be called from trap context` lines:
```
INFO 673557 [05/Apr/2024 16:51:46 UTC] Start processing all..
^Clog writing failed. can't be called from trap context
^Clog writing failed. can't be called from trap context
^Clog writing failed. can't be called from trap context
^Clog writing failed. can't be called from trap context
^Clog writing failed. can't be called from trap context
^Clog writing failed. can't be called from trap context
```

So the only option to stop it was `SIGKILL`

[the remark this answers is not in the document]
jlledom: Nice

lib/3scale/backend/worker_async.rb
```diff
@@ -94,23 +98,14 @@ def clear_queue
         barrier.wait
       end
 
-      def start_thread_to_fetch_jobs
-        Thread.new do
-          Sync { @job_fetcher.start(@jobs) }
-        end
+      def start_to_fetch_jobs
+        Async(transient: true) { @job_fetcher.start(@jobs) }
```
mayorova, line 102: Why is this task transient?...

mayorova:
I didn't do an extremely detailed review, and I didn't even try to understand the updates in the tests, but in general it looks good! These code updates are actually easier to understand than the previous implementation. Well done :+1: 

I just had on comment on `transient` param.

jlledom (commented):
I think there are still some tests missing, but I'll write them myself in order to preserve @akostadinov mental health.

lib/3scale/backend/worker_async.rb
jlledom: This means when all other tasks in the reactor have finished, don't wait for this task, just kill it.

https://socketry.github.io/async/guides/asynchronous-tasks/index#reactor-lifecycle

mayorova: Yeah, that's what I read in the docs, and that makes me wonder... Don't we need the job fetching to be running forever? (even if other non-transient tasks are completed)

Maybe I just don't understand how it works, or I don't know what other jobs are running in the same reactor.

If you do understand it and it looks good, then OK :)

spec/integration/worker_async_spec.rb
```diff
@@ -38,47 +81,30 @@ module Backend
...
-        without_resque_spec do
-          Timecop.freeze(current_time) do
-            n_reports.times do
-              Transactor.report(
-                  provider_key,
-                  service_id,
-                  0 => { app_id: app_id, usage: { metric_name => 1 } }
-              )
-            end
+        Timecop.freeze(current_time) do
```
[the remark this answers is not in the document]
akostadinov: I think it is related to key generation, it was like that, it works, so don't want to mess with it.

lib/3scale/backend/storage_async/client.rb
```diff
@@ -77,6 +78,7 @@ def initialize(opts)
           :ltrim,
           :mget,
           :ping,
+          :rpoplpush,
```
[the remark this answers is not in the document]
akostadinov: not used, just works, you htink we better remove?

lib/3scale/backend/worker_async.rb
akostadinov: there is one task fetching (this one) and another processing jobs. The processing loop also spawns individual tasks but lets leave that alone. If we don't have processing loop, we don't want to fetch jobs from redis as we will just lose them without anything to perform them.

PR #379, opened 2024-03-29
improve error logging

akostadinov:
Locally at least, by default exceptions are logged without any message and backtrace.

Has to be tested how it works in production and tuned in case there is duplicate logging there.


lib/3scale/backend/logging.rb
```diff
@@ -39,7 +39,11 @@ def enable_logging
       end
 
       def logger_notify_proc(logger)
-        Logging::External.notify_proc || logger.method(:error).to_proc
+        external_notify_proc = Logging::External.notify_proc
+        proc do |exception, *args, &block|
+          logger.error('Exception') { {exception: {class: exception.class, message: exception.message, backtrace: exception.backtrace[0..3]}} }
+          external_notify_proc&.call(exception, *args, &block)
+        end
```
jlledom, line 46: I would need some help to understand this. The old code tries to use the external `notify_proc` (Bugsnag, I assume). If no external logger, then it uses `Logger#error`, so it's either one or the other.

However, this new code seems to use both, first `Logger#error`, then Bugsnag if configured. That's a change in the behavior, why is this required?

On the other hand, when I place some wrong code in the worker to force an error, I see the same console output with both old and new code.

akostadinov: If you compare `notify` before and now, it is not possible to see same message.

akostadinov: For example add a `raise` statement somewhere where notify is called, in the fetcher decode for example

jlledom: I checked this better. Some comments:

1. This changes only apply when `Worker.logger.notify(e)` is called explicitly, for instance, the job fetcher catches exceptions and calls it, but errors happening in other places aren't affected. That's why I didn't see a difference, I tried adding a `nil.no_method` in the first line at `WorkerAsync#work` And that error never gets sent to `#notify`
2. When the error is not sent to `#notify`, it looks like this:
   ```
   Uncaught exception: Test error
	/home/jlledom/redhat/apisonator/lib/3scale/backend/worker_async.rb:33:in `work'
	/home/jlledom/redhat/apisonator/lib/3scale/backend/worker.rb:47:in `work'
	/home/jlledom/redhat/apisonator/bin/3scale_backend_worker:7:in `debug'
	/home/jlledom/redhat/apisonator/bin/3scale_backend_worker:36:in `<top (required)>'
	/home/jlledom/.asdf/installs/ruby/3.0.2/lib/ruby/gems/3.0.0/gems/debase-3.0.0.beta.10/lib/debase.rb:224:in `load_protect'
	/home/jlledom/.asdf/installs/ruby/3.0.2/lib/ruby/gems/3.0.0/gems/debase-3.0.0.beta.10/lib/debase.rb:224:in `debug_load'
   ```
   But this is how it looks when `#notify` is involved:
   ```
   ERROR 78607 [04/Apr/2024 15:05:53 UTC] {:exception=>{:class=>StandardError, :message=>"Test error", :backtrace=>["/home/jlledom/redhat/apisonator/lib/3scale/backend/job_fetcher.rb:50:in `fetch'", "/home/jlledom/redhat/apisonator/lib/3scale/backend/job_fetcher.rb:73:in `block in start'", "/home/jlledom/redhat/apisonator/lib/3scale/backend/job_fetcher.rb:66:in `loop'", "/home/jlledom/redhat/apisonator/lib/3scale/backend/job_fetcher.rb:66:in `start'"]}}
   ```
   The former without JSON seems more readable to me. Did you choose JSON format for any reason?
3. Somehow Bugsnag is able to catch errors wherever they happen, it would be good to understand how it does it and implement the same for `#notify` so when don't have two different paths for errors, and two output formats.

akostadinov: > The former without JSON seems more readable to me. Did you choose JSON format for any reason?

In general I agree. In opensearch though it's hard to retrieve (if possible) multi-line messages.

> Somehow Bugsnag is able to catch errors wherever they happen

There is some built-in Rack integration. I assume we need to send explicitly to `notify` in the worker only (where Rack is not involved).

jlledom (approved):

PR #382, opened 2024-04-10
Update test stuite

jlledom:
Some changes in the test suite after we added new feature to multiple job fetching.

Read each commit description for more details

### DIFF
--- a/spec/integration/worker_async_spec.rb
+++ b/spec/integration/worker_async_spec.rb
@@ -1,4 +1,3 @@
-require_relative '../spec_helper'
 require 'timecop'
 require 'daemons'
 require '3scale/backend/worker_async'
--- a/spec/integration/worker_async_spec.rb
+++ b/spec/integration/worker_async_spec.rb
@@ -13,7 +13,8 @@ module ThreeScale
       subject { Worker.new(async: true, job_fetcher: job_fetcher) }
 
       let(:storage) { Service.storage }
-      let(:redis_client ) { storage.instance_variable_get(:@redis_async) }
+      let(:redis_client ) { job_fetcher.instance_variable_get(:@redis).instance_variable_get(:@redis)
+                                       .instance_variable_get(:@redis).instance_variable_get(:@redis_async) }
 
       let(:provider_key) { 'a_provider_key' }
       let(:service_id) { 'a_service_id' }
--- a/spec/integration/worker_async_spec.rb
+++ b/spec/integration/worker_async_spec.rb
@@ -13,8 +13,7 @@ module ThreeScale
       subject { Worker.new(async: true, job_fetcher: job_fetcher) }
 
       let(:storage) { Service.storage }
-      let(:redis_client ) { job_fetcher.instance_variable_get(:@redis).instance_variable_get(:@redis)
-                                       .instance_variable_get(:@redis).instance_variable_get(:@redis_async) }
+      let(:resque_redis ) { job_fetcher.instance_variable_get(:@redis) }
 
       let(:provider_key) { 'a_provider_key' }
       let(:service_id) { 'a_service_id' }
@@ -41,9 +40,9 @@ module ThreeScale
           num.times { job_adder.call }
 
           # add some jobs to all queues
-          %w[resque:queue:main resque:queue:stats].each do |queue|
+          %w[queue:main queue:stats].each do |queue|
             2.times do
-              moved = redis_client.call('BLMOVE', 'resque:queue:priority', queue, 'RIGHT', 'LEFT', 5)
+              moved = resque_redis.brpoplpush("queue:priority", queue, 5)
               expect(moved).to be_truthy
               expect(moved).not_to be_empty
             end
--- a/spec/integration/worker_async_spec.rb
+++ b/spec/integration/worker_async_spec.rb
@@ -1,4 +1,3 @@
-require 'timecop'
 require 'daemons'
 require '3scale/backend/worker_async'
 
--- a/spec/integration/worker_async_spec.rb
+++ b/spec/integration/worker_async_spec.rb
@@ -1,3 +1,4 @@
+require 'timecop'
 require 'daemons'
 require '3scale/backend/worker_async'
 
@@ -80,29 +81,31 @@ module ThreeScale
         # verify that the stats usage keys have been updated correctly.
 
         n_reports = 10
-
-        without_resque_spec do
-          multi_job_adder.call(n_reports) # add jobs before worker started
-        end
-
-        work_task = Async { subject.work }
-
         stats_key = Stats::Keys.application_usage_value_key(
           service_id, app_id, metric_id, Period[:day].new(current_time)
         )
+        work_task = nil
 
-        report_waiter.call(stats_key, n_reports)
+        Timecop.freeze(current_time) do
+          without_resque_spec do
+            multi_job_adder.call(n_reports) # add jobs before worker started
+          end
 
-        without_resque_spec { 5.times { job_adder.call } } # add jobs after worker started and finished previous
-        report_waiter.call(stats_key, n_reports + 5)
+          work_task = Async { subject.work }
+
+          report_waiter.call(stats_key, n_reports)
+
+          without_resque_spec { 5.times { job_adder.call } } # add jobs after worker started and finished previous
+          report_waiter.call(stats_key, n_reports + 5)
+        end
 
         subject.shutdown
-        work_task.wait
+        work_task&.wait
 
         expect(storage.get(stats_key).to_i).to eq(n_reports + 5)
       rescue StandardError => e
         subject.shutdown
-        work_task.wait
+        work_task&.wait
         RSpec::Expectations.fail_with(e.message)
       end
     end
--- a/spec/unit/job_fetcher_spec.rb
+++ b/spec/unit/job_fetcher_spec.rb
@@ -1,5 +1,3 @@
-require 'spec_helper'
-
 module ThreeScale
   module Backend
     describe JobFetcher do
--- a/spec/unit/job_fetcher_spec.rb
+++ b/spec/unit/job_fetcher_spec.rb
@@ -175,11 +175,11 @@ module ThreeScale
           end
 
           context 'and `max` is nil' do
-            it 'returns all available jobs' do
-              expect(test_redis).to receive(:lpop).and_return(enqueued_jobs)
+            it 'returns one job' do
+              expect(test_redis).to receive(:lpop).and_return(enqueued_jobs.slice(0,1))
               result = subject.fetch(wait: false)
 
-              expect(result.size).to eq 3
+              expect(result.class).to eq Resque::Job
             end
           end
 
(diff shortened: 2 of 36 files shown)


spec/unit/job_fetcher_spec.rb
```diff
@@ -1,3 +1,5 @@
+require 'spec_helper'
+
```
akostadinov, line 2: I think this should be done by adding `-r spec_helper` to `rspec` parameters. Otherwise we better go and add this to all spec files and stop adding `-r`. For consistency I mean. I think it should either be required in the file or on command line but not some mix.

btw it is documented in https://github.com/3scale/apisonator/blob/master/DEVELOPMENT.md#running-tests-advanced

spec/integration/worker_async_spec.rb
```diff
@@ -14,7 +14,7 @@ module Backend
       subject { Worker.new(async: true, job_fetcher: job_fetcher) }
 
       let(:storage) { Service.storage }
-      let(:resque_redis ) { job_fetcher.instance_variable_get(:@redis) }
+      let(:redis_client ) { storage.instance_variable_get(:@redis_async) }
```
akostadinov, line 17: We have to use the client that `job_fetcher` uses. Otherwise, when the database for these is different, the test will not work. e.g.

```
CONFIG_REDIS_ASYNC=true CONFIG_INTERNAL_API_USER=system_app CONFIG_INTERNAL_API_PASSWORD=password CONFIG_QUEUES_MASTER_NAME=localhost:6379/5 CONFIG_REDIS_PROXY=localhost:6379/6 time -- bundle exec rake spec
```

I know that presently async client disregards logical database numbers. But
* we can sue separate databases
* I have in my backlog to add support for logical databases, see https://github.com/3scale/apisonator/pull/350#issuecomment-2020501532

spec/integration/worker_async_spec.rb
```diff
@@ -81,24 +81,21 @@ module Backend
...
+        end
 
-          work_task = Async { subject.work }
+        work_task = Async { subject.work }
 
-          stats_key = Stats::Keys.application_usage_value_key(
-            service_id, app_id, metric_id, Period[:day].new(current_time)
-          )
+        stats_key = Stats::Keys.application_usage_value_key(
+          service_id, app_id, metric_id, Period[:day].new(current_time)
```
akostadinov, line 92: So this should work as long as don't run the tests precisely at midnight so that some records go to last day and some to next day?

Maybe good enough, but still why remove timecop?

spec/unit/job_fetcher_spec.rb
```diff
@@ -138,6 +138,62 @@ module Backend
...
+              expect(test_redis).to receive(:blpop).and_return(enqueued_job)
+              result = subject.fetch
+
+              expect(result.class).to eq Resque::Job
+            end
+          end
+
+          context 'and `max` is not nil' do
+            it 'returns one job' do
+              expect(test_redis).to receive(:blpop).and_return(enqueued_job)
```
akostadinov, line 161: I doubt  the utility of such tests. Later on we may want to upgrade to `BLMPOP` and this test will fail. So it only tests that the implementation is what we have implmented but doesn't validate the behavior is what we want in case of an upgrade.

That's the reason I don't write such tests.

I can imagine an integration test, where `fetch` is called, then we insert a record and validate what we want is returned, but then how do we make sure that method was called only once...

Perhaps there is value, to make whoever updates the code later to think whether the new implementation is appropriate. So I guess good to stay. Some things are not really feasible to test though :/ same goes to the other sode about `lpop`, etc.

spec/unit/job_fetcher_spec.rb
jlledom: I get your point. IMO we should require it from code for all specs, that way we can run the tests out of the  box from Ruby Mine and also from console or `rake`. There's probably one single point during tests initialization where all tests pass through and we can just `require spec_helper` from there.

jlledom: OK. It seems there's not such single point. I'll just remove the require from all tests. When we need to launch a particular test from RubyMine, we'll have to manually add `-r rspec_helper` to the launcher.

akostadinov: If you edit the `rspec` template in idea, any automatically created run configurations inherit it.

mayorova: I'm OK either way. If `-r spec_helper` is already a documented way to run the tests, let's stick with it.

spec/integration/worker_async_spec.rb
jlledom: Makes sense, I made some changes. Ugly but effective

spec/integration/worker_async_spec.rb
jlledom: I'm my tests locally, `:current_time` is evaluated only once when running

```ruby
let(:current_time) { Time.now }
```

Then it keeps the same value during all the test execution

spec/unit/job_fetcher_spec.rb
jlledom: I think unit tests are never superfluous. They ensure the method behavior is correct, which is a safety measure when refactoring. When making changes on the code, it's normal to sometimes have to fix tests because they became invalid.

akostadinov: My take is that test is good when it tests the behavior that matters. I don't care what method is called as long as the desired behavior is met.

Here we test that `blpop` is being called. But then after 2 weeks I decide to refactor this and of course I have no recollection whatsoever that I've ever seen this code, so change it to `lpop`. Then the test fails because I'm not calling `blpop`. My reaction is to update the test with `lpop`.

But then the behavior would be incorrect.

If the test doesn't tell me somehow that I need to be blocking and I don't check that it blocks manually somehow, then this test didn't stop me from introducing an issue.

If the test could catch whether I'm blocking instead of checking for the specific method name, then if I update the code using `blmpop` the test will pass. And if I use `lpop`, the test would fail. So showing me usefully that behavior is or is not the intended one.

In the current form, the test relies on my personal discretion and I wouldn't recommend that 8-)

spec/integration/worker_async_spec.rb
akostadinov: `current_time` will obviously stay the same. But we are adding entries which don't use `current_time` but the actual time they are created. My assumption is that's the reason for timecop usage originally.

spec/integration/worker_async_spec.rb
```diff
@@ -14,7 +12,8 @@ module Backend
       subject { Worker.new(async: true, job_fetcher: job_fetcher) }
 
       let(:storage) { Service.storage }
-      let(:resque_redis ) { job_fetcher.instance_variable_get(:@redis) }
+      let(:redis_client ) { job_fetcher.instance_variable_get(:@redis).instance_variable_get(:@redis)
+                                       .instance_variable_get(:@redis).instance_variable_get(:@redis_async) }
```
akostadinov, line 16: I think it's much better if we use `::Resque.redis` and add `#blmove` to the async client. If ever we start relying on this method in sync client as well, we should write tests that validate same behavior.

I think the mess with getting instance variables is bigger than potential method behavior difference.

spec/unit/job_fetcher_spec.rb
jlledom: The behavior being tested here is `when 'wait' is set to true and 'max' is not nil it returns one job`. So when updating the test after changes, that's the behavior you should ensure this test is testing. The method used to verify it returns only one job doesn't matter as long as it actually verifies it returns one job. It also doesn't matter if this particular test blocks or not, because the test is not testing the blocking behavior.

Above there's a test that verifies the blocking behavior (`when 'wait' is set to true it calls a blocking redis command`). Your comment makes more sense applied to that test, because the blocking behavior is verified by expecting a call to `blpop`, but I couldn't find a better way to test that. If somebody updates the test to make it expect a call to `lpop` the test would pass and that would be incorrect, but I expect whoever updates the test in the future to read the test title and understand what is it trying to test before making changes to it.

spec/unit/job_fetcher_spec.rb
```diff
@@ -136,6 +136,62 @@ module Backend
...
+            expect(test_redis).to receive(:lpop).at_least(:once)
+            subject.fetch(wait: false)
+          end
+
+          context 'and `max` is nil' do
+            it 'returns all available jobs' do
+              expect(test_redis).to receive(:lpop).and_return(enqueued_jobs)
+              result = subject.fetch(wait: false)
+
+              expect(result.size).to eq 3
```
akostadinov, line 182: In fact it should return a single job when `max` is `nil`, waiting or not waiting.

So this shows the point that while describing the behavior is useful, because we stub database logic, at the end we receive what we have expected.

This is because `max` is set to `1` when not set. 

https://github.com/3scale/apisonator/blob/a3234d650cbf84644214ff936afdee79d67faa67/lib/3scale/backend/job_fetcher.rb#L70

The test exposes though something that I didn't know what to do with. That is 

https://github.com/3scale/apisonator/blob/a3234d650cbf84644214ff936afdee79d67faa67/lib/3scale/backend/job_fetcher.rb#L37

Presently if database returned more results, then `#fetch` returns them. I was wondering whether to install any special error reporting. But thought that such error is in general should be a rather extreme database misbehavior that I don't see how we would be supposed to recover from.

The caller would probably fail, report the error and we will have an indication what happened.

So we can either keep it or maybe add special reporting if database returned more entries than requested.

akostadinov: btw this test is the same as
https://github.com/3scale/apisonator/pull/382/files#diff-dcde5659e5ac647cc27329d886893e94a01e744c54eca60ccaff43f09ea489deL51-L55

spec/integration/worker_async_spec.rb
jlledom: Right, I'll add it back.

spec/unit/job_fetcher_spec.rb
jlledom: > In fact it should return a single job when max is nil, waiting or not waiting.

Right, the test was wrong, fixed.

> So this shows the point that while describing the behavior is useful, because we stub database logic, at the end we receive what we have expected.

Unit tests are about testing a unit independent from it's context, so stubs are needed. The problem is this test was wrong.

> this test is the same as https://github.com/3scale/apisonator/pull/382/files#diff-dcde5659e5ac647cc27329d886893e94a01e744c54eca60ccaff43f09ea489deL51-L55

It's not the same, that one is testing the returned job contains a valid value for the queue and payload, and it tests this in a particular context: when the job contain invalid UTF-8 characters. My test is part of a series of tests about what the method does according to received parameters. For instance, fetch returns a `Job` in some cases, and an `Array` in other cases, and the array can have one or many elements. What the method returns is not intuitive, e.g. It returns an array of one element when `wait` is `true` and `max` not `nil`, but it returns a `Job` when `wait` is `false` and `max` is `nil`. So depending on `wait`, it returns a class or another. The code is already merged and the caller depends on that behavior so we must test it and have tests that will break if that behavior changes.

> The test exposes though something that I didn't know what to do with. That is

That code is already merged, so that's what tests must test right now.

PR #384, opened 2024-05-08
THREESCALE-8404: Bump redis to 5.2.0

jlledom:
This comes from https://github.com/3scale/apisonator/pull/350. That PR is too big, it has a lot of outdated info and so many comments and commits it could lead to confusion. I decided to split it into smaller PRs easier to review and understand.

This PRs updates the redis gem to 5.2.0 but it doesn't include any new functionality compared to master.

Commits:
- https://github.com/3scale/apisonator/commit/5125eed61e7f71be67b54712aab7f88784876ba1: It updates the Gemfile and fixes some breaking changes.
- https://github.com/3scale/apisonator/commit/9fe39fe2acf17a14e48c34fdd322b7f5c56ee5eb: Updates some calls to the redis API that changed in redis 5. In particular `:sadd?`, `:srem?`, `:exists?` and `:pipelined`.
- https://github.com/3scale/apisonator/commit/d80612315dc0876686c2802d425a84500bcb7dea: The new API requires some major changes in the async client. For instance, a new implementation of the pipeline logic.
- https://github.com/3scale/apisonator/commit/640a5e4ea756cb61fa5432feb32fa613de851c1f: Update tests.


jlledom:
This comes from https://github.com/3scale/apisonator/pull/350#pullrequestreview-2053678308

> IIRC, when `@fetch_timeout` is reached, the method will return `nil`, not raise. A read timeout on the other hand is a low-level TCP timeout that shouldn't happen regularly.
> 
> On the other hand, we don't need to be very much bothered by a connection timeout, if we can just reconnect. It's good to be resilient in case of network glitches.
> 
> I would suggest though to log the error, can be without a trace, just dome text like:
> 
> > Que redis (ip/hostname of server if possible) connection timeout at #{**FILE**}:#{**LINE**}
> 
> In this way we can have an indication of network stability and see whether particular servers or locations are more problematic. This will help debugging and improving the cluster setup.

@akostadinov You're right, the docs say it should return `nil`, however it raises an exception when connecting to sentinels. I opened an issue upstream to fix it: https://github.com/redis/redis-rb/issues/1279

In the meantime we must capture the exception. And I don't think we need to log it because it would flood the logs with this error happening all the time, since it's the expected behavior.

jlledom:
From https://github.com/3scale/apisonator/pull/350#discussion_r1599011490:

> Why can provider key be nil? Should we conditionally add or should we raise an error?

@akostadinov I honestly don't remember why I did this, and it indeed looks wrong. I'm removing this line for now, let's see if something breaks.

lib/3scale/backend/job_fetcher.rb
```diff
@@ -85,6 +85,8 @@ def try_pop_from_queue(max=nil)
       def wait_pop_from_queue
         queue, encoded = @redis.blpop(*@queues, timeout: @fetch_timeout)
         [queue, [encoded]] if encoded
+      rescue RedisClient::ReadTimeoutError => _e
+        # Do nothing, we expect frequent timeouts from the sync client when the queue is empty
```
akostadinov, line 89: Could you then change the comment to clarify that this is a bug with sentinels and we should log these errors once it is fixed? Or in case it is easy to know that this sentinel or not, log for the cases where sentinels are not used?

jlledom: > Could you then change the comment to clarify that this is a bug with sentinels and we should log these errors once it is fixed?

OK

> Or in case it is easy to know that this sentinel or not, log for the cases where sentinels are not used?

Detecting sentinels from here is possible but too messy, I'd prefer to not do it. About logging, do you mean for `RedisClient::ReadTimeoutError` or for any error?

lib/3scale/backend/storage_async/pipeline.rb
```diff
@@ -73,6 +80,37 @@ def run(redis_async_client)
...
+          # there is an error in any of the commands of the pipeline we will
+          # raise an error (the first one that occurred).
+
+          first_err = nil
+
+          res = @commands.size.times.map do
+            begin
+              async_pipe.read_response
+            rescue ::Protocol::Redis::ServerError => e
+              first_err ||= e
```
akostadinov, line 103: Why not immediately raise here? Or is it related to freeing Redis resources?

akostadinov: I see the original code was like this. Fine with me to stay as is if you don't want to bother now figuring this out.

lib/3scale/backend/storage_async/methods.rb
```diff
@@ -0,0 +1,127 @@
...
+        #      For example, Redis returns 0 or 1 for the EXISTS command, but
+        #      redis-rb transforms that into a boolean.
+        #   3) There are a few methods that need to be treated differently and
+        #      do not fit in any of the previous categories. For example, SSCAN
+        #      which accepts a hash of options in redis-rb.
+        #
+        # All of this might be simplified a bit in the future using the
+        # "methods" in async-redis
+        # https://github.com/socketry/async-redis/tree/master/lib/async/redis/methods
+        # but there are some commands missing, so for now, that's not an option.
```
akostadinov, line 24: seems like these were removed and likely no plan to bring them back, perhaps remove that comment

lib/3scale/backend/job_fetcher.rb
akostadinov: About logging I mean `RedisClient::ReadTimeoutError`. Would be good to log any connectivity errors, but it will probably require more investigation. And I expect anything else to be pretty rare anyway so raising full blown should be ok.

lib/3scale/backend/storage_helpers.rb
```diff
@@ -241,6 +246,38 @@ def cfg_sentinels_handler(options)
             options
           end
```
akostadinov, line 247: Should we make sure that here sentinels are not empty? 

Up there we have `next if sentinel.empty?` and then `end.compact` so basically we can end up with an empty `options[:sentinels]`

lib/3scale/backend/storage_helpers.rb
```diff
@@ -98,9 +98,9 @@ def config_with(config,
               h[k] = val if val
             end.merge(options)
 
-            cfg_with_sentinels = cfg_sentinels_handler cfg
-
-            defaults.merge(ensure_url_param(cfg_with_sentinels))
+            cfg_compacted = cfg_compact cfg
+            cfg_with_sentinels = cfg_sentinels_handler cfg_compacted
+            cfg_defaults_handler cfg_with_sentinels, defaults
```
akostadinov, line 103: As far as I could track this down, it appears to me correct.

It's not very clear for the reader though without thoroughly tracking what's going on.

It first handles sentinel options that don't work with the `url`, `path`, etc. options. And then we normalize these options regardless of config being sentinel or not.

I think it would be much better to have something like:
```suggestion
            cfg_compacted = cfg_compact cfg
            cfg_sentinels_handler(cfg_compacted) || cfg_defaults_handler(cfg_compact, defaults)
```

i.e. if sentinels config is found, then return it, otherwise handle as a non-sentinel config. Do not apply both handler one after the other. To make clear that these are very different things and avoid the possibility of introducing conflicts between the two handlers in the future.

lib/3scale/backend/worker.rb
```diff
@@ -74,11 +74,11 @@ def perform(job)
       end
 
       def register_worker
-        redis.sadd(:workers, self)
+        redis.sadd(:workers, self.class.name)
```
akostadinov, line 77: :thinking: wouldn't this register all workers under the same name? Isn't the point here to be able to identify registered workers? For example `self.to_s` may result in `worker-backend-pod-10-10...` while `self.class.name` would be all the same for any pod.

lib/3scale/backend/storage_async/pipeline.rb
jlledom: Yeah, the comment above explains it, it's to match `redis-rb` behavior.

```ruby
# Redis returns an answer for each of the commands sent in the
# pipeline. But in order to keep compatibility with redis-rb, here, if
# there is an error in any of the commands of the pipeline we will
# raise an error (the first one that occurred).
```

jlledom:
> From [#350 (comment)](https://github.com/3scale/apisonator/pull/350#discussion_r1599011490):
> 
> > Why can provider key be nil? Should we conditionally add or should we raise an error?
> 
> @akostadinov I honestly don't remember why I did this, and it indeed looks wrong. I'm removing this line for now, let's see if something breaks.

@akostadinov I found why I did it. It's because some [tests fail](https://app.circleci.com/pipelines/github/3scale/apisonator/1123/workflows/49e83ff4-f0af-496c-9efc-6268e532a4e8/jobs/2038?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary). `redis-rb` 5 doesn't accept `nils` as parameters for it's redis calls, only strings. However, I think it's the tests what were wrong, not the code. I pushed some changes (https://github.com/3scale/apisonator/pull/384/commits/b4c0c8788a21526d54377679c8933b9a39f35f31)

lib/3scale/backend/storage_async/pipeline.rb
akostadinov: well, if we immediately raise, then the comment will still hold true. Why go over the remaining results if we are going to raise anyway? The only reaon I can think of is in case database resources will not be freed.

test/unit/storage_sync_test.rb
```diff
@@ -38,23 +40,16 @@ def test_redis_malformed_url
     end
   end
 
-  def test_redis_url_without_scheme
-    assert_nothing_raised do
-      StorageSync.send :new, url('foo')
-    end
-  end
```
akostadinov, line 45: we don't support urls without a schema anymore, correct?

test/unit/storage_sync_test.rb
```diff
@@ -115,9 +107,8 @@ def test_sentinels_simple_url
     }
 
     conn = StorageSync.send :new, Storage::Helpers.config_with(config_obj)
-    assert_sentinel_connector(conn)
-    assert_client_config(conn, url: "redis://#{config_obj[:url]}",
-                               sentinels: [{ host: '127.0.0.1', port: 26_379 }])
+    assert_sentinel_config(conn, url: "redis://#{config_obj[:url]}",
+                           sentinels: [{ host: '127.0.0.1', port: 26_379 }])
```
akostadinov, line 111: I hope we at least log any `a_malformed_url` detected. This is to inform users about typos.

lib/3scale/backend/storage_async/methods.rb
jlledom: The comment is outdated. `redis-async` now relies on https://github.com/socketry/protocol-redis to implement its methods. And if I'm not wrong, we could use that gem for `redis-rb` too. I'll take a look at some point in the future to make this simpler.

For now, I updated the comment on 054d0053e98e0a3fa82a99079867e7eba5446a35

lib/3scale/backend/storage_async/pipeline.rb
jlledom: I don't know the implementation details of the pipeline in `redis-rb`, maybe this can be done in some other way, but this is how it was before https://github.com/3scale/apisonator/commit/730442ef15a17daebfb3e584a582c965bb4b86df and it works so I didn't touch it.

lib/3scale/backend/storage_helpers.rb
jlledom: Are we not already doing it?

https://github.com/3scale/apisonator/blob/5125eed61e7f71be67b54712aab7f88784876ba1/lib/3scale/backend/storage_helpers.rb#L219

lib/3scale/backend/storage_helpers.rb
jlledom: The call to `cfg_defaults_handler` is required in all scenarios. First, because it ensures some defaults are always sent to the client:

https://github.com/jlledom/apisonator/blob/054d0053e98e0a3fa82a99079867e7eba5446a35/lib/3scale/backend/storage_helpers.rb#L42-L59

It also ensures there's an `url` parameter, which is used from both sentinels and non-sentinels clients. For non-sentinels, it contains the redis URL; for sentinels, it contains the master name.

The only thing `cfg_defaults_handler` does which is exclusive for non-sentinels is translating `url: 'unix:///path/to/socket'` to `path: '/path/to/socket'`.

lib/3scale/backend/storage_helpers.rb
akostadinov: I saw that but it is before the `compat` call. So we may still end up with empty sentinels list.

lib/3scale/backend/storage_async/methods.rb
```diff
@@ -0,0 +1,125 @@
...
+        #   2) Methods that need to be "boolified". These are methods for which
+        #      redis-rb returns a boolean, but redis just returns an integer.
+        #      For example, Redis returns 0 or 1 for the EXISTS command, but
+        #      redis-rb transforms that into a boolean.
+        #   3) There are a few methods that need to be treated differently and
+        #      do not fit in any of the previous categories. For example, SSCAN
+        #      which accepts a hash of options in redis-rb.
+        #
+        # All of this might be simplified a bit in the future using
+        # https://github.com/socketry/protocol-redis
```
akostadinov, line 22: I don't understand how it would help but at least link opens :smile:

jlledom: So `protocol-redis` implements a ruby method for each redis command, which ends up calling `call [COMMAND], [ARGS]`. AFAIK Both `async-redis` and `redis-rb` implement the `:call` method so theoretically it would be possible to put `protocol-redis` in the middle and make apisonator use the `procol-redis` API rather than the `redis-rb` API. Which would allow us to remove this class entirely. However I didn't investigate enough to be sure it's possible.

lib/3scale/backend/worker.rb
jlledom: Right, my bad. Fixed at https://github.com/3scale/apisonator/pull/384/commits/c1048fb2cf3564ce9d1e7502596daf7359dc1b8a

test/unit/storage_sync_test.rb
jlledom: Apparently we do. I don't know why I removed that test. Fixed in https://github.com/3scale/apisonator/pull/384/commits/461a712c4ed8242de9d61e66119b5dd17c6b5a3d

lib/3scale/backend/storage_helpers.rb
jlledom: You're right, I made some changes here: https://github.com/3scale/apisonator/pull/384/commits/9f85a48c3cd64c94f631f0660f5fccac99d45c7e

lib/3scale/backend/storage_async/methods.rb
mayorova: Thanks for the explanation @jlledom By looking at the README of that repo I also didn't understand... 
But now I understand.
Another link for reference (NOT saying that it needs to be added to the comment). https://www.rubydoc.info/gems/protocol-redis/0.2.0/Protocol/Redis/Methods/Hashes

lib/3scale/backend/alert_limit.rb
```diff
@@ -9,7 +9,7 @@ class AlertLimit
       attr_accessor :service_id, :value
 
       def save
-        storage.sadd(key_allowed_set(service_id), value.to_i) if valid?
+        storage.sadd?(key_allowed_set(service_id), value.to_i) if valid?
```
mayorova, line 12: @jlledom How did you figure out which calls needed to be updated (from `sadd` to `sadd?`) and which are not? Failing tests or smth else?

Cause there are still a lot of occurrences of `sadd` (for example). I guess it's fine as long as we're not expecting a boolean from them. But how can we make sure?

test/unit/storage_sync_test.rb
```diff
@@ -38,9 +40,15 @@ def test_redis_malformed_url
...
+  def test_redis_unknown_scheme
+    assert_raise ArgumentError do
+      StorageSync.send :new, url('myscheme://127.0.0.1:6379')
+    end
+  end
+
   def test_redis_url_without_scheme
     assert_nothing_raised do
-      StorageSync.send :new, url('foo')
+      StorageSync.send :new, url('backend-redis:6379')
```
mayorova, line 51:
```suggestion
      storage = StorageSync.send :new, url('backend-redis')
      assert_client_config(storage, url: URI('redis://backend-redis:6379'))
```

Just a suggestion... not sure if a good one (obviously, `assert_nothing_raised` is not needed in this case)

I was just surprised that the port was deleted, I think this actually modifies the test. And also, there is another test, `test_redis_host_and_port` which already tests host+port.

test/unit/storage_sync_test.rb
```diff
@@ -251,22 +237,27 @@ def test_sentinels_empty
 
   private
 
-  def assert_connection(client)
-    client.flushdb
-    client.set('foo', 'bar')
-    assert_equal 'bar', client.get('foo')
-  end
+  def assert_client_config(conn, **conf)
```
mayorova, line 240: Following the convention that typically the `assert_` methods follow, I would place the expected config (`**conf` in this case as a second argument, and the actual one (which will get derived from `conn`) as the 2nd argument.

That probably means that a normal hash should be used as the expected param instead of kwargs, but for me this order is a bit confusing... 

Same for `assert_sentinel_config`.

But maybe that's fine...

lib/3scale/backend/alert_limit.rb
jlledom: > How did you figure out which calls needed to be updated (from sadd to sadd?) and which are not? 

By reading the [Changelog](https://github.com/redis/redis-rb/blob/master/CHANGELOG.md), from **4.1.3** to **5.2.0**.

For `:pipelined` and `:exists?` it's easier because I replaced all occurrences. In the case of `:sadd` and `:sadd?` (same for `:srem[?]`). What I did is check all calls, and add the `?` to those where the return value is being used and needs to be a boolean. Then passing tests and making a lot of manual testing in different scenarios. There are not many occurrences of those, so it's doable.

test/unit/storage_sync_test.rb
jlledom: I applied this: https://github.com/3scale/apisonator/pull/384/commits/de82ec273a4888f7f5d0caf7ab96786d7cac90c1

test/unit/storage_sync_test.rb
jlledom: Sound reasonable. I did it in https://github.com/3scale/apisonator/pull/384/commits/65d05ef730d3684e2113b7e206cdad204c639813

spec/unit/service_spec.rb
```diff
@@ -368,44 +368,44 @@ module Backend
       describe '.active?' do
         it 'returns true when service is active' do
           [
-            { state: :active, id: '8001' },
-            { state: 'active', id: '8001' },
+            { state: :active, id: '8001', provider_key: 'foo' },
```
mayorova, line 371: I don't understand this change with adding `provider_key` to all `Service.save!` calls...

is this required now for some reason? I think it shouldn't?

Why were the tests passing without the provider key before?

mayorova:
Looks good @jlledom great job! :clap: 


I'm just concerned about that change with the `provider_key` in tests. Otherwise, all looks nice.

spec/unit/service_spec.rb
jlledom: See this: https://github.com/3scale/apisonator/pull/384#issuecomment-2109914809
Then this: https://github.com/3scale/apisonator/pull/384#issuecomment-2110046219

Do you think the tests were right and it should be possible that `persist_sets` actually receives an empty provider hey?

mayorova: I've been checking, and it seems that indeed `provider_key` is an important part of the Service model on apisonator. I thought it wasn't the case, because `service_token` is used for most operations on the Service Management API, but it seems that when Service is created or updated through the internal API, the`provider_key` is indeed provided :grimacing: https://github.com/3scale/porta/blob/4b2eb62c3acea37ec32b0f1339a2d63f6b8ddef1/app/lib/backend/model_extensions/service.rb#L20

So, I guess it's fine to add it to tests.

mayorova (approved):

akostadinov (approved):

PR #385, opened 2024-05-09
CircleCI: Sentinels

jlledom:
This is a draft, only for launching CircleCI workflows


jlledom:
Closing this after giving up. Circle CI doesn't provide a way to mount files on the containers using the docker executor. And if we use the machine executor we would need to mount the entire ruby 3 framework from a raw ubuntu image, and invent a way to properly cache it to not have to do it in every pipeline run, it's not worth IMO. Also, using machine executor I couldn't manage to connect to redis.

jlledom:
Let's try again...

lib/3scale/backend/async_redis/client.rb
```diff
@@ -19,7 +19,7 @@ def call(opts)
             uri = URI(opts[:url])
 
             credentials = [ uri.user || opts[:username], uri.password || opts[:password]]
-            db = uri.path[1..-1].to_i if uri.path
+            db = uri.path[1..-1] if uri.path
```
akostadinov, line 22: could you add a test that in absence of a `db` part of URL, the protocol will not set that? I think it is worth it.

jlledom: Sure

bors.toml
```diff
@@ -1,3 +1,3 @@
-status = ["workflow"]
-pr_status = ["workflow"]
+status = ["ci/circleci: build"]
+pr_status = ["ci/circleci: build", "workflow"]
```
akostadinov, line 2: what is this doing?

.circleci/ca-circleci-key.pem
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
akostadinov, line 1: acn we somehow mark the flle as no secret? Actually is it a secret?

bors.toml
jlledom: There's a `ci/circleci: build` job always expected and required but never executed, because I removed it.

image

I thought it would disappear when I remove it from `bors.toml`, but it didn't work.

.circleci/ca-circleci-key.pem
jlledom: It's not a secret because is a fake cert I created for circleci, I marked it as non-secret using `rh-gitleaks` (https://github.com/3scale/apisonator/pull/385/commits/b1a50d44859008697bfd84e294768441c81691fa#diff-f40ac94aac00bf364ae4ebef5689f7a2b61d63788030902c529d48d6d7d30ffc).

Check https://source.redhat.com/departments/it/it-information-security/wiki/pattern_distribution_server#handling-false-positives

bors.toml
akostadinov: in repo settings probably

jlledom: Yeah, it was there, thanks.

script/config/podman-compose.yml
```diff
@@ -3,88 +3,134 @@ services:
   redis-master:
     image: redis:6.2-alpine
     container_name: redis-master
-    network_mode: "host"
-    ports:
-      - "6379:6379"
+    network_mode: host
```
akostadinov, line 6: btw, host mode is the default for root. If we just want to connect the different things together, then also running containers inside a single pod is a good way to isolate. But just saying, it's fine to stay.

script/config/tls/master_template.conf
```diff
@@ -0,0 +1,11 @@
+port 0
+tls-port 46380
+tls-cert-file /etc/redis.crt
+tls-key-file /etc/redis.key
+tls-ca-cert-file /etc/ca-root-cert.pem
+tls-auth-clients optional
+tls-replication yes
+user default off
+user porta on >sup3rS3cre1! ~* &* +@all
+user apisonator on >secret#Passw0rd ~* &* +@all
```
akostadinov, line 10: Should we mark these as non-sectets too?

script/config/tls/replica2_template.conf
```diff
@@ -0,0 +1,11 @@
...
+tls-port 46382
+tls-cert-file /etc/redis.crt
+tls-key-file /etc/redis.key
+tls-ca-cert-file /etc/ca-root-cert.pem
+tls-auth-clients optional
+tls-replication yes
+user default off
+user porta on >sup3rS3cre1! ~* &* +@all
+masteruser porta
+masterauth sup3rS3cre1!
```
akostadinov, line 11: Should we mark these as non-sectets too?

script/config/tls/sentinel1_template/sentinel.conf
```diff
@@ -0,0 +1,19 @@
...
+tls-key-file "/etc/redis.key"
+tls-ca-cert-file "/etc/ca-root-cert.pem"
+tls-auth-clients optional
+tls-replication yes
+user default off sanitize-payload &* -@all
+user sentinel on >secret#Passw0rd ~* &* +@all
+sentinel auth-user redis-master porta
+sentinel auth-pass redis-master sup3rS3cre1!
+sentinel sentinel-user sentinel
+sentinel sentinel-pass secret#Passw0rd
```
akostadinov, line 18: Should we mark these as non-sectets too?

script/config/tls/sentinel2_template/sentinel.conf
```diff
@@ -0,0 +1,19 @@
...
+tls-key-file "/etc/redis.key"
+tls-ca-cert-file "/etc/ca-root-cert.pem"
+tls-auth-clients optional
+tls-replication yes
+user default off sanitize-payload &* -@all
+user sentinel on >secret#Passw0rd ~* &* +@all
+sentinel auth-user redis-master porta
+sentinel auth-pass redis-master sup3rS3cre1!
+sentinel sentinel-user sentinel
+sentinel sentinel-pass secret#Passw0rd
```
akostadinov, line 18: Should we mark these as non-sectets too?

script/config/tls/sentinel3_template/sentinel.conf
```diff
@@ -0,0 +1,19 @@
...
+tls-key-file "/etc/redis.key"
+tls-ca-cert-file "/etc/ca-root-cert.pem"
+tls-auth-clients optional
+tls-replication yes
+user default off sanitize-payload &* -@all
+user sentinel on >secret#Passw0rd ~* &* +@all
+sentinel auth-user redis-master porta
+sentinel auth-pass redis-master sup3rS3cre1!
+sentinel sentinel-user sentinel
+sentinel sentinel-pass secret#Passw0rd
```
akostadinov, line 18: Should we mark these as non-sectets too?

script/config/tls/master_template.conf
jlledom: Well, these ones are obviously fake, but for some reason don't trigger an alert, so I think we are fine.

jlledom: Same for all other occurrences of this

script/config/podman-compose.yml
jlledom: Is there a particular way to run the service inside a pod? I though they were already doing so. Also, I need some of them, like the master or the sentinels to be accessible from the test suite which runs in the Ubuntu VM in CircleCI. I tried a lot of things and only `network_mode: host` worked.

script/config/sentinel1_template/sentinel1.conf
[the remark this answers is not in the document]
jlledom: I moved the templates under the `examples` folder here: https://github.com/3scale/apisonator/pull/385/commits/40797b5f9eee6a0b478ebe764d11701d3d637bea

test/test_helpers/twemproxy.rb
```diff
@@ -0,0 +1,168 @@
...
+
+        def unmock_storage_clients
+          STORAGE_CLASSES.each { |klass| unmock_storage_client!(klass) }
+        end
+
+        private
+
+        def mock_storage_client!(storage_client_class)
+          class << storage_client_class
+            # ensure this does not get overwritten
```
mayorova, line 64: Just wondering - where it can potentially be overwritten?

jlledom: No idea. I should have mentioned I copied and pasted this from the old code, when twemproxy was the only way to run tests: https://github.com/3scale/apisonator/pull/370/files#diff-51187addae661483cd1f02434a8e6f4ab93ebacf0b3fe7ae54e6949231696da5

There's a lot on this file that feels like arcane magic for me.

script/config/podman-compose.yml
akostadinov: For accessing containers from host, one would need to map ports when starting containers.

.circleci/config.yml
```diff
@@ -1,42 +1,217 @@
...
+attach-to-workspace: &attach-to-workspace
+  attach_workspace:
+    at: /home/circleci
 
-    steps:
-      - checkout
+executors:
+  ubuntu_vm:
+    machine:
+      image: ubuntu-2204:2024.04.4
```
akostadinov, line 19: Do we need to hardcode the version?

.circleci/config.yml
```diff
@@ -1,42 +1,217 @@
...
+            fi
+            echo '. "$HOME/.asdf/asdf.sh"' > ~/.bashrc
+            source .bashrc
+            asdf plugin add ruby https://github.com/asdf-vm/asdf-ruby.git
+            asdf install ruby 3.0.2
+            echo 'ruby 3.0.2' > $HOME/project/.tool-versions
+      - save_cache:
+          paths:
+            - ../.asdf
+          key: asdf-ruby-3.0.2-v1
```
akostadinov, line 71: better use YAML anchor instead of hardcoding the key so that it can be changed in a single place later

.circleci/config.yml
```diff
@@ -1,42 +1,217 @@
...
       - run:
-          name: bundle install
+          name: Install dependencies
           command: |
             # Required for the license_finder
             # The Gemfile.lock includes nokogiri but bundler
             # would install native gem instead if not forced
-            bundle config set --local force_ruby_platform true
+            source $HOME/.bashrc
+            bundle config set --local force_ruby_platform 'true'
```
akostadinov, line 85: we shouldn't need this but since it was there , I guess it can stay for now.

akostadinov (approved):
Ideally we would run the tests in a proper container instead of the host. That can be achieved by attaching a volume to the container.

Also maybe run all containers in the same pod for seemless network visibility.

Similar to https://github.com/3scale/searchd/blob/fde88c8aed92461a9fd2d7127ae6cb6058037dec/.github/workflows/container-image-buildah.yml#L117-L126

But this workflow is already such a huuge improvement that it's totally fine to go in as it is and make further improvements whenever that's a high enough priority.

script/config/podman-compose.yml
jlledom: I tried port mapping first but if I remember correctly I couldn't manage to make containers see each other. Maybe sharing pod could have helped.

.circleci/config.yml
jlledom: No, we could use `current`, but I think is better to stick to one version for stability.

https://circleci.com/developer/machine/image/ubuntu-2004

.circleci/config.yml
jlledom: Done at https://github.com/3scale/apisonator/pull/385/commits/b1054977693e2509b970d67ab9812ba101327327

.circleci/config.yml
jlledom: Done at https://github.com/3scale/apisonator/pull/385/commits/b1054977693e2509b970d67ab9812ba101327327

jlledom:
> Ideally we would run the tests in a proper container instead of the host. That can be achieved by attaching a volume to the container.
> 

I tried that, but CircleCI doesn't allow attaching a volume to a container when launching containers from the docker executor.

https://circleci.com/docs/using-docker/#docker-benefits-and-limitations

akostadinov:
> I tried that, but CircleCI doesn't allow attaching a volume to a container when launching containers from the docker executor.

If you can run random commands, it can be achieved like this. But as I said, that can be explored in the future.

PR #389, opened 2024-05-20
Async mode: Add support for Redis logical DBs

jlledom:
Now the Async mode supports Redis logical databases, and honors URLs like `redis://localhost:6379/6`

The support has been also added for sentinels. A typical config would look like this:

```ruby
CONFIG_REDIS_PROXY=redis://redis-master/6
CONFIG_REDIS_SENTINEL_HOSTS=redis://localhost:26379,redis://localhost:26380,redis://localhost:26381
```

### DIFF
--- a/lib/3scale/backend/storage_async/client.rb
+++ b/lib/3scale/backend/storage_async/client.rb
@@ -1,6 +1,7 @@
 require 'async/io'
 require 'async/redis/client'
 require 'async/redis/sentinels'
+require 'async/redis/protocol/extended_resp2'
 
 module ThreeScale
   module Backend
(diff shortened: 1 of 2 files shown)


mayorova:
Hmm, is there a JIRA for this?

akostadinov:
Instead of monkey patching upstream protocol class, it is a much better approach to use our own class and use the upstream protocol class with the composition pattern. Like I commented here: https://github.com/3scale/apisonator/pull/350#issuecomment-2020501532

For me this worked very easy OOB. ALthough it's not necessary to be just a copy of it. Just to use the upstream protocol class instead of monkey patching it. Unless I miss some details...

jlledom:
> Hmm, is there a JIRA for this?

No, Sorry. This comes from https://github.com/3scale/apisonator/pull/350. I forgot to mention it in the PR description. Anyway, this is not directly related to TLS so maybe it  should have it's own Jira.

jlledom:
@akostadinov @mayorova I renamed the branch to add `THREESCALE-11061` to the name and Github automatically closed this issue. I opened https://github.com/3scale/apisonator/pull/390

PR #390, opened 2024-05-21
THREESCALE-11061: Async mode: Add support for Redis logical DBs

jlledom:
This comes from of https://github.com/3scale/apisonator/pull/350.

Now the Async mode supports Redis logical databases, and honors URLs like `redis://localhost:6379/6`

The support has been also added for sentinels. A typical config would look like this:

```ruby
CONFIG_REDIS_PROXY=redis://redis-master/6
CONFIG_REDIS_SENTINEL_HOSTS=redis://localhost:26379,redis://localhost:26380,redis://localhost:26381
```

**Jira**: https://issues.redhat.com/browse/THREESCALE-11061

### DIFF
--- a/lib/3scale/backend/storage_async/client.rb
+++ b/lib/3scale/backend/storage_async/client.rb
@@ -1,6 +1,7 @@
 require 'async/io'
 require 'async/redis/client'
 require 'async/redis/sentinels'
+require 'async/redis/protocol/extended_resp2'
 
 module ThreeScale
   module Backend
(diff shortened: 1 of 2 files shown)


jlledom:
From https://github.com/3scale/apisonator/pull/389#issuecomment-2121163703:

@akostadinov 

> Instead of monkey patching upstream protocol class, it is a much better approach to use our own class and use the upstream protocol class with the composition pattern. Like I commented here: [#350 (comment)](https://github.com/3scale/apisonator/pull/350#issuecomment-2020501532)
> 

I don't understand what you mean. I'm not monkey patching the original protocol class, I added a new class `ExtendedRESP2` which in fact is pretty similar to the example you mentioned:

https://github.com/socketry/async-redis/blob/feb67542dc27947fddbc647ad4b7289645975698/examples/auth/wrapper.rb#L52-L74

lib/3scale/backend/storage_async/client.rb
```diff
@@ -71,20 +72,34 @@ def initialize_client(opts)
...
           role = opts[:role] || :master
+          protocol = make_redis_protocol(opts)
+
+          Async::Redis::SentinelsClient.new(name, opts[:sentinels], role, protocol)
+        end
+
+        # RESP2 with support for logical DBs
+        def make_redis_protocol(opts)
+          uri = URI(opts[:url] || "")
+          db = uri.path[1..-1]
```
mayorova, line 92: I'm thinking whether we need to report an error if the `db` is not an integer.
Or just let the client respond to `select {invalid-value}` with whatever it likes?

akostadinov: Yeah, I wouldn't bother reraising this.

jlledom: agree with @akostadinov

akostadinov (approved):
Nice! It's a bummer that `Redis` is somehow conflicting though. I'm fine to keep it as is. One possible option is to call the module `AsyncRedis` for example. But I leave it to your preference.

jlledom:
> Nice! It's a bummer that `Redis` is somehow conflicting though. I'm fine to keep it as is. One possible option is to call the module `AsyncRedis` for example. But I leave it to your preference.

Yeah, I think I'm gonna rename it.

jlledom:
> > Nice! It's a bummer that `Redis` is somehow conflicting though. I'm fine to keep it as is. One possible option is to call the module `AsyncRedis` for example. But I leave it to your preference.
> 
> Yeah, I think I'm gonna rename it.

@akostadinov done in https://github.com/3scale/apisonator/pull/390/commits/7a2da06b635818bfda7d0401b7357d322e177655

mayorova (approved):

akostadinov (approved):

PR #391, opened 2024-05-27
THREESCALE-8404: Add ACL and TLS support for Redis

jlledom:
This supersedes https://github.com/3scale/apisonator/pull/350. That PR is too messy and contains a lot of wrong or outdated info.

### Jira Issue:
[THREESCALE-8404](https://issues.redhat.com/browse/THREESCALE-8404)

### Summary of changes:
- Implement support for ACL and TLS (https://github.com/3scale/apisonator/commit/914f97c59faf0be8cbdef5cf9323bccd640aaded):
  - The existing async client (`StorageAsync::Client`) is very simple now. It got too complex due to all scenarios it must support, so I extracted most of its logic to the new class `AsyncRedis::Client`.
  - `AsyncRedis::Client` is a wrapper for the two Redis clients we use in async mode: one for single instances and one for sentinels. This implements support for all possible scenarios.
  - `AsyncRedis::EndpointHelpers` is a utility module which contains some logic that is used from several places.
  - Our custom protocol `AsyncRedis::Protocol::ExtendedRESP2` now accepts ACL credentials.
  - We also use our own client for Sentinels: `AsyncRedis::SentinelsClientACLTLS` which extends the upstream client `Async::Redis::SentinelsClient` to add it ACL and TLS support.
- Add new ENV variables to configure ACL and TLS (https://github.com/3scale/apisonator/commit/8104713eb612f1d3f4d95f83f62d8a10fb80cb22):
  - These new variables are going to be set directly from the operator.
  - Check: [THREESCALE-11020](https://issues.redhat.com/browse/THREESCALE-11020) and [THREESCALE-11021](https://issues.redhat.com/browse/THREESCALE-11021).
- Update tests (https://github.com/3scale/apisonator/commit/a7ddc79e76a84ca5ea762704f1edf874d25b25d8):
  - It installs `fakefs`. This is used to dynamically generate RSA, DSA and EC keys to be tested from `StorageAsyncTest`. In async mode we only support these 3 algorithms for TLS but I think that's enough for most of cases and we can add new algorithms in the future if some client requires it.
  - `TestHelpers::Certificates` contains the logic to generate keys and certificates for tests.
  - Added TLS and ACL tests in `StorageAsyncTest` and `StorageSyncTest`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -22,50 +22,107 @@ variables.
 ###  CONFIG_REDIS_PROXY
 
 - Redis URL for the data DB.
-- Optional. Defaults to "redis://localhost:6379".
+- Optional. Defaults to `redis://localhost:6379`.
 - Applies to: listener, worker, cron.
 - Format: string.
+
+###  CONFIG_REDIS_USERNAME
+
+- Redis user name
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+###  CONFIG_REDIS_PASSWORD
+
+- Redis password
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+###  CONFIG_REDIS_SSL
+
+- Whether use SSL to connect to Redis
+- Optional. Defaults to not set
+- Will be considered `true` when the URL schema is `rediss`
+- Applies to: listener, worker, cron.
+- Format: `true` or `false`.
+
+###  CONFIG_REDIS_CA_FILE
+
+- Certification authority to validate Redis server TLS connections with
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
+
+###  CONFIG_REDIS_CERT
+
+- The path to the client SSL certificate
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
+
+###  CONFIG_REDIS_PRIVATE_KEY
+
+- The path to the client SSL private key
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
 
 ###  CONFIG_REDIS_SENTINEL_HOSTS
 
 - URL of Redis sentinels.
 - Optional. Required only when using a Redis cluster with sentinels.
 - Applies to: listener, worker, cron.
-- Format: list of URLs separated by ",".
+- Format: list of URLs separated by `,`.
+
+### CONFIG_REDIS_SENTINEL_USERNAME
+
+- Sentinels user name
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+### CONFIG_REDIS_SENTINEL_PASSWORD
+
+- Sentinels password
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
 
 ###  CONFIG_REDIS_SENTINEL_ROLE
 
 - Asks the sentinel for the URL of the master or a slave.
-- Optional. Defaults to master. Applies only when using a Redis cluster with
+- Optional. Defaults to `master`. Applies only when using a Redis cluster with
 sentinels.
 - Applies to: listener, worker, cron.
-- Format: master or slave.
+- Format: `master` or `slave`.
 
 ###  CONFIG_REDIS_CONNECT_TIMEOUT
 
 - Connect timeout.
-- Optional. Defaults to 5.
+- Optional. Defaults to `5`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
 ###  CONFIG_REDIS_READ_TIMEOUT
 
 - Read timeout.
-- Optional. Defaults to 3.
+- Optional. Defaults to `3`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
 ###  CONFIG_REDIS_WRITE_TIMEOUT
 
 - Write timeout.
-- Optional. Defaults to 3.
+- Optional. Defaults to `3`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
 ### CONFIG_REDIS_MAX_CONNS
 
 - Max number of connections.
-- Optional. Defaults to 10.
+- Optional. Defaults to `10`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
@@ -75,51 +132,108 @@ sentinels.
 ### CONFIG_QUEUES_MASTER_NAME
 
 - Redis URL for the queues DB.
-- Required when `RACK_ENV` is not "test" or "development". Defaults to
-"redis://localhost:6379" in those 2 environments.
+- Required when `RACK_ENV` is not `test` or `development`. Defaults to
+`redis://localhost:6379` in those 2 environments.
 - Applies to: listener, worker, cron.
 - Format: string.
+
+###  CONFIG_QUEUES_USERNAME
+
+- Redis user name
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+###  CONFIG_QUEUES_PASSWORD
+
+- Redis password
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+###  CONFIG_QUEUES_SSL
+
+- Whether use SSL to connect to Redis
+- Optional. Defaults to not set
+- Will be considered `true` when the URL schema is `rediss`
+- Applies to: listener, worker, cron.
+- Format: `true` or `false`.
+
+###  CONFIG_QUEUES_CA_FILE
+
+- Certification authority certificate Redis should trust to accept TLS connections
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
+
+###  CONFIG_QUEUES_CERT
+
+- User certificate to connect to Redis through TLS
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
+
+###  CONFIG_QUEUES_PRIVATE_KEY
+
+- User key to connect to Redis through TLS
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: path to file as string.
 
 ### CONFIG_QUEUES_SENTINEL_HOSTS
 
 - URL of Redis sentinels.
 - Optional. Required only when using a Redis cluster with sentinels.
 - Applies to: listener, worker, cron.
-- Format: list of URLs separated by ",".
+- Format: list of URLs separated by `,`.
+
+### CONFIG_QUEUES_SENTINEL_USERNAME
+
+- Sentinels user name
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+### CONFIG_QUEUES_SENTINEL_PASSWORD
+
+- Sentinels password
+- Optional. Defaults to not set.
+- Applies to: listener, worker, cron.
+- Format: string.
 
 ### CONFIG_QUEUES_SENTINEL_ROLE
 
 - Asks the sentinel for the URL of the master or a slave.
-- Optional. Defaults to master. Applies only when using a Redis cluster with
+- Optional. Defaults to `master`. Applies only when using a Redis cluster with
 sentinels.
 - Applies to: listener, worker, cron.
-- Format: master or slave.
+- Format: `master` or `slave`.
 
 ### CONFIG_QUEUES_CONNECT_TIMEOUT
 
 - Connect timeout.
-- Optional. Defaults to 5.
+- Optional. Defaults to `5`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
 ### CONFIG_QUEUES_READ_TIMEOUT
 
 - Read timeout.
-- Optional. Defaults to 3.
+- Optional. Defaults to `3`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
 ### CONFIG_QUEUES_WRITE_TIMEOUT
 
 - Write timeout.
-- Optional. Defaults to 3.
+- Optional. Defaults to `3`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
 ### CONFIG_QUEUES_MAX_CONNS
 
 - Max number of connections.
-- Optional. Defaults to 10.
+- Optional. Defaults to `10`.
 - Applies to: listener, worker, cron.
 - Format: integer (seconds).
 
@@ -129,14 +243,14 @@ sentinels.
 ### CONFIG_INTERNAL_API_USER
 
 - The user needed when sending requests to the internal API.
-- Required when `RACK_ENV` is "production", optional otherwise.
+- Required when `RACK_ENV` is `production`, optional otherwise.
 - Applies to: listener.
 - Format: string.
 
 ### CONFIG_INTERNAL_API_PASSWORD
 
 - The password needed when sending requests to the internal API.
-- Required when `RACK_ENV` is "production", optional otherwise.
+- Required when `RACK_ENV` is `production`, optional otherwise.
 - Applies to: listener.
 - Format: string.
 
@@ -167,7 +281,7 @@ account of Porta.
 
 - Name of the metric configured in the master account of Porta to track the
 number of report calls. Applies only when `CONFIG_MASTER_SERVICE_ID` is set.
-- Optional. Defaults to "transactions".
+- Optional. Defaults to `transactions`.
 - Applies to: listener.
 - Format: string.
 
@@ -175,7 +289,7 @@ number of report calls. Applies only when `CONFIG_MASTER_SERVICE_ID` is set.
 
 - Name of the metric configured in the master account of Porta to track the
 number of authorize calls. Applies only when `CONFIG_MASTER_SERVICE_ID` is set.
-- Optional. Defaults to "transactions/authorize".
+- Optional. Defaults to `transactions/authorize`.
 - Applies to: listener.
 - Format: string.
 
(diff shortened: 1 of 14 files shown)


lib/3scale/backend.rb
```diff
@@ -60,8 +60,8 @@ class << self
       def new_resque_redis
         QueueStorage.connection(
           environment,
-          configuration,
-          )
+          configuration
+        )
```
akostadinov, line 64: original seems alright

test/test_helpers/certificates.rb
```diff
@@ -0,0 +1,67 @@
...
+require 'openssl'
+
+module TestHelpers
+  module Certificates
+    def create_key(alg)
+      case alg
+      when :rsa
+        OpenSSL::PKey::RSA.new(2048)
+      when :dsa
+        OpenSSL::PKey::DSA.new(2048)
```
akostadinov, line 25: I believe DSA is generally deprecated. Better don't add support for it.

docs/configuration.md
```diff
@@ -26,13 +26,69 @@ variables.
 - Applies to: listener, worker, cron.
 - Format: string.
 
+###  CONFIG_REDIS_USERNAME
+
+- Redis ACL user name
```
akostadinov, line 31: Do we need `ACL` here? When I read it I'm starting to wonder, what is the difference between a username and an ACL username :angel:

docs/configuration.md
```diff
@@ -26,13 +26,69 @@ variables.
...
+
+- Redis ACL password
+- Optional. Defaults to empty.
+- Applies to: listener, worker, cron.
+- Format: string.
+
+###  CONFIG_REDIS_SSL
+
+- Whether use SSL to connect to Redis
+- Optional. Defaults to empty.
```
akostadinov, line 46: boolean defaults to `empty`?

docs/configuration.md
```diff
@@ -26,13 +26,69 @@ variables.
...
+- Format: path to file as string.
+
 ###  CONFIG_REDIS_SENTINEL_HOSTS
 
 - URL of Redis sentinels.
 - Optional. Required only when using a Redis cluster with sentinels.
 - Applies to: listener, worker, cron.
 - Format: list of URLs separated by ",".
 
+### CONFIG_REDIS_SENTINEL_USERNAME
```
akostadinov, line 78: We can't use CONFIG_REDIS_USERNAME?

lib/3scale/backend/configuration.rb
```diff
@@ -39,11 +39,12 @@ def parse_int(value, default)
...
 
       # Add configuration sections
-      config.add_section(:queues, :master_name, :sentinels, :role,
-                         :connect_timeout, :read_timeout, :write_timeout, :max_connections)
-      config.add_section(:redis, :url, :proxy, :sentinels, :role,
-                         :connect_timeout, :read_timeout, :write_timeout, :max_connections,
-                         :async)
+      config.add_section(:queues, :master_name, :username, :password, :ssl, :ssl_params, :sentinels,
+                         :sentinel_username, :sentinel_password, :role, :connect_timeout, :read_timeout, :write_timeout,
+                         :max_connections)
```
akostadinov, line 44: So for Que we don't use async, correct?

lib/3scale/backend/storage_async/client.rb
```diff
@@ -35,7 +35,7 @@ def instance(reset = false)
         end
 
         def initialize(opts)
-          @redis_async = initialize_client(opts)
+          @redis_async = AsyncRedis::Client.call(opts)
```
akostadinov, line 38:
```suggestion
          @redis_async = AsyncRedis::Client.connect(opts)
```
idk, `call` doesn't sound right to me but leave it as you prefer

openshift/3scale_backend.conf
```diff
@@ -22,14 +22,34 @@ ThreeScale::Backend.configure do |config|
...
+  config.queues.sentinel_password = "#{ENV['CONFIG_QUEUES_SENTINEL_PASSWORD']}"
   config.queues.role = "#{ENV['CONFIG_QUEUES_SENTINEL_ROLE']}".to_sym
   config.queues.connect_timeout = parse_int_env('CONFIG_QUEUES_CONNECT_TIMEOUT')
   config.queues.read_timeout = parse_int_env('CONFIG_QUEUES_READ_TIMEOUT')
   config.queues.write_timeout = parse_int_env('CONFIG_QUEUES_WRITE_TIMEOUT')
   config.queues.max_connections = parse_int_env('CONFIG_QUEUES_MAX_CONNS')
   config.redis.proxy = "#{ENV['CONFIG_REDIS_PROXY']}"
+  config.redis.username = "#{ENV['CONFIG_REDIS_USERNAME']}"
+  config.redis.password = "#{ENV['CONFIG_REDIS_PASSWORD']}"
+  config.redis.ssl =  parse_boolean_env('CONFIG_REDIS_SSL')
```
akostadinov, line 44: Do we have a use case where user can't just set `rediss://` instead? Even for centinels I understand we set uri. So we can infer this from the uri. I mean, can we remove this option?

lib/3scale/backend.rb
jlledom: IMO it looks better like this, anyway I think the indentation changes were automatic. I'll revert them.

akostadinov: it's not only indentation, it is also the `,` which is nice IMO because it limits changed lines for review.

test/test_helpers/certificates.rb
jlledom: I don't think we can decide what is supported or not, because the given key is simply passed to `OpenSSL::PKey` (https://ruby-doc.org/stdlib-3.0.2/libdoc/openssl/rdoc/OpenSSL/PKey.html) which supports DSA, at least for Ruby 3.0.2

docs/configuration.md
jlledom: Yes, it defaults to not be set.

docs/configuration.md
jlledom: No because ACL for sentinels can be different to ACL for master. See https://github.com/redis/redis-rb/issues/1213.

lib/3scale/backend/configuration.rb
jlledom: Actually we do. We send the params in the `queues` section to the redis client:

https://github.com/3scale/apisonator/blob/04e28f36fddde408c41afd960aa0240b0ccb87ea/lib/3scale/backend/queue_storage.rb#L4-L13

However, `Storage` will read `config.redis.async` to decide either using the sync or async redis client:

https://github.com/3scale/apisonator/blob/593beaba401b4322aeae779b889f3e321800a3fe/lib/3scale/backend/storage.rb#L30-L36

lib/3scale/backend/storage_async/client.rb
jlledom: No problem 👍

docs/configuration.md
jlledom: I'm gonna remove it.

openshift/3scale_backend.conf
jlledom: Initially, I thought it could be removed, I even created a PR for porta: https://github.com/3scale/porta/pull/3808. However, there's an scenario where it's required: when using sentinels, and providing the name through the `:name` param. In that scenario the `:url` is not needed so there's no other way to set `ssl: true` rather than the `:ssl` param.

For the porta PR what I decided is to take the ssl value from the URL but only in case it's not provided as a parameter. I think in general is good to have the two ways to set the values: first by param, if not present, then from the URL. That's how it works for `:name` and `:db`.

What I could do is to implement the same behavior for `:ssl` as I did in that porta PR.

test/test_helpers/certificates.rb
akostadinov: I think it is safe to support only non-deprecated keys. We can add if anybody requires it. It is deprecated for a reason and supporting it can only require more cumbersome process to remove later.

jlledom: But how we detect the key type and size? I don't think it's straightforward, it would probably require it's own PR.

My idea was to wait for a newer Ruby version where DSA is not supported, so we stop supporting it in a future Ruby update.

docs/configuration.md
akostadinov: maybe 
```suggestion
- Optional. Defaults to not set, in which case URI schema will be used to decide.
```

jlledom: That's fine, but it's currently not true for apisonator, I can fix it, it shouldn't be too hard.

akostadinov: If possible, I'd suggest remove this setting and use schema. The less configuration, the better.

docs/configuration.md
akostadinov: outrageous!

test/test_helpers/certificates.rb
akostadinov: ok, this is for testing, I first didn't realize, in production code we don't deal with it

openshift/3scale_backend.conf
akostadinov: Duplicating the configuration is a double-edge sword as we need to resolve situations of conflict.

For example, with the above implementation, if user has configured `ssl` to `false` but used `rediss` schema, then schema will take precedence. If `ssl` is set to `true` but schema is `redis`, then `ssl` will take precedence.

I'm not saying this behavior is bad. Actually it is a safe behavior because TLS connection type will take precedence if it is configures on either place. It should be well documented though, to avoid confusion.

I still would personally opt to require `url` to be configured in case TLS connection is desired and avoid double config. But if you prefer having both - that's also fine with me, with configuration option descriptions being made more thorough.

lib/3scale/backend/configuration.rb
akostadinov: Got it, thanks

docs/configuration.md
akostadinov: Ok, let's discuss there.

test/test_helpers/certificates.rb
```diff
@@ -0,0 +1,67 @@
...
+      ef.issuer_certificate = cert
+      cert.extensions = [
+        ef.create_extension("basicConstraints","CA:TRUE", true),
+        ef.create_extension("subjectKeyIdentifier", "hash"),
+      # ef.create_extension("keyUsage", "cRLSign,keyCertSign", true),
+      ]
+      cert.add_extension ef.create_extension("authorityKeyIdentifier",
+                                             "keyid:always,issuer:always")
+
+      cert.sign key, OpenSSL::Digest::SHA1.new
```
akostadinov, line 55: This is test cert and not security sensitive but lets be cool and use a safe hash like SHA512. SHA1 is deemed insecure already and at some point new servers may start rejecting it by default. Just a nit pick.

lib/3scale/backend.rb
akostadinov: hmm, I can't understand why this shows changed, but anyway, so be it if you want it like that.

docs/configuration.md
akostadinov: :thinking:

test/unit/storage_async_test.rb
```diff
@@ -182,20 +190,98 @@ def test_sentinels_empty
...
     url = URI(conf[:url])
     host, port = client.endpoint.address
     assert_equal url.host, host
     assert_equal url.port, port
+
+    unless conf[:username].to_s.strip.empty? && conf[:password].to_s.strip.empty?
+      assert_instance_of ThreeScale::Backend::AsyncRedis::Protocol::ExtendedRESP2, client.protocol
+      username, password = client.protocol.instance_variable_get(:@credentials)
+      assert_equal conf[:username], username
+      assert_equal conf[:password], password
```
akostadinov, line 268: I notice no tests for sentinels credentials. Maybe add one?

openshift/3scale_backend.conf
jlledom: I prefer to have both. I'll document the precedence.

test/test_helpers/certificates.rb
jlledom: I'll do it

test/unit/storage_async_test.rb
```diff
@@ -182,20 +190,98 @@ def test_sentinels_empty
...
+      key = create_key alg
+      cert = create_cert key
+      FakeFS.with_fresh do
+        ca_file_path = File.expand_path(File.join(__FILE__, '..', '..', '..', 'script', 'config', 'ca-root-cert.pem'))
+        key_path = File.expand_path(File.join(__FILE__, '..', '..', '..', 'script', 'config', 'redis-dsa.pem'))
+        cert_path = File.expand_path(File.join(__FILE__, '..', '..', '..', 'script', 'config', 'redis-dsa.crt'))
+        FileUtils.mkdir_p(File.dirname(ca_file_path))
+        File.write(ca_file_path, ca_file.to_pem)
+        File.write(key_path, key.to_pem)
+        File.write(cert_path, cert.to_pem)
```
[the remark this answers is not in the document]
jlledom: Yes, better remove a dependency

test/unit/storage_sync_test.rb
```diff
@@ -244,6 +244,53 @@ def test_sentinels_empty
     end
   end
 
+  def test_tls_no_client_certificate
```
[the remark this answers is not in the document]
jlledom: I'll add some tests

test/unit/storage_async_test.rb
jlledom: Done

jlledom:
@akostadinov I think I addressed all your comments.

test/unit/storage_async_test.rb
```diff
@@ -203,28 +203,29 @@ def test_tls_no_client_certificate
 
   [:rsa, :dsa, :ec].each do |alg|
     define_method "test_tls_client_cert_#{alg}" do
-      ca_file = create_cert
+      ca_cert_file = Tempfile.new('ca-root-cert.pem')
+      ca_cert_file.write(create_cert.to_pem)
+      ca_cert_file.rewind
```
akostadinov: I'm not sure about your intention. Probably you mean `flush` and then `close` and `unlink` in an `ensure` block?

test/unit/storage_async_test.rb
```diff
@@ -302,10 +338,42 @@ def assert_sentinel_config(conf, conn)
...
+      username, password = client.protocol.instance_variable_get(:@credentials)
+      assert_equal conf[:username], username
+      assert_equal conf[:password], password
+    end
+
+    unless conf[:sentinel_username].to_s.strip.empty? && conf[:sentinel_password].to_s.strip.empty?
+      sentinel_username, sentinel_password = client.instance_variable_get(:@sentinel_credentials)
+      assert_equal conf[:sentinel_username], sentinel_username
+      assert_equal conf[:sentinel_password], sentinel_password
+    end
```
akostadinov, line 356: should we test that `sentinel_username` is `nil` when `conf[:sentinel_username].to_s.strip.empty?`? Now we only do positive testing as far as I can tell.

test/unit/storage_async_test.rb
jlledom: I took that from your link (https://discuss.rubyonrails.org/t/validates-immutable-true/81479/2). I assumed `#rewind` was necessary to ensure the file is read from the beginning.

About unlinking, I'll add an ensure block.

akostadinov: If you use the same file handle - yes. But you pass the files as paths. So they are opened again.

jlledom: Oh, true. I'll fix this.

test/unit/storage_async_test.rb
jlledom: Done in https://github.com/3scale/apisonator/pull/391/commits/fef9bcd273ebab868841b3900c811ea9a9ba5699

test/unit/storage_async_test.rb
```diff
@@ -302,10 +338,42 @@ def assert_sentinel_config(conf, conn)
...
+  end
+
+  def assert_tls_certs(conf, client, test_cert_type)
+    unless conf[:ssl_params].to_s.strip.empty?
```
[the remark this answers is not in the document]
jlledom: That configuration would make no sense, I think it's fine the connection fails in that situation.

test/unit/storage_async_test.rb
jlledom: https://github.com/3scale/apisonator/pull/391/commits/c27f6c43a46bc21667d3c5830a6e5e764c0a610f

test/unit/storage_async_test.rb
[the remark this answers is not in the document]
jlledom: oh wait, that situation does make sense indeed. If they set `ssl: true` but provide no certs, then the certs from the system should be used by default. That would surely work in sync mode, but I'm not sure it's implemented in async mode. Let me take a look.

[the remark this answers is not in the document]
jlledom: I've been investigating and I don't know how could I test this scenario. I would need to have a cert signed by one of the well-known CAs which certificates are installed in fedora, and also the private key corresponding to that cert. Then configure my local redis server to use that certificate and try to connect from apisonator with `ssl: true` but no `ssl_params`.

I can't test it myself but my guess is it would work fine, also in async mode, because we rely on `Async::IO::SSLEndpoint.new`.

test/unit/storage_async_test.rb
```diff
@@ -286,23 +290,28 @@ def test_acl_tls
...
   end
 
   private
 
   def create_certs(alg)
     ca_cert_file = Tempfile.new('ca-root-cert.pem')
     ca_cert_file.write(create_cert.to_pem)
-    ca_cert_file.rewind
+    ca_cert_file.flush
+    ca_cert_file.close
```
akostadinov, line 303: I'm not sure if closing will trigger auto-deleting the file. Just not sure. Probably `flush` is enough. But if it works, that's fine with me.

test/unit/storage_async_test.rb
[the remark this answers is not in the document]
akostadinov: well, we just test that the configuration is as expected I guess.

akostadinov (approved):
Looks great, thank you!

test/unit/storage_async_test.rb
jlledom: I also thought that, but I would say the file is not destroyed until the variable is garbage-collected, and since we are returning the variables to the calling method, they should stay. The tests pass.

test/unit/storage_async_test.rb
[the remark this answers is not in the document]
akostadinov: btw if we really want to test scerario where OS CA should be used, we can setup /etc/pki with fakefs with custom certificates. But I think we don't need to test ruby implementation.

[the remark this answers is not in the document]
jlledom: > btw if we really want to test scerario where OS CA should be used, we can setup /etc/pki with fakefs with custom certificates.

Yeah, that's right.

> But I think we don't need to test ruby implementation.

Yeah, that's right.

PR #393, opened 2024-06-18
Remove empty credentials

jlledom:
ACL Support works fine when both `user` and `password` are provided. But for backwards compatibility we need to properly send the credentials when the `user` is not specified.

In this scenario, our protocol needs to send the next command:

```
AUTH <password>
```

Check the [docs](https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl/).

This PR just adds a small fix to remove empty elements from the `credentials` array provided to the protocol.


akostadinov (approved):
Makes sense, thank you. Would be nice to have a test though maybe in a future PR.

PR #394, opened 2024-06-20
THREESCALE-10688: Enable `Content-Security-Policy`

jlledom:
**What this PR does / why we need it**:

Running Rapidast against 3scale reveals the next warning about apisonator:

```
Content Security Policy (CSP) Header Not Set
```

In order to fix the warning, we must return the CSP header in all requests to the listener

**Which issue(s) this PR fixes** 

https://issues.redhat.com/browse/THREESCALE-10688

**Verification steps** 

Curl the listener with the appropriate parameters:

```
curl -v 'http://localhost:3001/transactions/authrep.xml?service_token=<TOKEN>&service_id=<SERVICE_ID>&user_key=<USER_KEY>&usage%5Bhits%5D=1'
```

You should see the CSP header in the response:

```
content-security-policy: default-src 'self'
```

**Special notes for your reviewer**:

Some useful links about CSP:
- https://content-security-policy.com/
- https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources


lib/3scale/backend/csp.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      CSP_VALUES = "default-src 'self'".freeze
+      private_constant :CSP_VALUES
+
+      CSP_HEADERS = {
+        'Content-Security-Policy'.freeze => CSP_VALUES
+      }.freeze
+      private_constant :CSP_HEADERS
+
+      stringify_consts :CSP_VALUES, :CSP_HEADERS
```
akostadinov, line 24: what is this stringify supposed to do?

akostadinov (approved):
Looks nice, just question about this stringify thing.

lib/3scale/backend/csp.rb
jlledom: I don't know, I took it from https://github.com/3scale/apisonator/blob/4f29aabd1dbd71c9eb06403614080aa42feb3d01/lib/3scale/backend/cors.rb#L53

I assumed it was needed for something, but I could be wrong.

akostadinov: I read it as if it creates a constant `CONST_NAME_S` with stringified value. I see no reason to do that because we use the original constant only as far as I can see.

jlledom: Yeah, no idea why is this here. Probably dead code

PR #396, opened 2024-07-19
Upgrade to ruby 3.3

akostadinov:


lib/3scale/dotenv.rb
```diff
@@ -5,7 +5,7 @@
 # If doesn't exist, try to load .env.test
 def env_file
   file = ".env.#{ENV['RACK_ENV']}"
-  File.exists?(file) ? file : '.env.test'
+  File.exist?(file) ? file : '.env.test'
```
akostadinov, line 8: this seems to break in Ruby 3.3

test/unit/storage_async_test.rb
```diff
@@ -251,7 +251,7 @@ def test_tls_no_client_certificate
     config_obj = {
       url: 'rediss://localhost:46379',
       ssl_params: {
-        ca_file: File.expand_path(File.join(__FILE__, '..', '..', '..', 'script', 'config', 'ca-root-cert.pem'))
+        ca_file: Tempfile.new('ca-root-cert.pem').tap { _1.write(create_cert.to_pem) && (_1.flush ; _1.close) }.to_path
```
akostadinov, line 254: apparently earlier OpenSSL silently ignored missing CA files. Newer OpenSSL fails. I mean the C library.

Gemfile.lock
```diff
@@ -81,7 +81,7 @@ GEM
     daemons (1.2.4)
     diff-lcs (1.5.1)
     docile (1.1.5)
-    dotenv (2.8.1)
+    dotenv (3.1.2)
```
akostadinov, line 84: initially I thought `dotenv` was the issue. Later it turned out not to be but I already had it updated. I see no harm in keeping it that way.

test/unit/storage_async_test.rb
akostadinov: Added a new commit, should be good now.

test/unit/storage_async_test.rb
```diff
@@ -308,11 +308,16 @@ def test_acl_tls
 
   private
 
+  def create_ca(alg)
```
jlledom, line 311: Two things:

1. This method fits better under `TestHelpers::Certificates` where `create_key` and `create_cert` are
2. The parameter `alg` is unused

akostadinov: I think so but I don't want to figure out now where we have non-dried methods to create cert files. Also if this has to be correct, the `#create_cert` should sign with the CA. Now CA and certs have nothing to do with each other. And both are marked as CAs... I'm just extracting the logic as a first step, then we can move to helpers once apisonator is working. Anyway, will push a change to at least make `alg` used.

PR #403, opened 2024-08-12
THREESCALE-11232: Add benchmark to pipeline

jlledom:
In the past, small changes like updating a library to the next minor release had a significant impact in performance. Also, small fixes in the worker caused a memory leak.

I'd like to have a way to detect this automatically in the pipeline, so we don't have to do the tests manually on our local boxes.

I added a new job to the pipeline that runs a new benchmark which tests performance and memory usage. Unfortunately, it never fails whatever the result, so we need to open the pipeline and check the results ourselves.

I'd like to have a way to make it fail if a memory leak or a performance degradation is detected, but how? I had two ideas, and I don't like any of them:

1. Just declare some expected values for time and memory, as constants, and fail if the benchmark result exceeds that value by a say 10%.
2. Same as before, but the expected values are stored in the cache from previous runs.

Any ideas?


bench/worker/worker_ci_bench.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'base'
+
+class WorkerCIBenchmark < Base
```
mayorova, line 3: What is the purpose of splitting the two?
They seem identical, just that the non-CI also runs with 1M reports?

I think we can either parameterize it, or enable/disable 1M for both modes.

Or, if you want to keep both, maybe refactor the `run` method to reuse most of the code.

bench/worker/base.rb
```diff
@@ -0,0 +1,56 @@
...
+  def create_reports(num)
+    warn "generating #{num} transactions.."
+    num.times { default_report }
+  end
+
+  def new_worker(fetcher = nil)
+    Worker.new(async: async?, job_fetcher: fetcher)
+  end
+
+  def clear_queue
```
mayorova, line 25: I would also suggest renaming this method to `process_queue` or something. Because now it looks as if it just... cleared it :upside_down_face: 
At first I didn't really get why are we benchmarking the cleaning operation.

mayorova:
> I'd like to have a way to make it fail if a memory leak or a performance degradation is detected, but how? I had two ideas, and I don't like any of them:
> 
> 1. Just declare some expected values for time and memory, as constants, and fail if the benchmark result exceeds that value by a say 10%.

That doesn't feel right... but if we don't find a better way, that's fine, I guess.

> 2. Same as before, but the expected values are stored in the cache from previous runs.

Hm, I am not sure how it is possible in CircleCI. I've looked through the docs quickly, and not sure if cache is the right way - I think it's not designed for it.

It has a possibility for storing artifacts, but I am not sure if they can be accessed from different jobs/workflows. And even in this case - we'd probably need some way to parse the results and compare them (but I guess that's doable).
 
> Any ideas?

To begin with, maybe we can change the behavior of the benchmark so that it stores the results in some structured file (looks like CircleCI likes XML... but maybe JSON could be fine? not sure), instead of just printing to the standard output.

This way we could at least probably visualize the results in the CircleCI UI.

And if we find a way to compare the results between runs, I think it would also be easier to use structured data file than parse the text output.

Just a suggestion....

bench/worker/worker_ci_bench.rb
jlledom: Done in https://github.com/3scale/apisonator/pull/403/commits/63b5c2d17b32271c7f638abc5d80736ce2baa561

bench/worker/worker_ci_bench.rb
```diff
@@ -0,0 +1,19 @@
...
+    warn "=" * 70
+    warn Benchmark::CAPTION
+    warn res.format(Benchmark::Tms::FORMAT).chop + " (#{res.label})"
+    warn "=" * 70
+    warn "Memory usage: #{rss} KB"
+    warn "=" * 70
+  end
+end
+
+WorkerCIBenchmark.new.run
```
akostadinov, line 19:
```suggestion
if __FILE__ == $PROGRAM_NAME
  WorkerCIBenchmark.new.run
end
```

bench/worker/worker_ci_bench.rb
akostadinov: To me this CI class also looks not very useful. Maybe use an env-variable for the number. But whatever.

bench/worker/worker_ci_bench.rb
jlledom: If I do this the rake task doesn't run the benchmark

bench/worker/worker_ci_bench.rb
jlledom: > To me this CI class also looks not very useful.

Have you checked https://github.com/3scale/apisonator/commit/63b5c2d17b32271c7f638abc5d80736ce2baa561 ? `WorkerBenchmark` and `WorkerCIBenchmark` do different things, one runs two benchmarks and the other one runs one benchmark + gets memory usage. Any ideas?

> Maybe use an env-variable for the number. But whatever.

For which number exactly? I understand this as reusing `WorkerBenchmark` but getting the amount of reports from an env var. But `WorkerBenchmark` runs two benchmarks with two amounts.

akostadinov:
wrt reporting, you can just generated a junit file, it is very simple xml. Copy an existing one end edit the strings.

bench/worker/worker_ci_bench.rb
akostadinov: I don't think we need to stick to this. I'm fine running just ones either way.

bench/worker/worker_ci_bench.rb
akostadinov: I don't like it but ok, now I remember not liking it before. Maybe refactor some day.

jlledom: What do you mean? you don't like running it as a rake task?

jlledom:
> wrt reporting, you can just generated a junit file, it is very simple xml. Copy an existing one end edit the strings.

I've been trying this. I'm not happy with the result so I removed the changes. Essentially, you can show some info in the `Tests` tab in CircleCI by generating a simple xml file. However, only failed tests are show in the tab, other wise you only see a "Great job" empty screen ([example](https://app.circleci.com/pipelines/github/3scale/apisonator/1258/workflows/46053a34-c5d6-43c0-b6e8-6ba61df47e34/jobs/2561/tests)). So in order to show the date here I should create an XML file simulating some failed tests, and I don't like the idea. An again, the benefits are not a big deal because the benchmark results can be found on the test logs in the `Steps` tab ([example](https://app.circleci.com/pipelines/github/3scale/apisonator/1258/workflows/46053a34-c5d6-43c0-b6e8-6ba61df47e34/jobs/2561/steps)).

jlledom:
@akostadinov @mayorova I simplified this. I removed the CI benchmark and now the CI launches the existing worker benchmark with a few changes. Also a new env variable `NUM_REPORTS` can be used to provide the number of transactions to run the benchmark.

Still there's no way to make the benchmark detect performance degradation or memory leaks, and fail. I haven't found any acceptable way to do so. But at least we can check the benchmark results easily with two clicks from the PR:

image

bench/worker/worker_bench.rb
```diff
@@ -7,6 +7,7 @@ class WorkerBenchmark
   def initialize
     super
     @async = ThreeScale::Backend.configuration.redis.async
+    @nreports = (ENV.fetch('NUM_REPORTS', nil) || 100_000).to_i
```
akostadinov, line 10:
```suggestion
    @nreports = ENV.fetch('NUM_REPORTS', "100000").to_i
```

akostadinov (approved):
Looks great. I'm thinking though that the multirun has a meaning. To probably catch memory leaks. So maybe good idea to have the variable be a comma separated list of iterations, for example might be `NUM_REPORTS=10000,100000,1000000` to have 3 runs of whatever numbers.

But that can be for some another time in the dark future.

akostadinov (approved):

jlledom:
> Looks great. I'm thinking though that the multirun has a meaning. To probably catch memory leaks. So maybe good idea to have the variable be a comma separated list of iterations, for example might be `NUM_REPORTS=10000,100000,1000000` to have 3 runs of whatever numbers.

We are already running 10k for the warm up and 100K as the main test. I'm not sure about processing 1M transactions in the pipeline because it takes too much time, I think it was like 6 minutes and it has to be run twice, for sync and async modes, so at least 12 minutes to complete the benchmark. Unlike porta, here we have a fast pipeline which completes in less than two minutes, and that's fantastic.

jlledom:
> The values are examples. It might be `NUM_REPORTS=10000,10000,10000` to see if memory changed between 3 identical runs.

I implemented this at https://github.com/3scale/apisonator/pull/403/commits/c66cf8c9265fec7ac16ee90f0c97dd775f704134

It looks like this:

```
$ NUM_REPORTS=50000,100000,1000000 bundle exec rake bench[worker/worker_bench.rb]
Running benchmark /home/jlledom/redhat/apisonator/bench/worker/worker_bench.rb
generating 10000 transactions..
Queue has 12000 jobs..
Shutting down worker..
generating 50000 transactions..
Queue has 60000 jobs..
Shutting down worker..
======================================================================
      user     system      total        real
 14.028837   1.787082  15.815919 ( 19.241702) (50000 reports)
======================================================================
Memory usage: 48240 KB
======================================================================
generating 100000 transactions..
Queue has 120000 jobs..
Shutting down worker..
======================================================================
      user     system      total        real
 27.002690   3.413024  30.415714 ( 35.961862) (100000 reports)
======================================================================
Memory usage: 48240 KB
======================================================================
generating 1000000 transactions..
Queue has 1200001 jobs..
Shutting down worker..
======================================================================
      user     system      total        real
270.930916  34.218637 305.149553 (351.948888) (1000000 reports)
======================================================================
Memory usage: 48752 KB
======================================================================
Benchmarks finished
```

However, in the pipeline we run the default: `10000` to warm up and `100000` to benchmark.

bench/worker/worker_bench.rb
```diff
@@ -55,6 +56,15 @@ def dup_resque_redis
   end
 
   def run
+    n_reports = [*@nreports.strip.split(',')].map(&:to_i)
```
akostadinov, line 59: I'm not sure why do we need `[*...]`. Also perhaps more robust to strip each value in case user enters some spaces in between, e.g. `100, 1000 , 10000`
```suggestion
    n_reports = @nreports.split(',').map(&:strip).map(&:to_i)
```

bench/worker/worker_bench.rb
```diff
@@ -55,6 +56,15 @@ def dup_resque_redis
...
+
+    # Warming up...
+    run_benchark(10000, print_report: false)
+
+    n_reports.each { run_benchark(_1) }
```
akostadinov, line 64: this is fine, just a nitpick, perhaps output has multiple times `    "      user     system      total        real\n"` (this is `CAPTION`) instead of the original table. But it is a minor thing so I don't care too much. Just saying.

jlledom: Yes, it prints a different table for each batch. Is that what you don't like?

I can refactor it to show only one table with all results.

akostadinov: It makes sense to be a single table but as I said, not super important.

akostadinov (approved):

bench/worker/worker_bench.rb
jlledom: Done in https://github.com/3scale/apisonator/pull/403/commits/2a7018873a3d54cde6565194bf38ad2b982f827e

Now it looks like this:

```
$ NUM_REPORTS=1000,1000,1000 bundle exec rake bench[worker/worker_bench.rb]
Running benchmark /home/jlledom/redhat/apisonator/bench/worker/worker_bench.rb
generating 10000 transactions..
Queue has 12000 jobs..
Shutting down worker..
generating 1000 transactions..
Queue has 1200 jobs..
Shutting down worker..
generating 1000 transactions..
Queue has 1200 jobs..
Shutting down worker..
generating 1000 transactions..
Queue has 1200 jobs..
Shutting down worker..
======================================================================
      user     system      total        real
  0.508081   0.067297   0.575378 (  1.776290) 48580KB (1000 reports)
  0.444679   0.070898   0.515577 (  1.707208) 48836KB (1000 reports)
  0.463991   0.059394   0.523385 (  1.651240) 48836KB (1000 reports)
======================================================================
Benchmarks finished
```

PR #405, opened 2024-09-05
THREESCALE-9682: Add support for OpenTelemetry

jlledom:
This adds OpenTelemetry instrumentation for Sinatra in Apisonator.

**How to verify:**

1. Follow the instructions on the [docs](https://opentelemetry.io/docs/languages/ruby/exporters/) to launch a Jaeger container.
2. Set `OTEL_EXPORTER_OTLP_ENDPOINT="http://localhost:4318"` and the new var  `CONFIG_OPENTELEMETRY_ENABLED=true`.
3. Work normally, e.g. send some transaction requests to the listener.
4. Visit Jaeger UI at http://localhost:16686 and verify the requests are reported.
5. Verify that not setting `CONFIG_OPENTELEMETRY_ENABLED` or setting it to `false` disables the metrics and you can't see the requests in Jaeger.

**Jira Issue:**
https://issues.redhat.com/browse/THREESCALE-9682


akostadinov:
Is it possible by any chance to test that the endpoint is operational with the said env variable? Maybe not, just asking.

jlledom:
> Is it possible by any chance to test that the endpoint is operational with the said env variable? Maybe not, just asking.

I don't think there's a way other than pinging the endpoint host and port directly.

akostadinov:
Fair enough. But how do we know whether this fixes the actual JIRA issue THREESCALE-9669?

jlledom:
> Fair enough. But how do we know whether this fixes the actual JIRA issue THREESCALE-9669?

The issue is not only about apisonator. As I understood it, the only apisonator work required is:

> 3scale Backend should be instrumented using opentelemetry SDK. Sinatra has instrumentation in place implemented in https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/sinatra

So there's no need to check any endpoint AFAIK. @eguzki ?

akostadinov (approved):

eguzki:
That's the one jira issue describing the work being done here: https://issues.redhat.com/browse/THREESCALE-9682

lib/3scale/backend/opentelemetry.rb
```diff
@@ -0,0 +1,8 @@
+require 'opentelemetry/sdk'
+require 'opentelemetry/instrumentation/sinatra'
+require 'opentelemetry-exporter-otlp'
+
+OpenTelemetry::SDK.configure do |c|
+  c.service_name = 'backend'
```
eguzki, line 6:
```suggestion
  c.service_name = '3scale-backend'
```

(if service names starting with a number are allowed)

eguzki:
It would be nice to drop an snapshot from jaeger UI with the trace being recorded from backend listener

lib/3scale/backend.rb
```diff
@@ -53,6 +53,7 @@
 require '3scale/backend/failed_jobs_scheduler'
 require '3scale/backend/transactor'
 require '3scale/backend/listener'
+require '3scale/backend/opentelemetry'
```
eguzki, line 56: This is only executed from the backend listener?

jlledom: I'd say it only makes sense for the listener, which is who listens to sinatra routes

eguzki: the worker is also importing this file. 

What about `lib/3scale/backend/server.rb` ?? I did not try myself, though. Looks like `lib/3scale/backend/listener.rb` is being imported by both worker and listener :shrug: 

Not important, though.. up to you.. not a request for change. The worker will not have the opt-in switch enabled, so it is harmless.

jlledom: Fixed: https://github.com/3scale/apisonator/pull/405/commits/a45900b1f8e9c19675a0996e8a0e6cfeee34a421

eguzki (approved):

PR #407, opened 2024-10-16
THREESCALE-10826: Add support for unix sockets in async mode

jlledom:
Connection to Redis through unix sockets is not supported in async mode. This PR adds the logic to implement it. Basically, before the PR the code only followed one path to connect using a TCP socket, now the code is split into two paths, one for TCP and one for Unix.

This is part of https://issues.redhat.com/browse/THREESCALE-10826

**How to verify**

1. Launch a redis container listening in the unix socket:
```
podman run --name redis-unix -v /path/to/your/run:/var/run/:z redis:6.2-alpine redis-server --unixsocket /var/run/redis.sock --unixsocketperm 777 --port 0
```
2. Set `CONFIG_QUEUES_MASTER_NAME` and `CONFIG_REDIS_PROXY` to `unix:///path/to/your/run/redis.sock`
3. Both listener and worker should work as usual

Additionally, ACL credentials can be provided through the corresponding environment variables, when the server is authenticated. Sentinels and TLS connections cannot be established through unix sockets, though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,7 @@ commands:
             cp -r script/config/examples/tls/sentinel1 ./sentinel1_tls
             cp -r script/config/examples/tls/sentinel2 ./sentinel2_tls
             cp -r script/config/examples/tls/sentinel3 ./sentinel3_tls
+            mkdir -m 1777 ./run
       - run:
           name: Start services
           command: |
@@ -119,7 +120,13 @@ workflows:
       - test_single:
           requires:
             - install_dependencies
+      - test_single_unix:
+          requires:
+            - install_dependencies
       - test_sentinels:
+          requires:
+            - install_dependencies
+      - test_acl_unix:
           requires:
             - install_dependencies
       - test_tls:
@@ -161,6 +168,18 @@ jobs:
       - start_services:
           services: redis-master
       - run_tests
+  test_single_unix:
+    executor:
+      name: ubuntu_vm
+    working_directory: ~/project
+    environment:
+      CONFIG_QUEUES_MASTER_NAME: unix://run/redis.sock
+      CONFIG_REDIS_PROXY: unix://run/redis.sock
+    steps:
+      - *attach-to-workspace
+      - start_services:
+          services: redis-master
+      - run_tests
   test_sentinels:
     executor:
       name: ubuntu_vm
@@ -174,6 +193,22 @@ jobs:
       - *attach-to-workspace
       - start_services:
           services: redis-master redis-replica1 redis-replica2 redis-sentinel1 redis-sentinel2 redis-sentinel3
+      - run_tests
+  test_acl_unix:
+    executor:
+      name: ubuntu_vm
+    working_directory: ~/project
+    environment:
+      CONFIG_QUEUES_MASTER_NAME: unix://run/redis.sock
+      CONFIG_QUEUES_USERNAME: porta
+      CONFIG_QUEUES_PASSWORD: sup3rS3cre1!
+      CONFIG_REDIS_PROXY: unix://run/redis.sock
+      CONFIG_REDIS_USERNAME: porta
+      CONFIG_REDIS_PASSWORD: sup3rS3cre1!
+    steps:
+      - *attach-to-workspace
+      - start_services:
+          services: tls-redis-master
       - run_tests
   test_tls:
     executor:
--- a/lib/3scale/backend/async_redis/client.rb
+++ b/lib/3scale/backend/async_redis/client.rb
@@ -16,6 +16,19 @@ module ThreeScale
           # @param opts [Hash] Redis connection options
           # @return [Async::Redis::Client]
           def call(opts)
+            return connect_tcp(opts) if url_present?(opts[:url])
+
+            connect_unix(opts)
+          end
+          alias :connect :call
+
+          private
+
+          def url_present?(url)
+            !url.to_s.strip.empty?
+          end
+
+          def connect_tcp(opts)
             uri = URI(opts[:url])
 
             credentials = [ uri.user || opts[:username], uri.password || opts[:password]]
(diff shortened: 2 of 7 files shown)


.circleci/config.yml
```diff
@@ -49,6 +49,7 @@ commands:
             cp -r script/config/examples/tls/sentinel1 ./sentinel1_tls
             cp -r script/config/examples/tls/sentinel2 ./sentinel2_tls
             cp -r script/config/examples/tls/sentinel3 ./sentinel3_tls
+            mkdir -m 777 ./run
```
akostadinov, line 52: Why is this 777?

.circleci/config.yml
```diff
@@ -161,6 +168,18 @@ jobs:
       - start_services:
           services: redis-master
       - run_tests
+  test_single_unix:
+    executor:
+      name: ubuntu_vm
+    working_directory: ~/project
+    environment:
+      CONFIG_QUEUES_MASTER_NAME: unix://./run/redis.sock
+      CONFIG_REDIS_PROXY: unix://./run/redis.sock
```
akostadinov, line 177:
```suggestion
      CONFIG_QUEUES_MASTER_NAME: unix://run/redis.sock
      CONFIG_REDIS_PROXY: unix://run/redis.sock
```

.circleci/config.yml
```diff
@@ -175,6 +194,22 @@ jobs:
       - start_services:
           services: redis-master redis-replica1 redis-replica2 redis-sentinel1 redis-sentinel2 redis-sentinel3
       - run_tests
+  test_acl_unix:
+    executor:
+      name: ubuntu_vm
+    working_directory: ~/project
+    environment:
+      CONFIG_QUEUES_MASTER_NAME: unix://./run/redis.sock
```
akostadinov, line 202:
```suggestion
      CONFIG_QUEUES_MASTER_NAME: unix://run/redis.sock
```

.circleci/config.yml
```diff
@@ -175,6 +194,22 @@ jobs:
...
+      CONFIG_QUEUES_USERNAME: porta
+      CONFIG_QUEUES_PASSWORD: sup3rS3cre1!
+      CONFIG_REDIS_PROXY: unix://./run/redis.sock
```
akostadinov, line 205:
```suggestion
      CONFIG_REDIS_PROXY: unix://run/redis.sock
```

.circleci/config.yml
akostadinov: Not that it matters much but maybe better
```suggestion
            mkdir -m 1777 ./run
```

script/config/podman-compose.yml
```diff
@@ -47,6 +49,7 @@ services:
     container_name: tls-redis-master
     network_mode: host
     volumes:
+      - ./run:/var/run:z
```
akostadinov, line 52: relabeling the whole `/var/run` doesn't seem right. Maybe it doesn't matter much but better create a subdir for the socket. e.g.
```suggestion
      - ./run:/var/run/redis:z
```

lib/3scale/backend/async_redis/endpoint_helpers.rb
```diff
@@ -14,20 +15,40 @@ class << self
...
           # @param port [Integer]
+          # @param path [String]
+          # @param ssl [Boolean]
           # @param ssl_params [Hash]
           # @return [Async::IO::Endpoint::Generic]
-          def prepare_endpoint(host, port, ssl = false, ssl_params = nil)
+          def prepare_endpoint(**kwargs)
+            return prepare_tcp_endpoint(**kwargs) if host_present?(kwargs[:host])
+
+            prepare_unix_endpoint(**kwargs)
```
akostadinov, line 25: IMO would read nicer like:
```suggestion
            host_present?(kwargs[:host]) ? prepare_tcp_endpoint(**kwargs) : prepare_unix_endpoint(**kwargs)
```

lib/3scale/backend/async_redis/endpoint_helpers.rb
```diff
@@ -14,20 +15,40 @@ class << self
...
 
             tcp_endpoint
           end
 
+          def prepare_unix_endpoint(path: '', ssl: false, ssl_params: nil)
+            unix_endpoint = Async::IO::Endpoint.unix(path, Socket::PF_UNIX)
+
+            return prepare_ssl_endpoint(endpoint: unix_endpoint, ssl_params: ssl_params) if ssl
+
+            unix_endpoint
```
akostadinov, line 43: A little bit effort to understand. Easy to miss this `if ssl` at the end of the long line. I'd suggest one of those.
```suggestion
            unix_endpoint = Async::IO::Endpoint.unix(path, Socket::PF_UNIX)
            return unix_endpoint unless ssl

            prepare_ssl_endpoint(endpoint: unix_endpoint, ssl_params: ssl_params)
```

or

```suggestion
            unix_endpoint = Async::IO::Endpoint.unix(path, Socket::PF_UNIX)

            ssl ? prepare_ssl_endpoint(endpoint: unix_endpoint, ssl_params: ssl_params) : unix_endpoint
```

lib/3scale/backend/async_redis/endpoint_helpers.rb
```diff
@@ -39,6 +60,10 @@ def format_ssl_params(ssl_params)
 
             updated_ssl_params
           end
+
+          def host_present?(host)
+            not host.to_s.strip.empty?
```
akostadinov, line 65: not? :grimacing: I think in our code base we  have always used `!`
```suggestion
            !host.to_s.strip.empty?
```

akostadinov: A more rails approach and more readable would be
```suggestion
            host.present?
```

akostadinov: And actually no need for a separate method for this, you can just inline wherever :)

jlledom:
@akostadinov I applied your suggestions

lib/3scale/backend/async_redis/client.rb
```diff
@@ -16,6 +16,19 @@ class << self
           # @param opts [Hash] Redis connection options
           # @return [Async::Redis::Client]
           def call(opts)
+            return connect_tcp(opts) if url_present?(opts[:url])
+
+            connect_unix(opts)
```
akostadinov, line 21: In this case do we support `unix://` URLs and how?

jlledom: `unix://` URLs are supported and inf fact they are the only way to setup a unix path on apisonator. So it matches how the porta side works

akostadinov: But then URL will be present and `connect_tcp` will be used instead of `connect_unix`? Maybe I'm missing something obvious in the logic..

jlledom: The `:url` param is removed here:

https://github.com/3scale/apisonator/blob/e79bef023dc250e7a916a4afe93d6fa0c34599e6/lib/3scale/backend/storage_helpers.rb#L280-L282

This is because `redis-rb` expects `:url` for TCP and `:path `for unix sockets, so we must emulate the same behavior for `async-redis`

akostadinov (approved):

PR #411, opened 2024-12-11
THREESCALE-11458: Bump Sinatra to 4.1.1

jlledom:
This is to properly implement https://github.com/3scale/apisonator/pull/410

Jira: https://issues.redhat.com/browse/THREESCALE-11458

Due to a dependency hell, updating `sinatra` forces to update:

- `rack/rackup`
- `falcon`
- `puma`

I also had to fork `rspec_api_documentation` because the gem it's abandoned and it's not compatible with Rack 3.

I also updated `hiredis-client` which I don't think it's necessary, but https://github.com/3scale/apisonator/pull/410 does it for some reason so I'll do it as well, just in case I'm missing something.

I did some testing locally and seems to work fine, also the test suite in CircleCI passes for all environments, and the benchmark results doesn't show ant impact in performance.


config/puma.rb
```diff
@@ -145,7 +145,7 @@
 # is preserved across a phased-restart. (incompatible with preload_app)
 # (off by default)
 
-prune_bundler
+# prune_bundler
```
jlledom, line 148: `prune_bundler` is required to phase restart puma. However, it's only enabled on `test` and `development` envs, where we never phase restart.

It's causing us a lot of problems to run our test suite in Puma 6, so I'm disabling it.

lib/3scale/backend/listener.rb
```diff
@@ -411,8 +406,7 @@ def do_api_method(method_name)
 
         raise_provider_key_error(params) if blank?(provider_key)
 
-        # no need to check params key encoding. Sinatra framework does it for us.
```
jlledom, line 414: Not anymore... https://github.com/rack/rack/pull/1797

lib/3scale/backend/rack/exception_catcher.rb
```diff
@@ -42,6 +40,11 @@ def call(env)
         rescue Backend::Error => e
           delete_sinatra_error! env
           respond_with e.http_code, prepare_body(e.to_xml, env)
+        rescue ArgumentError => e
+          return unhandled_exception(e) unless e.message.include?(INVALID_BYTE_SEQUENCE_ERR_MSG)
+
+          delete_sinatra_error! env
+          respond_with 400, Backend::NotValidData.new.to_xml
```
jlledom, line 47: Check https://github.com/3scale/apisonator/pull/411/commits/25d53c9feee8cb132156626d349dafc322ac0b16

Basically, an Invalid UTF8 error can be raised at any place in the code and we centralize the handling here.

lib/3scale/backend/logging/middleware.rb
```diff
@@ -35,7 +35,7 @@ def call(env)
             raise e
           end
 
-          header = ::Rack::Utils::HeaderHash.new(header)
+          header = ::Rack::Headers.new.merge(header)
```
jlledom, line 38: Check https://github.com/rack/rack/blob/main/CHANGELOG.md#removed-1

spec/api/listener/cors_spec.rb
```diff
@@ -5,14 +5,8 @@ module Backend
...
-      REQUIRED_ENDPOINT_HEADERS = [
-        'Access-Control-Allow-Origin',
-        'Access-Control-Expose-Headers',
-      ]
-      REQUIRED_OPTIONS_HEADERS = REQUIRED_ENDPOINT_HEADERS + [
-        'Access-Control-Allow-Methods',
-        'Access-Control-Allow-Headers',
-      ]
+      REQUIRED_ENDPOINT_HEADERS = %w[access-control-allow-origin access-control-expose-headers]
+      REQUIRED_OPTIONS_HEADERS = REQUIRED_ENDPOINT_HEADERS + %w[access-control-allow-methods access-control-allow-headers]
```
jlledom, line 9: For Rack 3: https://github.com/rack/rack/blob/main/CHANGELOG.md#spec-changes-2

spec/integration/worker_metrics_spec.rb
```diff
@@ -130,7 +130,7 @@ def process_jobs(worker, num)
 
       def shutdown_metrics_server
         # Yabeda does not expose this
-        ::Rack::Handler::WEBrick.shutdown
+        ::Rackup::Handler::WEBrick.shutdown
```
jlledom, line 133: Rackup was part of Rack, then moved to an independent gem. Rack 3 compatible apps use Rackup.

test/integration/authorize/basic_test.rb
```diff
@@ -566,7 +566,7 @@ def setup
       'HTTP_3SCALE_OPTIONS' => Extensions::REJECTION_REASON_HEADER
 
     assert_equal 409, last_response.status
-    assert_equal 'limits_exceeded', last_response.header['3scale-rejection-reason']
+    assert_equal 'limits_exceeded', last_response.headers['3scale-rejection-reason']
```
jlledom, line 569: https://github.com/rack/rack/blob/main/CHANGELOG.md#removed-1

test/integration/report_test.rb
```diff
@@ -938,13 +938,6 @@ def setup
     assert_error_resp_with_exc(ThreeScale::Backend::ProviderKeyOrServiceTokenRequired.new)
   end
 
-  test 'returns 400 when param key encoding is not valid utf-8' do
-    post '/transactions.xml', "\xf0\x90\x28\xbc" => 1
-
-    assert_equal 400, last_response.status
-    assert_equal ThreeScale::Backend::NotValidData.new.to_xml, last_response.body
-  end
```
jlledom, line 946: After https://github.com/rack/rack/pull/1797, this doesn't happen anymore.

config/puma.rb
mayorova: So, this `config/puma.rb` file is only used for dev/test?

What happens if `preload_app!` is enabled also? 

If it can be, we could probably simplify the config, and remove `openshift/config/puma.rb` ?

lib/3scale/backend/listener.rb
```diff
@@ -457,11 +451,12 @@ def valid_usage_params?
         params[:usage].nil? || params[:usage].is_a?(Hash)
       end
 
-      def check_params_value_encoding!(input_params, params_to_validate)
-        params_to_validate.each do |p|
-          param_value = input_params[p]
-          if !param_value.nil? && !param_value.valid_encoding?
-            halt 400, ThreeScale::Backend::NotValidData.new.to_xml
+      def check_params_encoding!(input_params)
```
mayorova, line 454: So, you're now checking all parameters, not only the list predefined in `REPORT_EXPECTED_PARAMS` ?

lib/3scale/backend/rack/exception_catcher.rb
```diff
@@ -117,13 +120,10 @@ def filter_encoding_error_response(resp, env)
...
 
-          if resp_body == INVALID_BYTE_SEQUENCE_ERR_MSG
-            delete_sinatra_error! env
-            resp = respond_with 400, Backend::NotValidData.new.to_xml
-          elsif resp_body.start_with?(EXPECTED_HASH_ERR_MSG)
+          if resp_body.start_with?(EXPECTED_HASH_ERR_MSG)
             delete_sinatra_error! env
             resp = respond_with 400, Backend::BadRequest.new.to_xml
-          elsif resp_body.start_with?(INVALID_PERCENT_ENCODING_ERR_MSG)
+          else resp_body.start_with?(INVALID_PERCENT_ENCODING_ERR_MSG)
```
mayorova, line 126:
```suggestion
          elsif resp_body.start_with?(INVALID_PERCENT_ENCODING_ERR_MSG)
```

test/integration/report_test.rb
mayorova: I don't quite understand how that PR fixes this behavior... but if this is now considered a valid input, maybe let's keep the test, and just change the expected response?

lib/3scale/backend/listener.rb
jlledom: yes, because `check_params_encoding!` is not only called from `/transactions.xml` like before, now it's also called from `do_api_method` which is used in many endpoints that receive different parameters.

It also checks for key encodings, not only values as before, since Sinatra doesn't do it anymore: https://github.com/rack/rack/pull/1797

lib/3scale/backend/rack/exception_catcher.rb
jlledom: Ouch!

config/puma.rb
jlledom: I don't know the answer to your questions, so I'll try to make up something convincing:

> So, this config/puma.rb file is only used for dev/test?

I don't remember why I said that, but I think it's true because the `Dockerfile` replaces the file with the one under `openshift/` folder:

https://github.com/3scale/apisonator/blob/a9dc73444f6da99de237a4a29686b2347abddb77/openshift/distro/ubi/8/release/Dockerfile#L60

> What happens if preload_app! is enabled also?

The docs say it conflicts with phased restart, so I haven't tried locally. I don't know what would happen.

> If it can be, we could probably simplify the config, and remove openshift/config/puma.rb ?

I think it should work, yes.

test/integration/report_test.rb
jlledom: That PR fixes the behavior accidentally, because it was never their intention to check for encoding validity, it was just a side-effect of using a regular expressions to parse parameters. Since they don't use regexps anymore, all params with invalid encoding now simply pass through rack and reach apisonator.

I added the test back https://github.com/3scale/apisonator/pull/411/commits/a8bd74cfc3cad819fa59c045560cb98441593626

config/puma.rb
mayorova: So, as it seems that we are not actually using phased restart, maybe we could just remove `openshift/config/puma.rb` and enable `preload_app!` in `config/puma.rb` and use the latter one everywhere. WDYT?

jlledom: > So, as it seems that we are not actually using phased restart, maybe we could just remove openshift/config/puma.rb and enable preload_app! in config/puma.rb and use the latter one everywhere. WDYT?

Yes, we could, but I think we should do it in another PR because we should also edit the dockerfiles to stop referencing `openshift/config/puma.rb`, which is out of this PR scope I think.

mayorova: OK

mayorova (approved):

config/puma.rb
akostadinov: I didn't read about this restart but I doubt we do restarting in pods. Rather I assume openshift pod restart takes place, which is somehow phased.

lib/3scale/backend/job_fetcher.rb
```diff
@@ -85,9 +85,6 @@ def try_pop_from_queue(max=nil)
       def wait_pop_from_queue
         queue, encoded = @redis.blpop(*@queues, timeout: @fetch_timeout)
         [queue, [encoded]] if encoded
-      rescue RedisClient::ReadTimeoutError => _e
-        # Ignore this exception, this happens because of a bug on redis-rb, when connecting to sentinels.
-        # Check: https://github.com/redis/redis-rb/issues/1279
```
akostadinov, line 90: :muscle:

lib/3scale/backend/listener.rb
```diff
@@ -457,11 +451,12 @@ def valid_usage_params?
...
+        input_params.each do |key, value|
+          raise ArgumentError, Rack::ExceptionCatcher::INVALID_BYTE_SEQUENCE_ERR_MSG unless key.valid_encoding?
+
+          if !value.nil? && !value.valid_encoding?
```
akostadinov, line 458: A little bit easier to read.
```suggestion
          if value && !value.valid_encoding?
```

But just a suggestion. Feel free to merge as you wish.

akostadinov (approved):
looks good

lib/3scale/backend/listener.rb
jlledom: This would behave differently when `value` is `false`, check: https://stackoverflow.com/questions/18060790/ruby-if-object-nil-or-if-object.

PR #413, opened 2025-01-08
Fix tests for recent openssl versions

jlledom:
After upgrading to Fedora 41, my openssl lib was updated and tests started to fail locally.

In my tests I found it was because using a DSA key and a SHA1 digest algorithm altogether fails with a `invalid digest` error. If it try to use SHA512 as a digest algorithm for DSA, then I receive a `ca md too weak` error. However, the combination DSA + SHA256 works fine for me locally.

I just refactored the code and copied the code to generate the fake certs from the [official docs](https://docs.ruby-lang.org/en/3.1/OpenSSL/X509/Certificate.html#class-OpenSSL::X509::Certificate-label-Creating+a+root+CA+certificate+and+an+end-entity+certificate), which use SHA256.

### DIFF
--- a/test/test_helpers/certificates.rb
+++ b/test/test_helpers/certificates.rb
@@ -28,31 +28,48 @@ module TestHelpers
       end
     end
 
-    def create_cert(key = create_key(:rsa))
-      public_key = get_public_key(key)
+    def create_ca(key = create_key(:rsa))
+      root_ca = OpenSSL::X509::Certificate.new
+      root_ca.version = 2 # cf. RFC 5280 - to make it a "v3" certificate
+      root_ca.serial = 0x1
+      root_ca.subject = OpenSSL::X509::Name.parse "/DC=test/DC=backend/CN=TestCA"
+      root_ca.issuer = root_ca.subject # root CA's are "self-signed"
+      root_ca.public_key = get_public_key(key)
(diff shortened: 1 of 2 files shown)


test/test_helpers/certificates.rb
```diff
@@ -28,31 +28,48 @@ def create_key(alg)
       end
     end
 
-    def create_cert(key = create_key(:rsa))
-      public_key = get_public_key(key)
+    def create_ca(key = create_key(:rsa))
```
mayorova, line 31: Hm... the description says that the working combination is DSA + SHA256, but I don't see anywhere where DSA is forced, it seems that everything defaults to RSA?...

jlledom: We have tests for each algorithm:

https://github.com/jlledom/apisonator/blob/b471437ee6a8915d1a00e51d8f3a6cc637b02ef0/test/unit/storage_async_test.rb#L234-L251

Before this PR, the tests for `:ec` and `:rsa` passed, but the test for `:dsa` failed with the `invalid digest` error.

mayorova: Ah, OK, that last phrase clarifies the issue. Thanks!

jlledom: Sorry for the confusion!

PR #418, opened 2025-01-28
Move `apisonator` to RHEL9 and Ruby 3.3

lvillen:
In order to fix [THREESCALE-11083: CVE-2024-4741 3scale-amp-backend-container: openssl: Use After Free with SSL_free_buffers [3amp-2.14]](https://issues.redhat.com/browse/THREESCALE-11083), we should move `apisonator` from RHEL8 to RHEL9.


jlledom (commented):
Have you built the images locally?

lvillen:
> It's strange that the dockerfiles were still referring to Ruby 3.0, while the code was apparently already upgraded to Ruby 3.3 previously 🙃 #396

Well, I thought it was strange too, that's why I updated it. I'm not sure if there's a reason not to do it or if it's just that we missed doing it 🤔

openshift/distro/ubi/9/release/Dockerfile
jlledom, line 1: I don't know what's this file for. `make build` uses the other `Dockerfile`

openshift/distro/ubi/9/Dockerfile
```diff
@@ -26,22 +26,19 @@
...
 WORKDIR "${HOME}/app"
 
-ARG RUBY_VERSION="3.0"
-ARG RUNTIME_DEPS="ruby"
+ARG RUBY_VERSION="3.3"
 
-RUN echo -e "[ruby]\nname=ruby\nstream=${RUBY_VERSION}\nprofiles=\nstate=enabled\n" > /etc/dnf/modules.d/ruby.module \
- && microdnf update --nodocs \
- && microdnf install --nodocs ${RUNTIME_DEPS} \
+RUN microdnf module enable -y "ruby:${RUBY_VERSION}" \
```
jlledom, line 36: you can use `microdnf` to enable the `ruby` module, so there's no need to install it manually. This change it's not crucial because it worked with the previous code, but I think it's more readable.

openshift/distro/ubi/9/Dockerfile
```diff
@@ -26,22 +26,19 @@
...
  && chown -R 1001:1001 "${HOME}"
 
 ARG BUILD_DEPS="tar make file findutils git patch gcc automake autoconf libtool redhat-rpm-config openssl-devel ruby-devel"
 
-RUN microdnf install --nodocs ${BUILD_DEPS}
+RUN microdnf install -y --nodocs ${BUILD_DEPS}
```
jlledom, line 41: Apparently `microdnf` in RHEL8 didn't stop to query the user. Now we have to add `-y` to ensure the process doesn't require user intervention

openshift/distro/ubi/9/Dockerfile
```diff
@@ -54,7 +51,7 @@ ARG BUNDLE_WITHOUT=development:test
 COPY --chown=1001:1001 ${BUNDLE_GEMFILE}.lock "${HOME}/app/"
 
 RUN mkdir -p "${HOME}/.gem/bin" \
- && echo "gem: --bindir ~/.gem/bin" > "${HOME}/.gemrc"
+ && echo "gem: --bindir ~/.gem/bin --install-dir /usr/share/gems" > "${HOME}/.gemrc"
```
jlledom, line 54: This was weird. `gem` is configured by default to use `/usr/local/share/gems` as install dir, but somehow browsing through the fs I found the gem files are installed under `/usr/share/gems`. Not sure why this happens but manually providing the proper install dir is now required for `gem uninstall` below to work properly

openshift/distro/ubi/9/Dockerfile
```diff
@@ -82,6 +79,7 @@ RUN cp -n openshift/3scale_backend.conf /etc/ \
  && gem unpack "apisonator-${BACKEND_VERSION}.gem" --target=/opt/ruby \
  && cd "/opt/ruby/apisonator-${BACKEND_VERSION}" \
  && cp -a ${HOME}/app/.bundle "/opt/ruby/apisonator-${BACKEND_VERSION}/" \
+ && bundle config --local path "/opt/ruby/apisonator-${BACKEND_VERSION}/" \
```
jlledom, line 82: This is a bit tricky, apparently bundle gems where installed at `${HOME}/app/.bundle` but the `bundle` command were looking for them at some other location, not sure where. Anyway this fixes the problem.

openshift/distro/ubi/9/release/Dockerfile
lvillen: Ok, so I've been tracking this and it was introduced in #268, which is mentioned in https://github.com/3scale/3scale-operator/pull/578 and it's related with nightly builds. The file where this `Dockerfile` was referenced was deleted as part of https://github.com/3scale/3scale-operator/pull/636 and I don't see any reference to that file in https://github.com/3scale/3scale-operator so I think we're safe to remove it.

jlledom (approved):

PR #419, opened 2025-02-12
THREESCALE-11536: remove `rspec_api_documentation`

jlledom:
**Description**:

This removes the `rspec_api_documentation` gem. It's used to generate some documentation for the backend API from a few tests suites written using a particular DSL.

I'm not sure where is this documentation used/published/needed... I never heard of it before. The gem is a problem because it's abandoned and can't be updated. Check the description in https://github.com/3scale/apisonator/pull/411

**Issue**:

https://issues.redhat.com/browse/THREESCALE-11536

**Notes for reviewers**:

This PR consists on removing the gem and migration the affected test suites to the regular rspec syntax, basically:

- move `header` calls to inside a `before` callback
- replace `resource` by `describe`
- replace `example` and `example_request` by `it`
- Launch the requests directly through method names (`get`, `post`, etc), instead of calling `do_request` (under `example`) or making the call implicit (under `example_request`)
- remove calls to `parameter`, they were intended for documentation purposes and to generate the request from inside `example_request`. Not needed in  regular rspec.

### DIFF
--- a/docs/specs/backend_internal.yaml
+++ b/docs/specs/backend_internal.yaml
@@ -1,0 +1,939 @@
+openapi: 3.0.1
+info:
+  title: 3scale Backend Internal API
+  version: 1.0.0
+servers:
+- url: http://localhost:3001/internal
+paths:
+  /services/{service_id}/alert_limits/:
+    get:
+      tags:
+      - Alert limits
+      parameters:
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          description: Getting alert limits
+    post:
+      tags:
+      - Alert limits
+      parameters:
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AlertLimit'
+        required: true
+      responses:
+        "201":
+          description: Add an alert limit
+  /services/{service_id}/alert_limits/{value}:
+    delete:
+      tags:
+      - Alert limits
+      parameters:
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
+      - name: value
+        in: path
+        description: Limit value
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          description: Delete an alert limit
+  /services/{service_id}/applications/{app_id}/keys/:
+    get:
+      tags:
+      - Application keys
+      parameters:
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
+      - name: app_id
+        in: path
+        description: Application ID
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          description: Getting application keys
+    post:
+      tags:
+      - Application keys
+      parameters:
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
+      - name: app_id
+        in: path
+        description: Application ID
+        required: true
+        schema:
+          type: string
+      requestBody:
+        description: Application key value
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ApplicationKey'
+        required: true
+      responses:
+        "201":
+          description: Add an application key
+  /services/{service_id}/applications/{app_id}/keys/{value}:
+    delete:
+      tags:
+      - Application keys
+      parameters:
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
+      - name: app_id
+        in: path
+        description: Application ID
+        required: true
+        schema:
+          type: string
+      - name: value
+        in: path
+        description: Application key value
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          description: Delete an application key
+  /services/{service_id}/applications/{app_id}/referrer_filters:
+    get:
+      tags:
+      - Application Referrer Filters
+      parameters:
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
+      - name: app_id
+        in: path
+        description: Application ID
+        required: true
+        schema:
+          type: string
+      responses:
+        "200":
+          description: Getting referrer filters
+    post:
+      tags:
+      - Application Referrer Filters
+      parameters:
+      - name: service_id
+        in: path
+        required: true
+        schema:
+          type: string
+      - name: app_id
+        in: path
+        required: true
+        schema:
+          type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ReferrerFilter'
+        required: true
+      responses:
+        "201":
+          description: Create a referrer filter
+        "400":
+          description: Try updating a referrer filter with invalid data
+        "404":
+          description: Try updating a referrer filter with invalid application id
+  /services/{service_id}/applications/{app_id}/referrer_filters/{filter}:
+    delete:
+      tags:
+      - Application Referrer Filters
+      parameters:
+      - name: filter
+        in: path
+        required: true
+        schema:
+          type: string
+      - name: service_id
+        in: path
+        description: Service ID
+        required: true
+        schema:
+          type: string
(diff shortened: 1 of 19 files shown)


akostadinov:
I even don't want to review everything in details. It looks good from what I saw. One question though is, can this be easily conerted to some other lib that generates openapi docs? Or at least tag the tests somehow, that they are kind of minimal tests that can be used for apidocs generation in case we want to use such a library in the future?

Apisonator might be a good testbed in case we want to do something similar with porta.

jlledom:
> One question though is, can this be easily conerted to some other lib that generates openapi docs? Or at least tag the tests somehow, that they are kind of minimal tests that can be used for apidocs generation in case we want to use such a library in the future?

I don't know of any library able to understand `rspec_api_documentation` DSL. I was looking for one a couple of months ago when I was working on the sinatra update and couldn't find one.

`rspec_api_documentation` can be used to generate openapi jsons: https://github.com/zipmark/rspec_api_documentation?tab=readme-ov-file#open_api

So we could generate them now and add them to the PR if they are good enough. If we make changes to the backend API later on, we'll have to update the openapi jsons manually

akostadinov:
IIRC looking at something like this in the past, there were a couple of recent libraries that used rspec dsl to generate api docs. So my question was whether migrating to one of these, instead of generic rspec would be easy enough. Or somehow mark the tests so it can be done at a laterpoint.

jlledom:
> Question: is the gem really abandoned? There is a whole bunch of commits from 2024.

We have been discussing this after the sync, it got some updates the first quarter last year, but nothing since 10 months ago, also no releases since 2018, not supporting OAS3, etc... and forcing us to maintain a fork because it doesn't support rack 3.

spec/spec_helper.rb
```diff
@@ -2,8 +2,8 @@
 require 'resque_spec'
 require 'async'
 
-# This fixes a `NoMethodError` error caused by `rspec_api_documentation`, which is fixed since 2022, but never released.
-# Ref: https://github.com/zipmark/rspec_api_documentation/commit/758c879893a21233c0eb977e79ef026f263fc37e
+require 'active_support'
+require 'active_support/core_ext/object/json'
```
mayorova, line 6: so, requiring `active_support` isn't pulling the `core_ext` dependencies?

jlledom: Apparently it doesn't. Tests fail if I remove any of the three `active_support` requires.

https://guides.rubyonrails.org/active_support_core_extensions.html

jlledom:
@akostadinov I removed the rake task: https://github.com/3scale/apisonator/pull/419/commits/7ef07fec2b6a26cd4f6692b96e2a2845a149f763

@mayorova I pushed the static OAS file: https://github.com/3scale/apisonator/pull/419/commits/e6122dc7dd742b67fea3f43992f93d93a3896e65

I made some small changes like renaming some endpoints to remove the `(Prefix: ...)` in the title, for consistency, because some endpoints had it and some others didn't

spec/acceptance/api/internal/alert_limits_spec.rb
```diff
@@ -1,94 +1,100 @@
...
-  end
 
-  post '/services/:service_id/alert_limits/' do
-    parameter :service_id, 'Service ID', required: true
-    parameter :alert_limit, 'Limit value', required: true
+    context 'POST' do
+      let(:service_id)  { '7575' }
 
-    let(:service_id)  { '7575' }
-    let(:raw_post) { params.to_json }
```
mayorova, line 39: For reference:
https://github.com/zipmark/rspec_api_documentation/blob/d32bf17932a379052e8bb7540fc29ce7e72712ae/README.md#raw_post

mayorova:
This looks great, I would suggest though to move the `backend_internal.yaml` to another folder.
I find `active_docs` misleading, because this spec will not actually be used for the admin portal's active docs. Moreover, in another PR I would remove the `Service Management API.json`, because it's obsolete. Maybe we can put this yaml to `docs/specs` or something like this.

spec/acceptance/api/internal/alert_limits_spec.rb
jlledom: Thanks for the info 👍

jlledom:
@mayorova a few points:

- I removed the fake endpoint
- endpoints `status` and `check.json` are fine. I though they wer out of the `internal` namespace but it turns out they don't. Because there are both `/status` and `/internal/status`, same for `check.json`
- everything under `/internal/` needs to  authorized by HTTP Basic Auth. I added the changes to the yaml but after a few hours I'm starting to think SwaggerUI has a bug, because it considers the yaml valid, it finds the auth and it adds the header on the auto-generated `curl` it shows as example, but it doesn't send the header under the hood.

> This looks great, I would suggest though to move the `backend_internal.yaml` to another folder. I find `active_docs` misleading, because this spec will not actually be used for the admin portal's active docs. Moreover, in another PR I would remove the `Service Management API.json`, because it's obsolete. Maybe we can put this yaml to `docs/specs` or something like this.

OK

mayorova:
> * I'm starting to think SwaggerUI has a bug, because it considers the yaml valid, it finds the auth and it adds the header on the auto-generated `curl` it shows as example, but it doesn't send the header under the hood.

I think it's not SwaggerUI, but the JetBrains plugin, because in https://editor.swagger.io/ the Authorization header is sent as expected (but it has an issue with CORS when trying to hit a locally running apisonator :grimacing:)

mayorova (approved):
Looks great to me :+1:

akostadinov:
It makes no sense to address localhost from a remote site... there might be certain protections. How about trying to hit whatever public accessible http server?

jlledom:
> It makes no sense to address localhost from a remote site... there might be certain protections. How about trying to hit whatever public accessible http server?

When using the RubyMine plugin, we are querying `localhost` from `localhost`. So no CORS problem here. I guess that's why I don't see the `OPTIONS` query in this scenario. However, the credentials are still not sent

akostadinov:
How is the request being performed? some js? or as a normal form submission?

jlledom:
> How is the request being performed? some js? or as a normal form submission?

I tried three ways to send the request:

1. Broswer: https://editor.swagger.io/
image
2. From RubyMine:
image
3. From console:
```
$ curl -X 'GET'   'http://localhost:3001/internal/status'   /
       -H 'accept: */*'   /
       -H 'Authorization: Basic c3lzdGVtX2FwcDpwYXNzd29yZA=='

{"status":"ok","version":{"backend":"3.4.3"}}
```

mayorova:
@jlledom @akostadinov Yes, it can easily be confirmed that the requests sent by the plugin to not contain the proper Authorization header by updading the spec URL to some echo API (despite the generated curl that does have the proper Authorization header)

As for CORS, inded, I think different browsers may show the CORS flow differently, I think it doesn't matter.
I enabled CORS on the internal API on my locally running apisonator, and it all works fine from editor.swagger.io.

So, no issue (just that the plugin is slightly less usable than I was hoping for)

docs/specs/backend_internal.yaml
```diff
@@ -0,0 +1,939 @@
+openapi: 3.0.1
+info:
+  title: 3scale Backend Internal API
+  version: 1.0.0
+servers:
+- url: http://localhost:3000/internal
```
mayorova, line 6:
```suggestion
- url: http://localhost:3001/internal
```

Because normally backend runs on `3001`, while `3000` is used for porta

mayorova: But we can leave it until further spec improvements.

mayorova:
> So, no issue (just that the plugin is slightly less usable than I was hoping for)

There is actually a workaround. If apisonator is started with empty `CONFIG_INTERNAL_API_USER` and `CONFIG_INTERNAL_API_PASSWORD` values, then the internal API will not have authentication.

jlledom:
> As for CORS, inded, I think different browsers may show the CORS flow differently, I think it doesn't matter. I enabled CORS on the internal API on my locally running apisonator, and it all works fine from editor.swagger.io.
> 
> So, no issue (just that the plugin is slightly less usable than I was hoping for)

Thanks for checking!

docs/specs/backend_internal.yaml
jlledom: I pushed the change

mayorova (approved):

PR #422, opened 2025-04-08
Internal API specification updates

mayorova:
Backend API specification based on OpenAPI v3.1

### DIFF
--- a/docs/specs/backend_internal.yaml
+++ b/docs/specs/backend_internal.yaml
@@ -1,4 +1,4 @@
-openapi: 3.0.1
+openapi: 3.1.0
 info:
   title: 3scale Backend Internal API
   version: 1.0.0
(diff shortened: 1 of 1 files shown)


docs/specs/backend_internal.yaml
```diff
@@ -1,586 +1,1266 @@
...
   version: 1.0.0
 servers:
-- url: http://localhost:3001/internal
+  - url: http://localhost:3001/internal
 paths:
   /services/{service_id}/alert_limits/:
     get:
       tags:
-      - Alert limits
+        - Alert limits
```
akostadinov, line 11: any important reason to change the indentation? :zipper_mouth_face:

mayorova: No... just my IDE does that automatically, apparently. Or, maybe it's https://editor-next.swagger.io/ that I'm using.

I'll try to revert those before opening the PR for review.

app/api/internal/internal.rb
```diff
@@ -31,6 +32,20 @@ class Internal < Sinatra::Base
           parse_json_params params
         end
 
+        before do
```
mayorova, line 35: This is just temporary to test it properly from https://editor-next.swagger.io/ 
(just to avoid running swagger-ui locally)

In fact, I was not going to commit it, it was an accident. I'll remove later.

docs/specs/backend_internal.yaml
jlledom: May the change is not needed but it's fine to leave it IMO.

akostadinov: This is two bytes of waste, and loss of priceless screen estate, unacceptable! :angel:

jlledom: Your point is irrefutable. However, in order to keep consistency with the rest of the file, those two spaces should be removed also from all new sections added in this PR.

Lacking of consistence inside within the same file... how unacceptable, right?

akostadinov: If such thing happens, we'll have to remove all copies of the source code of this project and then start it over. We must be careful lest nobody remains to tell the story.

jlledom: If you value consistency, yes, definitely we should.

jlledom (approved):
I couldn't send the requests because of the CORS error, but it looks good overall.

PR #427, opened 2025-05-13
THREESCALE-11822: Use `falcon host` to deploy

jlledom:
According to falcon docs, we should be using `falcon host` instead of `falcon serve` to deploy in production.

https://github.com/socketry/falcon/blob/main/guides/deployment/readme.md

As I understood it, the main reason to not use `serve` is the command line interface, which they don't guarantee to keep it stable in next versions. `host`, on the other hand, is configured through a config file. This is better, I think, because if they include breaking changes on future versions, we can fix them from our code base and won't need to edit the Containerfile or scripts.

This PR adds that new config file for falcon and set the same attributes we were setting for `serve`:

1. port
2. prometheus
3. number of workers

I think we can just use `host` for development too. I tried RubyMine, the debugger, etc and everything works fine.

**Jira Issue**

https://issues.redhat.com/browse/THREESCALE-11822


lib/3scale/backend/server/falcon.rb
```diff
@@ -14,16 +16,11 @@ def self.start(global_options, options, args)
...
           return unless manifest
 
-          argv = ['falcon']
-          argv_add argv, true, '--bind', 'http://0.0.0.0'
-          argv_add argv, options[:port], '--port', options[:port]
-
-          # Starts the prometheus server if needed. Just once even when spanning
-          # multiple workers.
-          argv_add argv, true, '--preload', 'lib/3scale/prometheus_server.rb'
+          ENV["PORT"] = options[:port] if options[:port]
```
jlledom, line 19: The port comes from `3scale_backend` script, and it's not directly reachable from the new config file `falcon.rb`. So I'm setting this env variable here to be read from the config file. Ugly but effective.

falcon.rb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env falcon host
+# frozen_string_literal: true
+
+require 'falcon/environment/rack'
+require 'falcon/environment/supervisor'
+require '3scale/backend/manifest'
+
+HOSTNAME = 'apisonator_listener'
```
akostadinov, line 8: `_` is not a valid domain name character

lib/3scale/backend/server/falcon.rb
```diff
@@ -4,6 +4,8 @@ module Server
       class Falcon
         extend ThreeScale::Backend::Server::Utils
 
+        CONFIG = 'falcon.rb'
```
mayorova, line 7: Isn't this the default value? I think even if we don't set it it will be used anyway?

spec/spec_helper.rb
```diff
@@ -64,7 +64,7 @@ def committed_at
   end
 
   config.around :each do |example|
-    Async.run do
+    Async do
```
akostadinov, line 67: Perhaps this would do as we can't run in parallel anyway
```suggestion
    Sync do
```

akostadinov: But fine to keep as it was like that

test/integration/huge_payload_test.rb
```diff
@@ -23,7 +23,7 @@ def setup
   end
 
   def test_report_handles_huge_payloads
-    payload = generate_payload((ENV['HUGE_PAYLOAD_SIZE'] || 10000).to_i)
+    payload = generate_payload((ENV['HUGE_PAYLOAD_SIZE'] || 2000).to_i)
```
jlledom, line 26: I had to reduce this number. This PR bumps rack to `3.1.14` which solves [CVE-2025-46727](https://access.redhat.com/security/cve/cve-2025-46727). The test was sending `20001` parameters but now only up to `4096` are permitted. Is it expected to receive more than `4096` parameters? because that scenario will fail from now on.

Do you know of any scenario where such amount of parameters can be received?

akostadinov: Good question. Maybe under high load. Why was it previously set to 10k? :thinking:

lib/3scale/backend/server/falcon.rb
jlledom: Yes, it is. We can remove it, I just add it to make it explicit like the `puma` file does.

https://github.com/3scale/apisonator/blob/3ab689d79d7d682ae89544ac46ffec2079a37e5c/lib/3scale/backend/server/puma.rb#L7

spec/spec_helper.rb
jlledom: I think it's fine to change it. https://github.com/3scale/apisonator/pull/427/commits/a2ed9ae99855f31999323c173af08010f630fb0e

test/integration/huge_payload_test.rb
jlledom: > Why was it previously set to 10k? 

I don't think there's a particular reason, just a high number probably.

> Maybe under high load

It's not directly related. Here we are not talking about 10000 requests, but one single request, with 20001 parameters in the URL. It's hard to think on a scenario when that can happen.

According to [docs](https://github.com/3scale/porta/blob/5a062aaf2ce3620a2969d0a7c50ce55283310c51/doc/active_docs/service_management_api.json#L388), only 1000 transactions per request are supported by this endpoint, that would require 2001 parameters in the URL. After the CVE fix, we accept up to 4096 parameters in the URL so I think we are fine.

falcon.rb
jlledom: https://github.com/3scale/apisonator/pull/427/commits/4b12b83e18f1522db9a24491ee9994637db79bdc

jlledom:
> We are using a single arg to `serve`.

We use for args: `--port`, `--bind`, `--preload` and `--count`

https://github.com/3scale/apisonator/blob/438ef605b13fa90663b2b63db7a6b61b870a2d88/lib/3scale/backend/server/falcon.rb#L17-L26

> What is the point in supporting a whole file with logic and stuff that can break?

Just because upstream docs say we should use `host` for deploying rather than `serve`, for future compatibility. I'll explain better in the description.

lib/3scale/backend/server/falcon.rb
mayorova: One problem I see with this is that it might be confusing to any user that they need (or can) set the port through the env var.
But if somebody sets both `PORT=3001` and provide `-p 3000`, the former will not be effective.

Ideally, we would have just one place where we'd set the port, and reuse it. But I understand that the flow is a bit clunky, and we can't easily fetch in `falcon.rb` what is passed as command arguments to `3scale_backend` :/

So, I'm fine with this.

mayorova (approved):
Looks good to me

test/integration/huge_payload_test.rb
akostadinov: I'd like us to see what production traffic we see and to make sure we have a good margin over it.

But to save me from looking, could you tell me whether this is query parameters or POST data?

akostadinov:
> Just because upstream docs say we should use host for deploying rather than serve, for future compatibility.

And it has worked for 10 years like that without compatibility issues. IMO not really worth unless it simplifies what we do. 

But at the end, as you wish. I abstain from voting on it.

lib/3scale/backend/server/falcon.rb
jlledom: Thanks.

Just replacing this line by 

```suggestion
          ENV["PORT"] |= options[:port]
```

will give preference to the env variable, if you think that's better.

mayorova: To be honest, I'm not sure what is better :grimacing: 
But I think having a possibility of relying on `PORT` could be nice.

test/integration/huge_payload_test.rb
jlledom: POST data

jlledom:
> And it has worked for 10 years like that without compatibility issues. IMO not really worth unless it simplifies what we do.
> 

Yeah, `host` doesn't simplify things, it makes things more complicated in fact, but I think being the official recommendation for production is enough to be worth the change. Also, `host` was introduced a few years ago, it didn't exist before. Also, async mode was not the default until our 2.15 release, so I don't think many clients, if any, ever used `falcon serve` before.

lib/3scale/backend/server/falcon.rb
jlledom: I've seen that the `-p` parameter is always being provided. It's enforced by the `3scale_backend` script, and set to `-p 3000` by default. So `PORT` env variable is always going to be ignored unless we give it priority. I pushed some changes: https://github.com/3scale/apisonator/pull/427/commits/7aa1a03f5402b2e5e6fa6ccadb6b04b2a9bcbac8

akostadinov:
> I think being the official recommendation for production is enough to be worth the change.

I would suggest to use whatever is simpler and easier to understand and maintain. We are only setting port which is very unlikely to change in `serve` ever. Moreover using a container gives us the opportunity to catch in case interface changed before releasing.

>  I don't think many clients, if any, ever used `falcon serve` before.

It doesn't matter whether many customers ever used `serve` or not. They just run the image and what is used is transparent.

My main concern is that I see no reason to complicate the way we start the server. I see no benefit despite a readme note that is not even there in all parts of documentation, to complicate our deployment. History knows many occasions where project didn't think through some changes. This one smells to me like one of those and will probably be changed orver time, to perhaps simplify it or drop it.

But again, I let you do what you like. Not such a big deal I hope.

jlledom:
> > I think being the official recommendation for production is enough to be worth the change.
> 
> I would suggest to use whatever is simpler and easier to understand and maintain. We are only setting port which is very unlikely to change in `serve` ever. Moreover using a container gives us the opportunity to catch in case interface changed before releasing.
> 
> > I don't think many clients, if any, ever used `falcon serve` before.
> 
> It doesn't matter whether many customers ever used `serve` or not. They just run the image and what is used is transparent.
> 
> My main concern is that I see no reason to complicate the way we start the server. I see no benefit despite a readme note that is not even there in all parts of documentation, to complicate our deployment. History knows many occasions where project didn't think through some changes. This one smells to me like one of those and will probably be changed orver time, to perhaps simplify it or drop it.
> 
> But again, I let you do what you like. Not such a big deal I hope.

@mayorova, you have to untie.

akostadinov:
>  I'm for using the libraries according to the documented way.

https://socketry.github.io/falcon/guides/getting-started/index.html

Just FYI. No need to untie. I'm giving my passive aggressive consent for the change.

jlledom:
> I'm giving my passive aggressive consent for the change.

OK. Please, can you give your passive aggressive approval?

test/integration/huge_payload_test.rb
jlledom: Check the `generate_payload` method, the received parameter is the number of parameters, but the total length must to be over 4096

PR #428, opened 2025-06-24
THREESCALE-8964: Fix key credentials validation

jlledom:
There issue warns about two scenarios:

1. `backend_version == 1` (auth by `user_key`): can auth by `app_id` only
2. `backend_version == 2` (auth by `app_id` + `app_key`): can auth by `user_key` + `app_key`

Apparently this behavior was expected, since there was a test to ensure `user_key` and `app_id` are exchangeable:
https://github.com/3scale/apisonator/blob/706507842c2b9eea6e99b187630ea703a92021d9/test/integration/backend_version_test.rb#L23

In this PR I modified the validator to  require `user_key` when backend version is 1 and require `app_id` when backend version is 2.

This might be a breaking change, but I don't see a reason why `app_id` and `user_key` should be exchangeable. The test was introduced in a very old [commit](https://github.com/3scale/apisonator/commit/7895d2f58515d6fd1d9e69b19faa7107b3d8b5ce#diff-ffdbe6975a1f996364c1f01ae334274340c7ffb5237701282d8ddb9114941ca3) with almost no description and not part of a PR so the original intention is hard to guess.

Anyway, I'd say very few clients could notice a difference, since most of clients don't access apisonator directly but behind apicast, which already does this parameter validation.

**Issue**: https://issues.redhat.com/browse/THREESCALE-8964

### DIFF
--- a/lib/3scale/backend/errors.rb
+++ b/lib/3scale/backend/errors.rb
@@ -260,7 +260,11 @@ module ThreeScale
 
     class UserKeyInvalid < Error
       def initialize(key)
-        super %(user key "#{key}" is invalid)
+        if key.blank?
+          super 'user key is missing'.freeze
+        else
+          super %(user key "#{key}" is invalid)
+        end
       end
     end
 
--- a/lib/3scale/backend/transactor.rb
+++ b/lib/3scale/backend/transactor.rb
@@ -67,10 +67,10 @@ module ThreeScale
         service_id = service.id
 
         app_id = params[:app_id]
-        # TODO: make sure params are nil if they are empty up the call stack
+        # Make sure params are nil if they are empty up the call stack
         # Note: app_key is an exception, as it being empty is semantically
         # significant.
-        params[:app_id] = nil if app_id && app_id.empty?
+        params[:app_id] = nil if app_id&.empty?
 
         # While OIDC without an app_id makes little sense, we would break existing
         # behaviour when calling non oauth_auth*.xml endpoints if we returned an
--- a/lib/3scale/backend/validators/key.rb
+++ b/lib/3scale/backend/validators/key.rb
@@ -3,12 +3,18 @@ module ThreeScale
     module Validators
       class Key < Base
         def apply
-          if service.backend_version.to_i == 1 ||
-              application.has_no_keys? ||
-              application.has_key?(params[:app_key])
-            succeed!
-          else
+          case service.backend_version.to_i
+          when 1 # Auth by user_key
+            return succeed! if params.has_key?(:user_key)
+
+            fail!(UserKeyInvalid.new(params[:user_key]))
+          when 2, 0  # 2 -> Auth by app_id. 0 -> not defined or invalid, must behave like 2
+            return succeed! if params.has_key?(:app_id) && (application.has_no_keys? || application.has_key?(params[:app_key]))
+
             fail!(ApplicationKeyInvalid.new(params[:app_key]))
+          else
+            # This should never happen
+            fail!(Error.new("Invalid backend version: #{service.backend_version.to_i}"))
           end
         end
       end
--- a/test/integration/backend_version_test.rb
+++ b/test/integration/backend_version_test.rb
@@ -20,13 +20,13 @@ class BackendVersionTest < Test::Unit::TestCase
     Application.save_id_by_key(@service_id, "user_key_#{@application.id}", @application.id)
   end
 
-  test 'test app_id and user_key are exchangeable regardless of the backend_version' do
+  test 'test app_id and user_key are not exchangeable regardless and are bound to the backend_version' do
     Service.save! id: @service.id, provider_key: @provider_key, backend_version: '1'
 
     get '/transactions/authorize.xml', :provider_key => @provider_key,
                                        :app_id       => @application.id
 
-    assert_authorized
+    assert_not_authorized
 
     get '/transactions/authorize.xml', :provider_key => @provider_key,
                                        :user_key     => "user_key_#{@application.id}"
@@ -43,7 +43,7 @@ class BackendVersionTest < Test::Unit::TestCase
     get '/transactions/authorize.xml', :provider_key => @provider_key,
                                        :user_key     => "user_key_#{@application.id}"
 
-    assert_authorized
+    assert_not_authorized
   end
 
   test 'service originally on backend_version 2 with apps with app_key does not complain about missing app_key when changed to backend_version 1' do
@@ -83,7 +83,6 @@ class BackendVersionTest < Test::Unit::TestCase
   test 'when backend_version is not declared should behave like backend_version two regarding the presence of app_key' do
     if @service.backend_version.nil? || @service.backend_version.empty?
       application_key_one = @application.create_key
-      _application_key_two = @application.create_key
 
       get '/transactions/authorize.xml', :provider_key => @provider_key,
                                        :app_id       => @application.id,
@@ -102,7 +101,6 @@ class BackendVersionTest < Test::Unit::TestCase
     Service.save! id: @service.id, provider_key: @provider_key, backend_version: '2'
 
     application_key_one = @application.create_key
-    application_key_two = @application.create_key
 
     get '/transactions/authorize.xml', :provider_key => @provider_key,
                                        :app_id       => @application.id
@@ -118,7 +116,7 @@ class BackendVersionTest < Test::Unit::TestCase
 
     Service.save! id: @service.id, provider_key: @provider_key, backend_version: 'oauth'
 
-    get '/transactions/authorize.xml', :provider_key => @provider_key,
+    get '/transactions/oauth_authorize.xml', :provider_key => @provider_key,
                                        :app_id       => @application.id
 
     assert_authorized
@@ -128,7 +126,6 @@ class BackendVersionTest < Test::Unit::TestCase
     Service.save! id: @service.id, provider_key: @provider_key, backend_version: '2'
 
     application_key_one = @application.create_key
-    application_key_two = @application.create_key
 
     get '/transactions/authorize.xml', :provider_key => @provider_key,
                                        :app_id       => @application.id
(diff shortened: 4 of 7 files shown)


akostadinov:
So you think with backend 2, we should not allow uth by user_key at all?

test/integration/backend_version_test.rb
```diff
@@ -118,7 +116,7 @@ def setup
 
     Service.save! id: @service.id, provider_key: @provider_key, backend_version: 'oauth'
 
-    get '/transactions/authorize.xml', :provider_key => @provider_key,
+    get '/transactions/oauth_authorize.xml', :provider_key => @provider_key,
```
mayorova, line 119: Why are all of these calls changed to `oauth_authorize` ?

jlledom: I think that's the correct endpoint because just above the backend version changed to `oauth`

```ruby
backend_version: 'oauth'
```

lib/3scale/backend/validators/key.rb
```diff
@@ -3,12 +3,18 @@ module Backend
...
-              application.has_no_keys? ||
-              application.has_key?(params[:app_key])
-            succeed!
-          else
+          case service.backend_version.to_i
+          when 1 # Auth by user_key
+            return succeed! if params.has_key?(:user_key)
+
+            fail!(UserKeyInvalid.new(params[:user_key]))
+          when 2, 0  # 2 -> Auth by app_id. 0 -> not defined or invalid, must behave like 2
```
mayorova, line 11: hm, is it actually possible to have `backend_version.to_i == 0` ? In what scenario?
Not sure if it's right...

mayorova:
I found it easier to compare the before and after behavior in a spreadsheet (I am a bit lost in tests): 
https://docs.google.com/spreadsheets/d/13FKZn_5kaXp8eA-GOT5OQEPAr-W4QqaYFeTK8semvmo/edit?gid=0#gid=0

I'll summarize here:

for App ID / App Key mode:
"valid user_key + valid app_key" combination used to be returning 200 OK, but now returns:
```
409
<status>
    <authorized>false</authorized>
    <reason>application key "955112421a0ea5c563663659890e96f2" is invalid</reason>
    <plan>aaa</plan>
</status>
```

I think this is misleading, because 1. the app key is actually valid, so the message is not accurate. 2. the actual problem is that `app_id` is not provided, while expected in this mode, so I think this is what the error/reason should say. In fact, I think it even makes more sense to return a 403 with something like

```
<error code="app_id_missing">application id is missing</error>
```

Because I think 409 with `status` assumes that we have identified the application, but for some reason couldn't authorize. But 403 means that we couldn't even identify the app. I think the latter is more suitable, because we are expecting `app_id` in this mode, and `user_key` should be irrelevant, so I think we shouldn't even use it for identifying the application?

Not sure though...

akostadinov:
> And [you agree](https://issues.redhat.com/browse/THREESCALE-8964?focusedId=21621877&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21621877) 😛

I swear I have nothing in common with the guy you are linking to.

lib/3scale/backend/validators/key.rb
```diff
@@ -3,12 +3,18 @@ module Backend
...
+            return succeed! if params.has_key?(:app_id) && (application.has_no_keys? || application.has_key?(params[:app_key]))
```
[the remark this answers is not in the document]
akostadinov: I wonder if we should just add logging as a first step and see if there are any valid uses skipping that parameter. In an ideal world we would do this but not sure we should allow this anyway.

lib/3scale/backend/validators/key.rb
jlledom: when `backend_version` is not a valid number, `to_i` returns `0`. AFAIK `backend_version` can only be `1`, `2` and `oauth`. So in this case it will be `0` for `oauth`.

lib/3scale/backend/validators/key.rb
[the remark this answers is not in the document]
jlledom: yeah, it doesn't seem to make sense. But there are tests to verify that behavior:

https://github.com/3scale/apisonator/blob/544859272203b5b60c9587a03761a13673ce14d5/test/unit/validators/key_test.rb#L21-L27

so I assume modifying it would be a breaking change.

jlledom:
> I found it easier to compare the before and after behavior in a spreadsheet (I am a bit lost in tests): https://docs.google.com/spreadsheets/d/13FKZn_5kaXp8eA-GOT5OQEPAr-W4QqaYFeTK8semvmo/edit?gid=0#gid=0
> 
> I'll summarize here:
> 
> for App ID / App Key mode: "valid user_key + valid app_key" combination used to be returning 200 OK, but now returns:
> 
> ```
> 409
> <status>
>     <authorized>false</authorized>
>     <reason>application key "955112421a0ea5c563663659890e96f2" is invalid</reason>
>     <plan>aaa</plan>
> </status>
> ```
> 
> I think this is misleading, because 1. the app key is actually valid, so the message is not accurate. 2. the actual problem is that `app_id` is not provided, while expected in this mode, so I think this is what the error/reason should say. In fact, I think it even makes more sense to return a 403 with something like
> 
> ```
> <error code="app_id_missing">application id is missing</error>
> ```
> 
> Because I think 409 with `status` assumes that we have identified the application, but for some reason couldn't authorize. But 403 means that we couldn't even identify the app. I think the latter is more suitable, because we are expecting `app_id` in this mode, and `user_key` should be irrelevant, so I think we shouldn't even use it for identifying the application?
> 
> Not sure though...

A few comments:

I refactored the PR completely. Before I was modifying the `Key` validator to implement the solution but now I moved all the logic to the `validate` method in the transactor. The goal was to return more meaningful messages and error codes, according to your comments. Unfortunately it's not easy because the logic for errors is spread between the [`Key`](https://github.com/3scale/apisonator/blob/ed5cbe80bd58a0bab41c676e92b0b28bbe4601ba/lib/3scale/backend/validators/key.rb#L5-L13) validator and [`Application#with_app_id_from_params`](https://github.com/3scale/apisonator/blob/9d00930d150c42e2acff2c0ae5980ec91225434f/lib/3scale/backend/application.rb#L152-L163). Due to that, there are inconsistencies in return codes.

Every `Error` has it's default error code, which is `403` [by default](https://github.com/3scale/apisonator/blob/ed5cbe80bd58a0bab41c676e92b0b28bbe4601ba/lib/3scale/backend/errors.rb#L16-L18), but for instance, `ApplicationNotFound` [sets it](https://github.com/3scale/apisonator/blob/ed5cbe80bd58a0bab41c676e92b0b28bbe4601ba/lib/3scale/backend/errors.rb#L46-L48) to `404`. On the other hand, validators [call](https://github.com/3scale/apisonator/blob/ed5cbe80bd58a0bab41c676e92b0b28bbe4601ba/lib/3scale/backend/validators/key.rb#L11) the `fail!` which always returns `409`. No matter which error you pass to `fail!`, the error code is always overwritten by `409`.

Also, apparently only error codes `200` and `409` are [expected](https://github.com/3scale/apisonator/blob/670b2dbdf4520e2057038013c8364e689659104d/lib/3scale/backend/listener.rb#L299) from an `auth` endpoint. But when an exception happens, the response code associated to the `Error` is returned, which seems contradictory to me.

To make it even more confusing, there's an `ErrorStorage` [call](https://github.com/3scale/apisonator/blob/670b2dbdf4520e2057038013c8364e689659104d/lib/3scale/backend/listener.rb#L76) which always stores `403` to matter the original error code.

These are the error codes as defined before this PR:
- `403`: User key invalid
- `404`: App id invalid, because application was not found
- `403`: Either user key or app id but not both
- `409`: App key invalid, because it's the validator who checks this, so it's treated as a validator error.

After my last commit, I'm returning the same errors, however, there might be some breaking changes, since the behavior changed. I updated the doc to summarize the new behavior:

https://docs.google.com/spreadsheets/d/13FKZn_5kaXp8eA-GOT5OQEPAr-W4QqaYFeTK8semvmo/edit?gid=0#gid=0

Changes:
- In previous commits in this PR, invalid user key was being returned as `403` or `409` depending on the error being raised by the validator or not. Not it's always `403`
- In master and in previous commits, invalid app id was never being returned, now it's returned as `404`.

Invalid app key is still being returned as `409`.

I know this is pretty confusing but I tried to fix the bug and keep the breaking changes as minimal as possible. If I'm not wrong, I think I only changed the response codes for the invalid combinations like `backend_version 1 + app_id` and `backend_version 2 + user_key`

akostadinov:
> Also, apparently only error codes 200 and 409 are [expected](https://github.com/3scale/apisonator/blob/670b2dbdf4520e2057038013c8364e689659104d/lib/3scale/backend/listener.rb#L299) from an auth endpoint.

Not sure this is exactly like this. I would say that in normal situation these are the two codes returned. But it doesn't mean clients expect only these two. There might be other error conditions.

I'm not concerned about backward compatibility. Clients should expect 200 to be success and anything else to be a failure.

akostadinov:
So use the design that you like the most.

lib/3scale/backend/transactor.rb
```diff
@@ -66,29 +66,25 @@ def validate(oauth, provider_key, report_usage, params, request_info)
...
-        validators = if oidc_service
-                       Validators::OIDC_VALIDATORS
-                     elsif oauth
-                       Validators::OAUTH_VALIDATORS
-                     else
-                       Validators::VALIDATORS
-                     end
+        raise AuthenticationError if app_id && user_key
+
+        if service.backend_version.to_i == 1
```
akostadinov, line 82: I think there was a quesiton already, shouldn't we fail the request if version is not `1` or `2`? Now we differenciate between `1` and whatever else.

akostadinov (approved):
Looks good, so can merge as far as I'm concerned if you don't like my suggestions.

lib/3scale/backend/transactor.rb
jlledom: Not in this method, because here `backend_version` can also be `oauth`:

https://github.com/3scale/apisonator/blob/ed5cbe80bd58a0bab41c676e92b0b28bbe4601ba/lib/3scale/backend/transactor.rb#L64

Expected values here are empty, `nil`, `1`, `2` and `oauth`. Look at this comment https://github.com/3scale/apisonator/pull/428#discussion_r2191684480. As I interpreted that test, this means: regarding keys validation, when `backend_version` is `1` we authenticate by `user_key`,  in all other scenarios we authenticate by `app_id + app_key`.

lib/3scale/backend/errors.rb
```diff
@@ -72,7 +72,11 @@ def initialize(id, user_key)
 
     class ApplicationNotFound < NotFound
       def initialize(id = nil)
-        super %(application with id="#{id}" was not found)
+        if id.blank?
+          super 'application id is missing'.freeze
+        else
+          super %(application with id="#{id}" was not found)
+        end
```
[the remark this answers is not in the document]
jlledom: Yes, that's because of copy-paste :). Fixed: https://github.com/3scale/apisonator/pull/428/commits/6111bf70d080b5bf1cabbfe172191f3a82831d9b

mayorova (approved):
I think the new behavior makes much more sense! :+1: 

What we would need though is to write decent release notes to inform about the change in the next release, and clarify the previous and the new behavior (probably we can use that google spreadsheet to help us).

PR #430, opened 2025-08-08
More fixes to deploy

jlledom:
After https://github.com/3scale/apisonator/pull/429, alpha still fails to deploy, the error is:

```
Permission denied - bind(2) for /opt/ruby/apisonator-3.4.3/supervisor.ipc
```

That's because the image is created to run with the user `1001` and grants write permissions to that user, but Openshift ignores that and uses another used instead, which doesn't have the write permissions.

In order to solve that, I'm making some changes in the falcon config file to install its temporary files under `/tmp` which should have the proper permissions.

On the other hand, I found a way to move the falcon config file to under `config/` without the monkey patch, just calling the `rackup_path` method.

### DIFF
--- a/apisonator.gemspec
+++ b/apisonator.gemspec
@@ -27,7 +27,6 @@ Gem::Specification.new do |s|
   s.files << 'CHANGELOG.md'
   s.files << 'Rakefile'
   s.files << 'config.ru'
-  s.files << 'falcon_config.rb'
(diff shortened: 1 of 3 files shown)


akostadinov (approved):
This now looks like the sunk cost fallacy to make `falcon host` work, this is how scams work 👼

PR #436, opened 2025-09-17
simplify Dockerfile

akostadinov:
Tried to simplify the Dockerfile to avoid issues running on latest Fedora 42.

```
🐚 podman run --rm -it --entrypoint=/bin/bash quay.io/3scale/apisonator
Error: creating container 3734a959f1fadf280ac561ba32099e50f3230bef2f30871d1d03a1805164012d workdir: mkdir /home/akostadi/.local/share/containers/storage/overlay/99149e0b7a284c2017042e1924fc05c57fcc6ef285cca20c1d85e9e4fedfc9e3/merged/opt/app: file exists
```

Works for me locally. Should work with porta-dev-tools too.

### DIFF
--- a/openshift/distro/ubi/9/Dockerfile
+++ b/openshift/distro/ubi/9/Dockerfile
@@ -6,10 +6,6 @@
 #
 # - RUBY_VERSION: Ruby version used.
 # - BUILD_DEPS: Packages needed to build/install the project.
-# - BUNDLE_VERSION_MATCH: Install the Bundler version used by the lockfile
-#                         instead of using the SCL version.
-# - BUNDLE_GEMFILE: Gemfile name to pin Bundler to.
-# - BUNDLE_WITHOUT: List of Bundler groups to skip.
 # - PUMA_WORKERS: Default number of Puma workers to serve the app.
 #
 # Profiles you should use:
@@ -19,9 +15,6 @@
 #
 # - Development/Test: just use default values.
 # - Release:
-#   - GEM_UPDATE: false
-#   - BUNDLE_VERSION_MATCH: false
-#   - BUNDLE_WITHOUT: development:test
 #   - PUMA_WORKERS: number of Puma worker processes - depends on intended usage,
 #                   with number of cpus being a good heuristic.
 #
@@ -32,7 +25,8 @@ ARG RUBY_VERSION="3.3"
 ARG BUILD_DEPS="tar make file findutils git patch gcc automake autoconf libtool redhat-rpm-config openssl-devel ruby-devel"
 
 RUN microdnf module enable -y "ruby:${RUBY_VERSION}" \
- && microdnf install -y --nodocs ${BUILD_DEPS}
+ && microdnf install -y --nodocs ${BUILD_DEPS} \
+ && microdnf clean all
 
 # Copy sources
 WORKDIR /opt/source
--- a/openshift/distro/ubi/9/Dockerfile
+++ b/openshift/distro/ubi/9/Dockerfile
@@ -38,41 +38,42 @@ RUN microdnf module enable -y "ruby:${RUBY_VERSION}" \
 WORKDIR /opt/source
 COPY ./ ./
 
-# Setup Bundler
+# setup system
+RUN cp -n openshift/3scale_backend.conf /etc/ \
+ && chmod 660 /etc/3scale_backend.conf \
+ && cp -n openshift/backend-cron /usr/local/sbin/backend-cron \
+ && mkdir -p -m 0770 /var/run/3scale/ \
+ && mkdir -p -m 0770 /var/log/backend/ \
+ && touch /var/log/backend/3scale_backend{,_worker}.log \
+ && chmod g+rw /var/log/backend/3scale_backend{,_worker}.log \
+ && chown -R 1001 . && chmod g+w /opt
+
+# Install bundler globally (as root)
 RUN BUNDLED_WITH=$(awk '/BUNDLED WITH/ { getline; print $1 }' Gemfile.lock) \
  && gem install --no-document bundler:$BUNDLED_WITH \
  && echo Using $(bundle --version), originally bundled with ${BUNDLED_WITH}
 
+USER 1001
+
 # Builds a clean source tree and deploys it with Bundler.
-RUN cp -n openshift/3scale_backend.conf /etc/ \
- && chmod 644 /etc/3scale_backend.conf \
- && BACKEND_VERSION=$(gem build apisonator.gemspec | \
+RUN BACKEND_VERSION=$(gem build apisonator.gemspec | \
       sed -n -e 's/^\s*Version\:\s*\([^[:space:]]*\)$/\1/p') \
- && gem unpack "apisonator-${BACKEND_VERSION}.gem" --target=/opt/ruby \
- && mv "/opt/ruby/apisonator-${BACKEND_VERSION}" /opt/app \
- && rm -rf /opt/ruby
+ && gem unpack "apisonator-${BACKEND_VERSION}.gem" --target=/tmp/ruby \
+ && mv "/tmp/ruby/apisonator-${BACKEND_VERSION}" /opt/app \
+ && rm -rf /tmp/ruby
 
 WORKDIR /opt/app
 
 # Install ruby dependencies and put things at the right places
 RUN bundle config set --local build.nokogiri --use-system-libraries \
- && bundle config --local without development:test \
+ && bundle config set --local without development:test \
  && bundle config set --local deployment true \
- && bundle config --local path /opt/app \
  && bundle install --jobs $(grep -c processor /proc/cpuinfo) \
  && cp /opt/source/openshift/config/puma.rb config/ \
- && cp -n /opt/source/openshift/backend-cron /usr/local/sbin/backend-cron \
  && cp -n /opt/source/openshift/entrypoint.sh ./ \
- && rm -rf /opt/source \
- && mkdir -p -m 0770 /var/run/3scale/ \
- && mkdir -p -m 0770 /var/log/backend/ \
- && touch /var/log/backend/3scale_backend{,_worker}.log \
- && chmod g+rw /var/log/backend/3scale_backend{,_worker}.log \
- && chmod -R g+w .
+ && rm -rf /opt/source
 
 EXPOSE 3000
-
-USER 1001
 
 ARG PUMA_WORKERS=1
 ARG RACK_ENV=production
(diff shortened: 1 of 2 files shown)


openshift/distro/ubi/9/Dockerfile
```diff
@@ -28,80 +28,53 @@
...
- && cd "/opt/ruby/apisonator-${BACKEND_VERSION}" \
- && cp -a ${HOME}/app/.bundle "/opt/ruby/apisonator-${BACKEND_VERSION}/" \
- && bundle config --local path "/opt/ruby/apisonator-${BACKEND_VERSION}/" \
- && bundle install --jobs $(grep -c processor /proc/cpuinfo) \
- && ln -s ${PWD} /opt/app \
- && cp ${HOME}/app/openshift/config/puma.rb ./config/ \
- && cp -n ${HOME}/app/openshift/backend-cron /usr/local/sbin/backend-cron \
- && cp -n ${HOME}/app/openshift/entrypoint.sh ./ \
- && rm -rf ${HOME}/app \
+ && chmod 660 /etc/3scale_backend.conf \
```
mayorova, line 43: I am getting a 
```
cp: cannot stat 'openshift/3scale_backend.conf': No such file or directory
```
around here :thinking:

openshift/distro/ubi/9/Dockerfile
```diff
@@ -28,80 +28,53 @@
...
- && bundle config --local disable_shared_gems 1 \
- && bundle config --local without ${BUNDLE_WITHOUT} \
- && bundle config --local gemfile ${BUNDLE_GEMFILE}
+RUN microdnf module enable -y "ruby:${RUBY_VERSION}" \
+ && microdnf install -y --nodocs ${BUILD_DEPS}
 
 # Copy sources
-COPY --chown=1001:1001 ./ "${HOME}/app/"
+WORKDIR /opt/source
+COPY ./ ./
```
jlledom, line 39: This line is hard to understand, it works because the `Makefile` arranges everything and at this point we are under a directory called `os-build` which is a symbolic link to a temporary directory where all the project sources are copied.

I think the comment above should mention that

openshift/distro/ubi/9/Dockerfile
jlledom: Did you try to build it with `make build`?. That's required because the `Makefile` arranges directories to work.

openshift/distro/ubi/9/Dockerfile
```diff
@@ -28,80 +28,53 @@
...
-    else \
-      gem install -N bundler --version "${BUNDLE_VERSION}"; \
-    fi \
- && echo Using $(bundle --version), originally bundled with ${BUNDLED_WITH} \
- && bundle config --local silence_root_warning 1 \
- && bundle config --local disable_shared_gems 1 \
- && bundle config --local without ${BUNDLE_WITHOUT} \
- && bundle config --local gemfile ${BUNDLE_GEMFILE}
+RUN microdnf module enable -y "ruby:${RUBY_VERSION}" \
+ && microdnf install -y --nodocs ${BUILD_DEPS}
```
jlledom, line 35: Maybe add:
```
microdnf clean all
```

So all downloaded packages aren't included in the FS layer

openshift/distro/ubi/9/Dockerfile
```diff
@@ -28,80 +28,53 @@
...
 WORKDIR /opt/app
 
+# Install ruby dependencies and put things at the right places
+RUN bundle config set --local build.nokogiri --use-system-libraries \
+ && bundle config set --local without development:test \
+ && bundle config set --local deployment true \
+ && bundle install --jobs $(grep -c processor /proc/cpuinfo) \
+ && cp /opt/source/openshift/config/puma.rb config/ \
+ && cp -n /opt/source/openshift/entrypoint.sh ./ \
+ && rm -rf /opt/source
```
jlledom, line 74: This will make `/opt/source` unavailable in the final image, but the folder and its contents will continue to exist in previous layer and contribute to the final image size. The solution would be using a multi-stage dockerfile, but I think it's not worth it since this is a development dockerfile and probably `/opt/source` won't take too much space anyway.

openshift/distro/ubi/9/Dockerfile
jlledom, line 43: Permissions where `644` before and are `660` now, I think they could be `640` or even `440` since we don't need to write there. Just a nitpick, not really important, though.

openshift/distro/ubi/9/Dockerfile
```diff
@@ -28,80 +28,53 @@
...
- && if test "${BUNDLE_VERSION_MATCH}x" = "truex"; then \
-      gem install -N bundler --version "${BUNDLED_WITH}"; \
-    else \
-      gem install -N bundler --version "${BUNDLE_VERSION}"; \
-    fi \
- && echo Using $(bundle --version), originally bundled with ${BUNDLED_WITH} \
- && bundle config --local silence_root_warning 1 \
- && bundle config --local disable_shared_gems 1 \
- && bundle config --local without ${BUNDLE_WITHOUT} \
- && bundle config --local gemfile ${BUNDLE_GEMFILE}
```
jlledom, line 68: Some variables are still in the header comment but they doesn't exist anymore. They should probably be remove from the header:

- `BUNDLE_GEMFILE`
- `BUNDLE_WITHOUT`
- `BUNDLE_VERSION_MATCH`
- `GEM_UPDATE`

On the other hand, some of those variables are set in the `Makefile`, so they can probably be removed from there as well.

openshift/distro/ubi/9/Dockerfile
```diff
@@ -28,80 +28,53 @@
...
+ && chmod 660 /etc/3scale_backend.conf \
+ && cp -n openshift/backend-cron /usr/local/sbin/backend-cron \
  && mkdir -p -m 0770 /var/run/3scale/ \
  && mkdir -p -m 0770 /var/log/backend/ \
  && touch /var/log/backend/3scale_backend{,_worker}.log \
- && chmod g+rw /var/log/backend/3scale_backend{,_worker}.log
-
-# Ensure safe defaults within the container for general case as well as to cover OpenShift specific case where the unprivileged user who runs the container is member of the root Linux group
-RUN chown -R 1001:0 /opt/ruby \
-    && chmod -R 750 /opt/ruby
```
jlledom, line 96: I had to add this because something failed, don't remember what.
Just to confirm, is `chown -R 1001 . && chmod g+w /opt` equivalent?

jlledom (commented):
I tested this locally and works fine for me. These are the commands I use to launch the containers

#### Sync
```
podman run --rm --name apisonator --workdir=/opt/app --rm --network=host -it -e CONFIG_REDIS_ASYNC=0 -e CONFIG_INTERNAL_API_PASSWORD=password -e CONFIG_INTERNAL_API_USER=system-app -e CONFIG_QUEUES_MASTER_NAME=redis://localhost:6379/5 -e CONFIG_REDIS_PROXY=redis://localhost:6379/6 -e RACK_ENV=production 4d9e7a917089 bin/3scale_backend -s puma start -e production -p 3001 -x /dev/stdout
```
```
podman run --rm --name apisonator-worker --workdir=/opt/app --rm --network=host -it -e CONFIG_REDIS_ASYNC=0 -e CONFIG_INTERNAL_API_PASSWORD=password -e CONFIG_INTERNAL_API_USER=system-app -e CONFIG_QUEUES_MASTER_NAME=redis://localhost:6379/5 -e CONFIG_REDIS_PROXY=redis://localhost:6379/6 -e RACK_ENV=production 4d9e7a917089 bin/3scale_backend_worker run
```

#### Async
```
podman run --rm --name apisonator --workdir=/opt/app --rm --network=host -it -e CONFIG_REDIS_ASYNC=1 -e CONFIG_INTERNAL_API_PASSWORD=password -e CONFIG_INTERNAL_API_USER=system-app -e CONFIG_QUEUES_MASTER_NAME=redis://localhost:6379/5 -e CONFIG_REDIS_PROXY=redis://localhost:6379/6 -e CONFIG_QUEUES_MAX_CONNS=15 -e CONFIG_REDIS_MAX_CONNS=15 -e RACK_ENV=production 4d9e7a917089 bin/3scale_backend -s falcon start -e production -p 3001 -x /dev/stdout
```
```
podman run --rm --name apisonator-worker --workdir=/opt/app --rm --network=host -it -e CONFIG_REDIS_ASYNC=1 -e CONFIG_INTERNAL_API_PASSWORD=password -e CONFIG_INTERNAL_API_USER=system-app -e CONFIG_QUEUES_MASTER_NAME=redis://localhost:6379/5 -e CONFIG_REDIS_PROXY=redis://localhost:6379/6 -e CONFIG_QUEUES_MAX_CONNS=15 -e CONFIG_REDIS_MAX_CONNS=15 -e RACK_ENV=production 4d9e7a917089 bin/3scale_backend_worker run
```

openshift/distro/ubi/9/Dockerfile
akostadinov: The issue is not rearranging files or some other Makefile magic. The issue is that the Dockerfile does not live in the project home directory as usual but in some subdirectory. I build the image just with podman like:
```
podman build --squash -f openshift/distro/ubi/9/Dockerfile -t -t quay.io/3scale/apisonator .
```

The `.` at the end is setting the context of the build that defaults to the PATH of the dockerfile which makes no sense in this case.

But I didn't want to move the Dockerfile in this pull request. 

@mayorova , probably this is the issue you faced and it has nothing to do with this PR.

openshift/distro/ubi/9/Dockerfile
akostadinov: if the config file is writable, then it would be possible to run the container and modify config before startup for debugging purposes. I think this might be valuable for debugging. Also I don't see any downsides in having it writable.

akostadinov: See also my comment below. https://github.com/3scale/apisonator/pull/436#discussion_r2362968382

openshift/distro/ubi/9/Dockerfile
akostadinov: updated, good point

openshift/distro/ubi/9/Dockerfile
akostadinov: Good housekeeping, removed all irrelevant references to these libraries.

openshift/distro/ubi/9/Dockerfile
jlledom: Fine, then it would be good to remove it, probably in another PR

akostadinov:
According to my local testing, apisonator listener and worker still operate properly. And I see no errors in the logs. So I'm merging this one.

PR #457, opened 2026-02-05
THREESCALE-12222: Invoke stats cleaner with no connection URLs by default

mayorova:
Fixes https://issues.redhat.com/browse/THREESCALE-12222

To test a specific rspec locally with twemproxy, here are the steps:

1. Copy the twemproxy file:
```
cp script/config/examples/twemproxy.yml script/config/
```
2. Stop the currently running redis (to free the port `6379`)
3. Run podman-compose with twemproxy:
```
podman-compose -f script/config/podman-compose.yml up redis-master twemproxy redis-shard1 redis-shard2 redis-shard3
```
4. Add the following to `.env.test`:
```
CONFIG_LOG_PATH=/dev/stdout

CONFIG_QUEUES_READ_TIMEOUT=10
CONFIG_QUEUES_WRITE_TIMEOUT=10
CONFIG_QUEUES_MASTER_NAME=redis://localhost:6379
CONFIG_REDIS_READ_TIMEOUT=10
CONFIG_REDIS_WRITE_TIMEOUT=10
CONFIG_REDIS_PROXY=redis://localhost:22121
```
5. Run a specific test in sync and async modes:
```
CONFIG_REDIS_ASYNC=true bundle exec rake "spec:specific[spec/unit/stats/cleaner_spec.rb]"
CONFIG_REDIS_ASYNC=false bundle exec rake "spec:specific[spec/unit/stats/cleaner_spec.rb]"
```

### DIFF
--- a/lib/3scale/backend/stats/cleaner.rb
+++ b/lib/3scale/backend/stats/cleaner.rb
@@ -80,25 +80,15 @@ module ThreeScale
           # @param [Array] redis_conns Instantiated Redis clients.
           # @param [IO] log_deleted_keys IO where to write the logs. Defaults to
           #             nil (logs nothing).
-          def delete!(redis_conns, log_deleted_keys: nil)
+          def delete!(redis_conns: [], log_deleted_keys: nil)
             services = services_to_delete
             logger.info("Going to delete the stats keys for these services: #{services.to_a}")
 
             unless services.empty?
-              _ok, failed = redis_conns.partition do |redis_conn|
-                begin
-                  delete_keys(redis_conn, services, log_deleted_keys)
-                  true
-                rescue => e
-                  handle_redis_exception(e, redis_conn)
-                  false
-                end
-              end
-
-              with_retries { remove_services_from_delete_set(services) } if failed.empty?
-
-              failed.each do |failed_conn|
-                logger.error("Error while deleting stats of server #{failed_conn}")
+              if redis_conns.empty?
+                delete_keys(Storage.instance, services, log_deleted_keys)
+              else
+                delete_from_server_list services, redis_conns, log_deleted_keys
               end
             end
 
@@ -136,6 +126,27 @@ module ThreeScale
           end
 
           private
+
+          # Legacy method that works for a list of individual servers, can be used with
+          # Twemproxy, but does not support such Redis connection options as username/password, TLS etc.
+          def delete_from_server_list(services, redis_conns, log_deleted_keys)
+            _ok, failed = redis_conns.partition do |redis_conn|
+              begin
+                delete_keys redis_conn, services, log_deleted_keys
+                true
+              rescue => e
+                handle_redis_exception(e, redis_conn)
+                false
+              end
+            end
+
+            with_retries { remove_services_from_delete_set(services) } if failed.empty?
+
+            failed.each do |failed_conn|
+              logger.error("Error while deleting stats of server #{failed_conn}")
+            end
+          end
+
 
           def handle_redis_exception(exception, redis_conn)
             # If it's a connection error, do nothing so we can continue with
--- a/lib/3scale/backend/stats/cleaner.rb
+++ b/lib/3scale/backend/stats/cleaner.rb
@@ -86,7 +86,7 @@ module ThreeScale
 
             unless services.empty?
               if redis_conns.empty?
-                delete_keys(Storage.instance, services, log_deleted_keys)
+                delete_from_storage services, log_deleted_keys
               else
                 delete_from_server_list services, redis_conns, log_deleted_keys
               end
@@ -126,6 +126,19 @@ module ThreeScale
           end
 
           private
+
+          # Delete the stags keys from the default storage Redis instance
+          def delete_from_storage(services, log_deleted_keys)
+            redis = Storage.instance
+
+            begin
+              delete_keys(redis, services, log_deleted_keys)
+              with_retries { remove_services_from_delete_set(services) }
+            rescue => e
+              handle_redis_exception(e, redis)
+              logger.error("Error while deleting stats from storage: #{redis}")
+            end
+          end
 
           # Legacy method that works for a list of individual servers, can be used with
           # Twemproxy, but does not support such Redis connection options as username/password, TLS etc.
--- a/lib/3scale/backend/stats/cleaner.rb
+++ b/lib/3scale/backend/stats/cleaner.rb
@@ -62,11 +62,13 @@ module ThreeScale
           # Deletes all the stats for the services that have been marked for
           # deletion.
           #
-          # This method receives a collection of instantiated Redis clients.
+          # This method optionally receives a collection of instantiated Redis clients.
           # Those clients need to connect to Redis servers directly. They cannot
           # connect to a proxy like Twemproxy. The reason is that this function
           # needs to scan the database using the "SCAN" command, which is not
           # supported by Twemproxy.
+          #
+          # If the list of servers is not provided, the standard Storage instance is used
           #
           # The services marked as deletion will be marked as done only when
           # this function finishes deleting the keys from all the Redis servers.
@@ -110,18 +112,12 @@ module ThreeScale
           # @param [IO] log_deleted_keys IO where to write the logs. Defaults to
           #             nil (logs nothing).
           def delete_stats_keys_set_to_0(redis_conns, log_deleted_keys: nil)
-            _ok, failed = redis_conns.partition do |redis_conn|
+            redis_conns.each do |redis_conn|
               begin
                 delete_stats_keys_with_val_0(redis_conn, log_deleted_keys)
-                true
               rescue => e
                 handle_redis_exception(e, redis_conn)
-                false
               end
-            end
-
-            failed.each do |failed_conn|
-              logger.error("Error while deleting stats of server #{failed_conn}")
             end
           end
 
@@ -136,7 +132,6 @@ module ThreeScale
               with_retries { remove_services_from_delete_set(services) }
             rescue => e
               handle_redis_exception(e, redis)
-              logger.error("Error while deleting stats from storage: #{redis}")
             end
           end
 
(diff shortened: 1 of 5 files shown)


lib/3scale/backend/stats/cleaner.rb
```diff
@@ -137,6 +127,40 @@ def delete_stats_keys_set_to_0(redis_conns, log_deleted_keys: nil)
...
+                handle_redis_exception(e, redis_conn)
+                false
+              end
+            end
+
+            with_retries { remove_services_from_delete_set(services) } if failed.empty?
+
+            failed.each do |failed_conn|
+              logger.error("Error while deleting stats of server #{failed_conn}")
+            end
```
akostadinov: makes a little more sense to log within `handle_redis_exception` but just  a nitpick. Or if there is too much other log, then maybe it's good to have it at the end. So up to you.

akostadinov (dismissed):
looks awesome!

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -80,25 +80,15 @@ def mark_service_to_be_deleted(service_id)
           # @param [Array] redis_conns Instantiated Redis clients.
           # @param [IO] log_deleted_keys IO where to write the logs. Defaults to
           #             nil (logs nothing).
-          def delete!(redis_conns, log_deleted_keys: nil)
+          def delete!(redis_conns: [], log_deleted_keys: nil)
```
jlledom, line 86: Converting positional argument to keyword argument. This is a breaking change, it's only called from the rake task, right?

jlledom (approved):

lib/3scale/backend/stats/cleaner.rb
mayorova: That's right, it's not used from anywhere else. I kind of wanted to keep the possibility to pass the list of URLs, but avoid it by default, and passing `nil` in the first argument in the default scenario looked ugly.

jlledom (dismissed):

jlledom:
I don't know why now twemproxy tests fail :crying_cat_face:. That suite is always causing problems

akostadinov:
> I don't know why now twemproxy tests fail 😿. That suite is always causing problems

It's a crap. I wonder, can we increase log verbosity to see which test freezes?

jlledom:
> > I don't know why now twemproxy tests fail 😿. That suite is always causing problems
> 
> It's a crap. I wonder, can we increase log verbosity to see which test freezes?

Rather touching the code, opening a PR, etc. I think it's easier to just connect via SSH and run the tests with whatever config required. Probably log level can be set via some  env var.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -120,27 +113,72 @@ def delete!(redis_conns, log_deleted_keys: nil)
...
+              false
+            rescue => e
+              logger.error("Error #{e} occurred when calling SCAN command.")
+              false
             end
           end
 
-          private
+          # Delete the stats keys from the default storage Redis instance
+          def delete_from_storage(services, log_deleted_keys)
```
akostadinov: Can this method call `delete_from_server_list` or `delete_from_server_list` call this method? The logic seems redundant except for the iteration over multiple connections.

mayorova: I thought about it, but then the current approach seemed clearer to me.
I can refactor, if you feel strongly about it.

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -120,27 +113,72 @@ def delete!(redis_conns, log_deleted_keys: nil)
...
+              logger.error("Storage instance doesn't support SCAN command (e.g. Twemproxy is used). Cannot proceed with stats deletion.")
+              return
+            end
+
+            begin
+              delete_keys(redis, services, log_deleted_keys)
+              with_retries { remove_services_from_delete_set(services) }
+            rescue => e
+              handle_redis_exception(e, redis)
+            end
```
akostadinov: Not as simple as this?

```suggestion
  delete_from_server_list(services, [redis], log_deleted_keys)
```

mayorova: I am not sure it would be that simple :upside_down_face: but well...

lib/3scale/backend/stats/cleaner.rb
```diff
@@ -123,10 +124,32 @@ def delete_stats_keys_set_to_0(redis_conns, log_deleted_keys: nil)
...
+                true
+              end
+            rescue Timeout::Error
+              logger.error("Timeout when calling SCAN command.")
+              false
+            rescue => e
+              logger.error("Error #{e} occurred when calling SCAN command.")
+              false
+            end
+          end
```
jlledom, line 139: I have a few concerns. I don't like the idea of having to write such logic as a workaround for a command that is unsupported by twemproxy. This time is acceptable because it's just this case, but I think we could benefit on having something less ad-hoc.

We have a problem: when our cilents (sync and async) send an unsupported command to twemproxy, they just block instead of returning and exception. This happens because twemproxy in fact doesn't return an error, just closes the connection, which triggers the client reconnect logic, which retries forever, because we don't want the container to crash.

The problem is twemproxy closes the connection instead of just returning an error, and therefore we don't have a reliable way to distinguish between an unsupported command and a legit connection error. Which makes it harder for us to implement a decent unsupported command handling logic.

I suggest to add a `using_twemproxy?` method to the `Storage` module, as a class method, that checks the port to determine whether there is a twemproxy or not. And call that from the test suite and from here. In this module, we skip deletion if `using_twemproxy?` is `true`. WDYT? That would faster than having to wait for a 3 seconds timeout. We could call the same method from any other place if we face this problem again in the future.

On the other hand, note how the storage helper in tests already detects if twemproxy is being used and overwrites a few commands. We could call the `using_twemproxy?` method from there as well.

test/test_helpers/storage.rb
```diff
@@ -101,13 +101,13 @@ def method_missing(m, *args, **kwargs, &blk)
                 inner.send(m, *args, **kwargs, &blk)
               end
 
-              def respond_to_missing?(m)
-                inner.respond_to_missing? m
+              def respond_to_missing?(m, include_all=false)
+                inner.respond_to?(m, include_all)
```
jlledom, line 108: What's this for?

lib/3scale/backend/stats/cleaner.rb
mayorova: Well, I thought about having `using_twemproxy?` as a global method in `Storage`, however, the implementation seemed hacky to me. Can we really rely assume that twemproxy will always use port `22121`? If the customer decides to use another port, then the `SCAN` will still be attempted and will get stuck.
In tests, however, as it's a controlled environment, we can assume that.

test/test_helpers/storage.rb
mayorova: Well, `include_all` is the expected second argument of `respond_to_missing?`.

Without it, I was getting the error:
```
ArgumentError: wrong number of arguments (given 2, expected 1)
```

Some internals were probably calling this method with both arguments, but there was just one in the method definition.

Now, changing from `respond_to_missing?` to `repond_to?` was due to another error, a more cryptic one:
```
Unsupported command argument type: TrueClass
```

So, this is because apparently, when `respond_to_missing?` with its both arguments is being invoked on `inner` (aka Redis client), it was basically somehow passing the boolean value of `include_all` as argument of `SCAN` command.

But indeed I need to double-check how this behaves outside of tests... well, actually not, because this is only used in tests 🙃

lib/3scale/backend/stats/cleaner.rb
akostadinov: Checking the availability of a command makes more sense to me, it is more generic and is unique to the task currently being run. How it is done is another (worthy) topic. 

It is a good insight that the issue is re-connection logic. Can we disable retries for individual calls? That would make the check nicer. But even in current state, it is IMO acceptable.

test/test_helpers/storage.rb
akostadinov: Also if we override a built-in method, we need to follow the API definition. I would have given it more thought if it wasn't just old test code.

lib/3scale/backend/stats/cleaner.rb
jlledom: @mayorova 
> Can we really rely assume that twemproxy will always use port 22121? If the customer decides to use another port, then the SCAN will still be attempted and will get stuck.

Yeah, you're right.

@akostadinov 
> It is a good insight that the issue is re-connection logic. Can we disable retries for individual calls? That would make the check nicer.

Regrettably we hit the same wall every time: the sync mode. We could do whatever we want with the async client, but it must behave exactly the same way the sync client does, for compatibility. So we can't do it unless `redis-rb` provides a way to do it and we can mimic it. Having to maintain such compatibility is a heavy burden that makes the async client much more complicated than it should be.

jlledom (dismissed):

lib/3scale/backend/stats/cleaner.rb
mayorova: I applied the suggestion here: https://github.com/3scale/apisonator/pull/457/changes/1933b6d6470a5c996626d5e1663a66c58295b6b8

lib/3scale/backend/stats/cleaner.rb
mayorova: @jlledom I've moved the `using_twemproxy?` method to `TestHelpers::Storage::Mock`, and also reusing it for mock servers.

akostadinov: IMO testing the command is better :grimacing:

akostadinov (approved):

lib/3scale/backend/stats/cleaner.rb
mayorova: I'm not sure what you mean :smiling_face_with_tear: 
Are you suggesting to add a test for a test helper? :grimacing:

jlledom (approved):

lib/3scale/backend/stats/cleaner.rb
akostadinov: I meant testing whether SCAN command is supported better than trying to detect twemproxy

PR #462, opened 2026-04-22
THREESCALE-14619: Remove the invalid default value for CONFIG_LOG_PATH

mayorova:
Fixes https://redhat.atlassian.net/browse/THREESCALE-14619

`dev/stdout` is an invalid log path. Removing the setting results in using other defaults for the log, and in OpenShift environment it defaults to `STDOUT` anyway, so we don't need to set the variable.

### DIFF
--- a/openshift/backend-cron
+++ b/openshift/backend-cron
@@ -3,11 +3,6 @@
 DEFAULT_RESCHEDULE_JOBS_FREQ = 300
 DEFAULT_DELETE_STATS_FREQ = 60*60*24 # 1 day
 
-# The logger used in the delete stats code defaults to /tmp/backend_logger.log
-# When running this executable STDOUT is more convenient to check that
-# everything is working correctly.
-ENV['CONFIG_LOG_PATH'] ||= 'dev/stdout'
-
 def run_rake_task_in_thread(name, freq, check_once = false)
   Thread.new do
     loop do


jlledom (dismissed):

openshift/backend-cron
```diff
@@ -6,7 +6,7 @@ DEFAULT_DELETE_STATS_FREQ = 60*60*24 # 1 day
 # The logger used in the delete stats code defaults to /tmp/backend_logger.log
 # When running this executable STDOUT is more convenient to check that
 # everything is working correctly.
-ENV['CONFIG_LOG_PATH'] ||= 'dev/stdout'
+ENV['CONFIG_LOG_PATH'] ||= '/dev/stdout'
```
akostadinov: what will happen if we don't set this variable? Will it log to stdout by default? It is worth trying because this is how https://redhat.atlassian.net/browse/THREESCALE-12057 was fixed.

mayorova: Well, this is what the PR fixes - if the variable is NOT set, previously, the error from the description is printed.
With this updated default, the log seems to be printed to the stdout successfully. This is what the `backend-cron` pod log looks like when `CONFIG_LOG_PATH=/dev/stdout` is set on the pod explicitly:

```
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
I, [2026-04-22T07:49:17.829096 #21] INFO -- : Going to delete the stats keys for these services: ["5"]
I, [2026-04-22T07:49:17.832578 #21] INFO -- : Finished deleting the stats keys for these services: ["5"]
** Invoke reschedule_failed_jobs (first_time)
** Execute reschedule_failed_jobs
Rescheduled: 0. Failed and discarded: 0. Pending failed jobs: 0.
** Invoke stats:cleanup (first_time)
** Execute stats:cleanup
fatal: not a git repository (or any of the parent directories): .git
** Invoke reschedule_failed_jobs (first_time)
** Execute reschedule_failed_jobs
Rescheduled: 0. Failed and discarded: 0. Pending failed jobs: 0.
fatal: not a git repository (or any of the parent directories): .git
** Invoke reschedule_failed_jobs (first_time)
** Execute reschedule_failed_jobs
Rescheduled: 0. Failed and discarded: 0. Pending failed jobs: 0.
```

akostadinov: previously it was not unset but it was set to `dev/stdout` which expectedly cannot be found.

akostadinov:
```
Errno::ENOENT: No such file or directory @ rb_file_s_stat - dev/stdout (Errno::ENOENT)
```

This means it cannot find `dev/stdout`.

jlledom: I think the change is not wrong. According to this:

https://github.com/3scale/apisonator/blob/58baa5c01e667707c01cfd2435eded69878b4d20/lib/3scale/backend/configuration.rb#L99-L113

This change would end up actually logging to `/dev/stdout`. I think removing the whole line would lead to the same result anyway, but I don't have the energy to test it and confirm it right now. IMO we should just merge this, fix the immediate problem and maybe review the whole logging system later on.

akostadinov: Ok, I prefer not to have redundant settings in there. Also the variable doesn't seem to be used as intended as a PATH. But not super strong opinion. You can merge as you like it.

mayorova: @akostadinov is right, not setting the env var has the same effect - STDOUT is picked by default.

jlledom (approved):

PR #465, opened 2026-04-28
Fix circular dependency when Bugsnag is enabled

jlledom:
The pods at staging fail to start because a regression caused by https://github.com/3scale/apisonator/pull/452

```
lib/3scale/backend/logging/external/impl/bugsnag.rb:58:in `block in configure': uninitialized constant ThreeScale::Backend::Util (NameError)

                  config.project_root = Backend::Util.root_dir
```

`Util` depends on `Logging` which depends on `Util`

The fix is to remove the `require` causing the loop, it's redundant anyway.

### DIFF
--- a/lib/3scale/backend/util.rb
+++ b/lib/3scale/backend/util.rb
@@ -1,6 +1,5 @@
 require 'pathname'
 require 'etc'
-require '3scale/backend/logging'
 
 module ThreeScale
   module Backend


akostadinov (approved):
How did it ever work? Maybe different code paths need it?

Approving anyway.

jlledom:
> How did it ever work? Maybe different code paths need it?
> 
> Approving anyway.

It only fails when Bugsnag is enabled. We didn't try to deploy the regression to staging until now, and I'm not sure we have delivered this to on premisses, or if they use Bugsnag.

akostadinov:
btw would have been nice to have this tested :grimacing:

PR #467, opened 2026-05-04
THREESCALE-14671: Fix leaking connections to sentinels [2.16 hotfix]

jlledom:
**Jira Issue:**

https://redhat.atlassian.net/browse/THREESCALE-14671

**Description:**

Apisonator 2.16.3 is still leaking connections. That's because the hotfix I wrote for THREESCALE-11941, which was released in 2.16.3, fixes the connections leaking to the redis masters, but when connecting to sentinels, the connections to sentinels themselves are being leaked too.

The bug is in the methods to resolve the master and slave address from the sentinels client class. Those methods open a connection to redis but never close it. For some reason, those connections are not being being GCd when the methods return. According to Claude, because the calling `Task` is kept alive and the `Async` reactor holds a reference during all process life.

This PR fixes the bug by:

1. Caching the connections so they are reused
2. Ensuring the connections are closed by overriding the close method
3. Limiting sentinels connection pool to 1. This is redundant according to Claude, but still.

### DIFF
--- a/lib/3scale/backend/async_redis/sentinels_client_acl_tls.rb
+++ b/lib/3scale/backend/async_redis/sentinels_client_acl_tls.rb
@@ -15,16 +15,28 @@ module ThreeScale
           @role = config[:role] || :master
 
           @sentinel_credentials = config[:sentinel_username], config[:sentinel_password]
+          @sentinel_clients = {}
           @protocol = protocol
           @config = config
           @pool = connect(**options)
         end
 
+        def close
+          @sentinel_clients.each_value(&:close)
+          @sentinel_clients.clear
+          super
+        end
+
         private
+
+        def sentinel_client_for(sentinel_endpoint)
+          @sentinel_clients[sentinel_endpoint] ||=
+            Async::Redis::Client.new(sentinel_endpoint, protocol: Protocol::ExtendedRESP2.new(credentials: @sentinel_credentials), limit: 1)
+        end
 
         def resolve_master
           @sentinel_endpoints.each do |sentinel_endpoint|
-            client = Async::Redis::Client.new(sentinel_endpoint, protocol: Protocol::ExtendedRESP2.new(credentials: @sentinel_credentials))
+            client = sentinel_client_for(sentinel_endpoint)
 
             begin
               address = client.call('sentinel', 'get-master-addr-by-name', @master_name)
@@ -40,7 +52,7 @@ module ThreeScale
 
         def resolve_slave
           @sentinel_endpoints.each do |sentinel_endpoint|
-            client = Async::Redis::Client.new(sentinel_endpoint, protocol: Protocol::ExtendedRESP2.new(credentials: @sentinel_credentials))
+            client = sentinel_client_for(sentinel_endpoint)
 
             begin
               reply = client.call('sentinel', 'slaves', @master_name)
--- a/lib/3scale/backend/async_redis/sentinels_client_acl_tls.rb
+++ b/lib/3scale/backend/async_redis/sentinels_client_acl_tls.rb
@@ -22,7 +22,11 @@ module ThreeScale
         end
 
         def close
-          @sentinel_clients.each_value(&:close)
+          @sentinel_clients.each_value do |client|
+            client.close
+          rescue StandardError => error
+            Backend.logger.warn "Failed to close sentinel connection: #{error.message}"
+          end
           @sentinel_clients.clear
           super
         end
--- a/lib/3scale/backend/storage_async/client.rb
+++ b/lib/3scale/backend/storage_async/client.rb
@@ -72,16 +72,20 @@ module ThreeScale
 
         private
 
+        # Retries the block on connection errors without destroying the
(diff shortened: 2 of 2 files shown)


lib/3scale/backend/async_redis/sentinels_client_acl_tls.rb
```diff
@@ -15,16 +15,28 @@ def initialize(uri, protocol = Async::Redis::Protocol::RESP2, config, **options)
...
 
           @sentinel_credentials = config[:sentinel_username], config[:sentinel_password]
+          @sentinel_clients = {}
           @protocol = protocol
           @config = config
           @pool = connect(**options)
         end
 
+        def close
+          @sentinel_clients.each_value(&:close)
```
akostadinov: maybe `try(:close)`?

akostadinov: should also ignore exceptions

jlledom: I don't think whe need `try`, because we are sure the objects there respond to `:close`

I added a log trace for exceptions: https://github.com/3scale/apisonator/pull/467/commits/f36f936b91089e267c099412c207b78fce2f6d7f

jlledom:
@akostadinov I tried what you suggested: killing the master while performing the stress tests. It works fine, the sentinels failover to one of the replicas and the services is reestablished after a few seconds. I tried killing many redis servers one after another and it always recovers the connection.

However, I noticed the number of connections also peaked a lot under this circumstances. I wrote a fix as well: https://github.com/3scale/apisonator/pull/467/commits/6a52a2fd08c81ca7e991707a5e2099553011dc2f

akostadinov:
I'm concerned that this requires more thorough investigation.

First of all, given that you saw leaking connections despite the `close` implementation that you added, signifies that there is some issue in the underlying library related to closing connections.

The other thing that worries me is that you observed this behavior with a limited set of Redis master connections when using the Sentinel class. You observed that `AsyncRedis::Client.connect` would re-resolve master and perhaps clean previously established connections. (correct me if I'm wrong)

What if we have 10 for example with the non-sentinel class? Once a connection fails, will `#connect` clear all previous connections and re-establish new ones? Otherwise what would clean up existing connections?

At least to me things look super complicated and non-transparent.

I find it less risky to dig out and fix the close leak rather than changing the retry logic. Unless of course you have a deeper understanding of how things work than me, which likely is the case, so please bear with me!

jlledom:
> First of all, given that you saw leaking connections despite the `close` implementation that you added, signifies that there is some issue in the underlying library related to closing connections.

The close was a problem because 20 threads were creating and destroying the same shared `@redis_async` instance variable, one after another, since fibers are sequential. And after every creation, a new `SentinelsClientACLTLS` was being created, with as much connections as sentinels. Then the connections were leaked because the Async reactor prevents the GC to clean.

> What if we have 10 for example with the non-sentinel class? Once a connection fails, will `#connect` clear all previous connections and re-establish new ones? Otherwise what would clean up existing connections?

It did that before, now it reuses the same client object, and delegates the reconnection logic to the pool.

> I find it less risky to dig out and fix the close leak rather than changing the retry logic. Unless of course you have a deeper understanding of how things work than me, which likely is the case, so please bear with me!

I think it's fixed now. In any case, in much better than what we had before.

I tried a few scenarios: with and without TLS, with and without sentinels, it consistently works fine, recovers from outage and doesn't leak a connection.

jlledom:
@akostadinov also verified that replicas are always read-only and when a replica is promoted to master, it stops being read only, and the other way around.

lib/3scale/backend/storage_async/client.rb
```diff
@@ -72,16 +72,20 @@ def close
...
+        # client. The underlying pool already handles individual dead
+        # connections by retiring them and creating new ones via its
+        # constructor block (which re-resolves the master through
+        # sentinels). We just need to wait before retrying so we don't
+        # spin during a master failover.
         def with_reconnect
           attempt = 0
           begin
-           yield connect
+            yield connect
```
akostadinov, line 84: My question is how does the underlying pool handle dead connection. If the underlying pool has lets say 10 connections. This fiber hits a `CONNECTION_ERROR`. Then the underlying pool may see that, I assume and clear the connection.

But then how would it know whether the other 9 connections are dead or not? Is it possible that this same task then retries but with another dead connection? So for 10 attempts it is only trying the existing connections and never tries to create a new connection?

Idk if this is a realistic scenario. But this is what comes to mind when I see the `close` call gone.

jlledom: For each connection, the pool will try to send a request, fail, destroy the resource and return an error. If it had 10 connections, it will fail ten times. The  10 fibers will hit the retry code, reuse the same client, the same `SentinelsClientACLTLS` object, the same pool, and the pool will create a new resource or take any valid existing one, try to send a request... etc.

akostadinov (approved):

PR #468, opened 2026-05-06
THREESCALE-14798: return 400 for query parameter limit exceeded

jlledom:
When a POST request body exceeds Rack's 4096 parameter limit, Rack raises QueryLimitError. Previously this surfaced as a 500.

Two things caused the 500: first, the ExceptionCatcher middleware did not handle QueryLimitError, so it propagated unhandled. Second, when the logging middleware tried to log the error, its extract_query_string method called request.params which triggered the same QueryLimitError again, since it only rescued IOError.

Now the ExceptionCatcher returns 400 Bad Request with the error message in XML, and the logging middleware rescues the error gracefully when extracting query string parameters.

**Jira issue:**

https://redhat.atlassian.net/browse/THREESCALE-14798

### DIFF
--- a/lib/3scale/backend/rack/exception_catcher.rb
+++ b/lib/3scale/backend/rack/exception_catcher.rb
@@ -41,27 +41,19 @@ module ThreeScale
           delete_sinatra_error! env
           respond_with e.http_code, prepare_body(e.to_xml, env)
         rescue ArgumentError => e
-          return unhandled_exception(e) unless e.message.include?(INVALID_BYTE_SEQUENCE_ERR_MSG)
+          raise e unless e.message.include?(INVALID_BYTE_SEQUENCE_ERR_MSG)
 
+          error = Backend::NotValidData.new
           delete_sinatra_error! env
-          respond_with 400, Backend::NotValidData.new.to_xml
-        rescue Exception => e
-          unhandled_exception(e)
+          respond_with error.http_code, error.to_xml
+        rescue ::Rack::QueryParser::QueryLimitError => e
+          delete_sinatra_error! env
+
+          error = Backend::BadRequest.new(e.message)
+          respond_with error.http_code, error.to_xml
         end
 
         private
-
-        # raise up the chain an unhandled exception - but test first for an edge
-        # case with Rack key space limit for parsing requests.
-        # See https://github.com/rack/rack/blob/2.0.4/lib/rack/query_parser.rb#L166
-        def unhandled_exception(e)
-          if e.class == RangeError &&
-              e.message == 'exceeded available parameter key space'.freeze
-            respond_with 400, Backend::NotValidData.new.to_xml
-          else
-            raise e
-          end
-        end
 
         # Private: Deletes 'sinatra.error' key in Rack's env hash.
         # Some external services report 'sinatra.error' and we don't want it when
@@ -74,7 +66,7 @@ module ThreeScale
           env['sinatra.error'.freeze] = nil
         end
 
-        # Private: Prepares the body to include in the reponse.
+        # Private: Prepares the body to include in the response.
         #
         # body - Proposed body String.
         # env - The environment Hash.
--- a/test/integration/listener_test.rb
+++ b/test/integration/listener_test.rb
@@ -65,6 +65,43 @@ class ListenerTest < Test::Unit::TestCase
     assert_equal 'bad_request', node['code']
   end
 
+  def test_query_parameter_limit_exceeded
+    params = (0..2050).map { |i|
+      "transactions[#{i}][user_key]=uk&transactions[#{i}][usage][hits]=1"
+    }.join('&')
+
+    post '/transactions.xml', params
+
+    assert_equal 400, last_response.status
+
+    node = xml.at('error')
+    assert_match(/total number of query parameters \(\d+\) exceeds limit/, node.content)
+    assert_equal 'bad_request', node['code']
+  end
+
+  def test_query_bytesize_limit_exceeded
+    big_value = 'x' * (4 * 1024 * 1024 + 1)
+
+    post '/transactions.xml', "big=#{big_value}"
+
+    assert_equal 400, last_response.status
+
+    node = xml.at('error')
+    assert_match(/total query size exceeds limit/, node.content)
+
+    assert_equal 'bad_request', node['code']
+  end
+
+  def test_invalid_utf8_byte_sequence
+    get '/transactions/authorize.xml?service_token=foo&user_key=%ff%fe'
+
+    assert_equal 422, last_response.status
+
+    node = xml.at('error')
+    assert_match(/all data must be valid UTF8/, node.content)
+    assert_equal 'not_valid_data', node['code']
+  end
+
   def test_unsupported_end_users_auth
     get '/transactions/authorize.xml', provider_key: 'pk', user_id: '123'
     check_end_users_not_supported_error(last_response)
--- a/test/integration/logger_test.rb
+++ b/test/integration/logger_test.rb
@@ -22,6 +22,18 @@ class LoggerTest < Test::Unit::TestCase
     end
   end
 
+  test 'log does not raise when query parameter limit is exceeded' do
+    params = (0..2050).map { |i|
+      "transactions[#{i}][user_key]=uk&transactions[#{i}][usage][hits]=1"
+    }.join('&')
+
+    default_log_writer.any_instance.expects(:log).once
+
+    assert_nothing_raised do
+      post '/transactions.xml', params
+    end
+  end
+
   test 'log failed requests with request info and error info' do
     redis = Backend::Storage.instance
     redis.expects(:get).raises(Timeout::Error)
--- a/lib/3scale/backend/logging/middleware/writer.rb
+++ b/lib/3scale/backend/logging/middleware/writer.rb
@@ -109,7 +109,7 @@ module ThreeScale
             if env[REQUEST_METHOD].to_s.upcase == STR_POST
               provider_key = begin
                 ::Rack::Request.new(env).params[STR_PROVIDER_KEY]
-              rescue IOError
+              rescue IOError, ::Rack::QueryParser::QueryLimitError
                 # happens when body does not parse
                 nil
               end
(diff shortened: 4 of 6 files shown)


lib/3scale/backend/rack/exception_catcher.rb
```diff
@@ -58,6 +58,8 @@ def unhandled_exception(e)
           if e.class == RangeError &&
               e.message == 'exceeded available parameter key space'.freeze
             respond_with 400, Backend::NotValidData.new.to_xml
+          elsif e.is_a?(::Rack::QueryParser::QueryLimitError)
```
mayorova: I am curious, why have you decided to handle this here, rather than within `def call(env)` method.

You could add a `rescue ::Rack::QueryParser::QueryLimitError` line there instead.

I am also not sure why this `RangeError` is handled inside `#unhandled_exception` and not `#call`... But maybe this has to do with the fact that we need to check the exception message, which is not the case of `::Rack::QueryParser::QueryLimitError`.

I mean, not saying that it should be one way or another, just curious 🙃

jlledom: Yeah, I'm not sure, the `call` method seems to catch more general exceptions, and `unhandled_exception` looked to me like the place to enumerate all other, particular errors. Anyway, not a strong opinion. I'll move the catch to `call` if you want.

mayorova: Yeah, dunno... this `unhandled_exception` thing appeared as a result of this change: https://github.com/3scale/apisonator/commit/42f6a4ab994e6592baf80cb49c4428e7bc74268d

I think having the prcessing of this exception in `call` would be preferrable, because well... why would we deal with an "unhandled exception" if we can just handle it :upside_down_face:

akostadinov: Actually the whole thing should go. Agreed with Daria.

See inside `#call`
```ruby
        rescue ArgumentError => e
          return unhandled_exception(e) unless e.message.include?(INVALID_BYTE_SEQUENCE_ERR_MSG)

          delete_sinatra_error! env
          respond_with 400, Backend::NotValidData.new.to_xml
```

I think we should use the same pattern for the Range error (as well not reuse `Backend::NotValidData` because it is misleading for the Range error).

For `QueryLimitError` you convert to `Backend::BadRequest` but it is in fact a `Backend::Error` which is handled in a slightly different way in `#call`. Like there we also call `delete_sinatra_error! env`, here we don't. We need to handle in the same way I guess, ideally with the same code if not ugly.

Hope you make sense out of my ramble, or your AI assistant will 😬

jlledom: > I think we should use the same pattern for the Range error (as well not reuse `Backend::NotValidData` because it is misleading for the Range error).

`RangeError` + `"error message"` is the old pattern used in Rack 2.x, but that became dead code when we updated to Rack 3: https://github.com/rack/rack/pull/1757. The same scenario now raises this `QueryLimitError` we are catching.

> For `QueryLimitError` you convert to `Backend::BadRequest` but it is in fact a `Backend::Error` which is handled in a slightly different way in `#call`.

What do you mean? why is it a `Backend::Error`?

> Like there we also call `delete_sinatra_error! env`, here we don't. We need to handle in the same way I guess, ideally with the same code if not ugly.

I made some changes: https://github.com/3scale/apisonator/pull/468/commits/a5b05c33ab7154b6e98e0876418b3f6b6e440bcd

I removed the `unhandled_exception` method because it only raised the received error in practice, and we can do that directly, no need to call a method for that.

Also now we have A rescue clause for QueryLimitError which catches "too many parameters" and "too big query size" scenarios.

Also added a few tests to ensure the three scenarios the `call` method handles are properly handled. I expect them to return 400 and a meaningful response body in xml.

A part from the test suite, you can reproduce the failing scenarios with these commands:

**Params limit:**

```
python3 -c "print('&'.join(f'p{i}=x' for i in range(4097)), end='')" > tmp/THREESCALE-14798/post_body_error_params.txt

curl -s -X POST http://localhost:3001/transactions.xml --data-binary @tmp/THREESCALE-14798/post_body_error_params.txt
```

**Size limit:**

```
$ curl -s -X POST http://localhost:3001/transactions.xml --data-binary @/path/to/any/big/file
```

**Invalid UTF-8:**

```
curl -s "http://localhost:3001/transactions/authorize.xml?service_token=foo&user_key=$(printf '\xff\xfe')"
```

test/integration/listener_test.rb
```diff
@@ -65,6 +65,41 @@ def test_malformed_hash_param
...
+    params = (0..2050).map { |i|
+      "transactions[#{i}][user_key]=uk&transactions[#{i}][usage][hits]=1"
+    }.join('&')
+
+    post '/transactions.xml', params
+
+    assert_equal 400, last_response.status
+
+    node = xml.at('error')
+    assert_match(/exceeds limit/, node.content)
```
mayorova:
```suggestion
    assert_match(/total number of query parameters \(\d+\) exceeds limit/, node.content)
```

test/integration/listener_test.rb
```diff
@@ -65,6 +65,41 @@ def test_malformed_hash_param
...
+
+  def test_query_bytesize_limit_exceeded
+    big_value = 'x' * (4 * 1024 * 1024 + 1)
+
+    post '/transactions.xml', "big=#{big_value}"
+
+    assert_equal 400, last_response.status
+
+    node = xml.at('error')
+    assert_match(/exceeds limit/, node.content)
```
mayorova:
```suggestion
    assert_match(/total query size exceeds limit/, node.content)

```

test/integration/listener_test.rb
```diff
@@ -65,6 +65,41 @@ def test_malformed_hash_param
...
+    assert_equal 'bad_request', node['code']
+  end
+
+  def test_invalid_utf8_byte_sequence
+    get '/transactions/authorize.xml?service_token=foo&user_key=%ff%fe'
+
+    assert_equal 400, last_response.status
+
+    node = xml.at('error')
```
mayorova, line 100:
```suggestion
    node = xml.at('error')
    assert_match(/all data must be valid UTF8/, node.content)
```

test/integration/listener_test.rb
mayorova: Not critical, but I think it's nice to be able to distinguish error messages, otherwise, it might not catch the case where the exception is by error handled by an incorrect procedure.

lib/3scale/backend/rack/exception_catcher.rb
akostadinov: > What do you mean? why is it a Backend::Error?

It inherits from it. That's what I meant. I will look at the changes a little later.

jlledom: Ah, you mean `Backend::BadRequest`, which inherits from `Backend::Error`. Do you think we should make the error pass through that `prepare_body` method?

The `Backend::NotValidData` also inherits from `Backend::Error`, then it would go through `prepare_body` as well.

mayorova: Hm, now revisiting this after reading your discussion...
I noticed that the error classes have their own `http_code`.

Which makes me think that probably a better way to raise would be:
```
          error = Backend::NotValidData.new
          respond_with error.http_code, error.to_xml
```
and 
```
          error = Backend::BadRequest.new(e.message)
          respond_with error.http_code, error.to_xml
```

Which actually uncovers another thing - `NotValidData` is a child of `NotValidData` whose code is `422`, so probably sending `NotValidData` with a different code (`400` in this case) is a bit confusing?...

akostadinov: maybe our error inheritance is less than optimal

jlledom: @mayorova https://github.com/3scale/apisonator/pull/468/commits/f1e889ad9f3f3a55c3cb473ac85274290bf1eddb

test/integration/listener_test.rb
jlledom: This and others: https://github.com/3scale/apisonator/pull/468/commits/2ade5da06cf2bc9bed5fb6452379cfbb5479362b

lib/3scale/backend/rack/exception_catcher.rb
jlledom: Returning a proper code broke a few tests: https://github.com/3scale/apisonator/pull/468/commits/2156eac2ae55a34e87c810d0213c7878bc8063cd

This might be a breaking change, so it deserves a release note.

lib/3scale/backend/rack/exception_catcher.rb
```diff
@@ -41,28 +41,20 @@ def call(env)
           delete_sinatra_error! env
           respond_with e.http_code, prepare_body(e.to_xml, env)
         rescue ArgumentError => e
-          return unhandled_exception(e) unless e.message.include?(INVALID_BYTE_SEQUENCE_ERR_MSG)
+          raise e unless e.message.include?(INVALID_BYTE_SEQUENCE_ERR_MSG)
+          error = Backend::NotValidData.new
```
akostadinov: Idk why rubocop did not complain about this major violation
```suggestion

          error = Backend::NotValidData.new
```

test/integration/logger_test.rb
```diff
@@ -22,6 +22,18 @@ def default_log_writer
...
   end
 
+  test 'log does not raise when query parameter limit is exceeded' do
+    params = (0..2050).map { |i|
+      "transactions[#{i}][user_key]=uk&transactions[#{i}][usage][hits]=1"
+    }.join('&')
+
+    default_log_writer.any_instance.expects(:log).once
+
+    assert_nothing_raised do
```
akostadinov, line 32: 😿

akostadinov (dismissed):

test/integration/logger_test.rb
akostadinov: http://blog.zenspider.com/blog/2012/01/assert_nothing_tested.html

jlledom: What's your alternative? if I just remove the `assert_nothing_raised` call, what the post says is still valid: an empty controller would pass. But that would be in fact valid, because the test is called `"log does not raise when query parameter limit is exceeded"`... so as long as you call the endpoint, send a bunch of params and it doesn't raise, the test is passing. IMHO: best test ever.

akostadinov: You have an expectation. That's reasonable IMO.

akostadinov (approved):

PR #469, opened 2026-05-06
THREESCALE-14823: Fix async Redis client leak in retry logic

jlledom:
**Jira issue:**

https://redhat.atlassian.net/browse/THREESCALE-14823

**Description:**

After backporting the hotfix for **THREESCALE-14671**. I tested the same scenarios in the `master` branch and I found that one of the leaks is in fact reproducible also in `master`.

See the jira description for a more detailed explanation, but summarizing, these were the two leaking scenarios:

1. **Connections to sentinels being disposable:** This could be fine, but the GC was not working properly and the disposable clients were not being disposed, therefore, leaking the connections. The solution to this was caching and reusing the sentinels connections instead of making them disposable. In `master`, this is not a problem because we replaced our buggy module by the `async-redis` gem class, which already caches the clients.
2. **Retry logic:** The retry logic was buggy and multiple fibers failing at the same time were causing the same shared client `@redis_async` to be opened and closed multiple times. And leaking a connection each time, due to the GC problem. This happens also in `master`, and this PR fixes it by removing the calls to `close`. That way the same `@redis_async` instance is reused, and the same connection pools continue to allocate and discard connections properly without leaking. That is, the connection management gets delegated to the connection pools.

### DIFF
--- a/lib/3scale/backend/storage_async/client.rb
+++ b/lib/3scale/backend/storage_async/client.rb
@@ -74,16 +74,20 @@ module ThreeScale
 
         private
 
+        # Retries the block on connection errors without destroying the
+        # client. The underlying pool already handles individual dead
+        # connections by retiring them and creating new ones via its
+        # constructor block (which re-resolves the master through
+        # sentinels). We just need to wait before retrying so we don't
+        # spin during a master failover.
         def with_reconnect
           attempt = 0
           begin
-           yield connect
+            yield connect
           rescue *CONNECTION_ERRORS => e
-            close
-
             if attempt < @opts[:reconnect_attempts]
-              Backend.logger.warn "Failed Redis connect (attempt #{attempt+1}/#{@opts[:reconnect_attempts]}): #{e.message}"
               attempt += 1
+              Backend.logger.warn "Redis connection lost, reconnecting (attempt #{attempt}/#{@opts[:reconnect_attempts]}): #{e.message}"
               sleep @opts[:reconnect_wait_seconds]
               retry
             else
--- a/spec/unit/storage_async/client_spec.rb
+++ b/spec/unit/storage_async/client_spec.rb
@@ -1,0 +1,88 @@
+require 'spec_helper'
+
+if ThreeScale::Backend.configuration.redis.async
+  module ThreeScale
+    module Backend
+      module StorageAsync
+        describe Client do
+          let(:opts) do
+            {
+              url: 'redis://localhost:6379',
+              reconnect_attempts: 2,
+              reconnect_wait_seconds: 0,
+              max_connections: 10
+            }
+          end
+          let(:client) { described_class.new(opts) }
+          let(:fake_conn) { double('async_redis_client') }
+
+          before do
+            allow(AsyncRedis::Client).to receive(:connect).and_return(fake_conn)
+          end
+
+          describe 'with_reconnect' do
+            context 'when a connection error occurs and then recovers' do
+              it 'retries without replacing the underlying client' do
+                call_count = 0
+                allow(fake_conn).to receive(:call) do
+                  call_count += 1
+                  raise Errno::ECONNRESET if call_count == 1
+                  'PONG'
+                end
+
+                result = client.call('PING')
+
+                expect(result).to eq('PONG')
+                expect(AsyncRedis::Client).to have_received(:connect).once
+              end
+            end
+
+            context 'when all retry attempts are exhausted' do
+              it 'raises the connection error' do
+                allow(fake_conn).to receive(:call).and_raise(Errno::ECONNRESET)
+
+                expect { client.call('PING') }.to raise_error(Errno::ECONNRESET)
+              end
+
+              it 'does not replace the underlying client between attempts' do
+                allow(fake_conn).to receive(:call).and_raise(Errno::ECONNRESET)
+
+                client.call('PING') rescue nil
+
+                expect(AsyncRedis::Client).to have_received(:connect).once
+              end
+            end
+
+            context 'when multiple fibers fail concurrently' do
+              let(:concurrency) { 10 }
+
+              it 'reuses the same client instance across all fibers' do
+                failed_fibers = {}
+                allow(fake_conn).to receive(:call) do
+                  fiber = Fiber.current
+                  unless failed_fibers[fiber]
+                    failed_fibers[fiber] = true
+                    raise Errno::ECONNRESET
+                  end
+                  'PONG'
+                end
+
+                barrier = Async::Barrier.new
+                concurrency.times { barrier.async { client.call('PING') } }
+                barrier.wait
+
+                expect(AsyncRedis::Client).to have_received(:connect).once
+              end
+            end
+          end
+        end
+      end
+    end
+  end
+else
+  describe 'StorageAsync::Client' do
+    it 'is skipped when not running in async mode' do
+      skip 'This test only runs in async mode'
+    end
+  end
+end


lib/3scale/backend/storage_async/client.rb
```diff
@@ -74,16 +74,20 @@ def close
...
+        # constructor block (which re-resolves the master through
+        # sentinels). We just need to wait before retrying so we don't
+        # spin during a master failover.
         def with_reconnect
           attempt = 0
           begin
-           yield connect
+            yield connect
           rescue *CONNECTION_ERRORS => e
-            close
```
akostadinov, line 82: That's good. To make it right we will need more infrastructure to avoid closing each-others' connections. Delegating to the underlying pool is great in terms that it has to take care of that.

I'm wondering about this scenario:
* many workers process jobs
* jobs end and workers are only waiting for jobs
* all connections silently die
* now a single job comes in

What's gonna happen? Will it try N times with the existing connections and have N*wait_time delay before a new connection is created or will it happen somehow fast?

Other than that looks good.

P.S. the major issue with the original version is that we called `close` which would go for the instance variable `@redis_async` instead of the previously resolved `connect`, so we should have stored the return value of `connect` into a local variable and close it instead. But in case the aboe scenario works reasonably well, it is way simpler to let the underlying pool handle the situation.
P.P.S. the second major issue is that to clear the instance variable, we first need to be sure that `@redis_async == local_connection_variable` so that we don't clear the new connection from another fiber.

jlledom: > I'm wondering about this scenario:
> 
>     * many workers process jobs
> 
>     * jobs end and workers are only waiting for jobs
> 
>     * all connections silently die
> 
>     * now a single job comes in
> 
> 
> What's gonna happen? Will it try N times with the existing connections and have N*wait_time delay before a new connection is created or will it happen somehow fast?

All workers (different processes) will have one fiber waiting on `BLPOP`, that fiber will go through the usual retry cycle:

Send the command -> fail -> wait period -> try again

So as many parallel retry cycles as number of workers.

When a new job comes in, we assume the connection to Redis is reestablished, otherwise no job could have came in. In that case, all workers trying to call `BLPOP` will be back to normal after the retry cycle.

Each worker will have N fibers, and each fiber would be waiting on the call to `@jobs.pop`, not trying to query Redis at all.

One of the workers will get the job via `BLPOP`, it will push it to `@jobs.pop`, one fiber will process it.

> 
> P.S. the major issue with the original version is that we called `close` which would go for the instance variable `@redis_async` instead of the previously resolved `connect`, so we should have stored the return value of `connect` into a local variable and close it instead. But in case the aboe scenario works reasonably well, it is way simpler to let the underlying pool handle the situation. P.P.S. the second major issue is that to clear the instance variable, we first need to be sure that `@redis_async == local_connection_variable` so that we don't clear the new connection from another fiber.

Yeah this is much simpler, and what is better: not maintained by us.

mayorova (approved):
Looks reasonable 👍

lib/3scale/backend/storage_async/client.rb
akostadinov: To be sure, you're saying that lets say 10 fibers will fail and wait simultaneously so we will be like in total one wait cycle once connections start working, correct? Our connection pool is same as number of fibers or larger? Or smaller?

jlledom: > To be sure, you're saying that lets say 10 fibers will fail and wait simultaneously so we will be like in total one wait cycle once connections start working, correct?

If 10 fibers fail at the same time, there will be 10 retry cycles occurring in parallel. Once the redis servers is back, all connections will work again as soon as their wait period finishes. Since waits are in parallel, it shouldn't take much more time than one single wait period in real time. Is that what you mean?

> Our connection pool is same as number of fibers or larger? Or smaller?

All fibers share the same single connection pool. If you meant the number of actual connections, that's not managed by us: the connection pool will open as many connections as it needs, but we limit this number to `max_connections`

spec/unit/storage_async/client_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'spec_helper'
+
+if ThreeScale::Backend.configuration.redis.async
+  module ThreeScale
+    module Backend
+      module StorageAsync
+        describe Client do
```
akostadinov, line 7: ChatGPT thinks this can be used as a conditional skip. It just looks somehow ugly to have this huge `if` statement.

```suggestion
        describe Client, skip: !Backend.configuration.redis.async do
```

akostadinov (approved):

lib/3scale/backend/storage_async/client.rb
akostadinov: Ok, as we further spoke, a multiple wait can happen only in isolated edge cases where we have many connections in the pool already, then we have all of a sudden very low load (e.g. a single request), and then all connection die in a way that we don't receive a connection reset from the server. But there is no clear way to protect against such condition without compromising the handling of other, more common failure conditions (e.g. a single connection failed). Moreover this worst edge case scenario above, should be cause by a pretty severe other issues where it is unlikely that a little extended retry wait in apisonator will significantly affect. So I'm fine with the current version, let me know if I'm missing something. Thank you!

spec/unit/storage_async/client_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+              reconnect_attempts: 2,
+              reconnect_wait_seconds: 0,
+              max_connections: 10
+            }
+          end
+          let(:client) { described_class.new(opts) }
+          let(:fake_conn) { double('async_redis_client') }
+
+          before do
+            allow(AsyncRedis::Client).to receive(:connect).and_return(fake_conn)
```
akostadinov, line 20: One other thing, this `fake_conn` is evaluated once I think. So if we create multiple `AsyncRedis::Client` , they will all use the same object `fake_conn`. I think we need to create a new double every time, otherwise we don't know if new client was created or not.

spec/unit/storage_async/client_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+                allow(fake_conn).to receive(:call) do
+                  call_count += 1
+                  raise Errno::ECONNRESET if call_count == 1
+                  'PONG'
+                end
+
+                result = client.call('PING')
+
+                expect(result).to eq('PONG')
+                expect(AsyncRedis::Client).to have_received(:connect).once
```
akostadinov, line 36: wait, shouldn't we have called `PING` once, then it failed and then we called it second time and it succeeded. Shouldn't this mean we should have called connect twice?

spec/unit/storage_async/client_spec.rb
akostadinov: ignore, maybe that's fine but later the test is confusing me, added another comment there.

spec/unit/storage_async/client_spec.rb
akostadinov: Ok, so we call connect on `AsyncRedis::Client` once and then we just `#call` and it creates actual Redis connections as needed. All is fine.

spec/unit/storage_async/client_spec.rb
jlledom: This doesn't work because the async `Client` class is not loaded in sync mode, so the call to `describe` fails. It can be solved with a few changes but I don't think it's much worth it.

spec/unit/storage_async/client_spec.rb
jlledom: The `fake_conn` object is reset after each test, that's what `let` does.

spec/unit/storage_async/client_spec.rb
jlledom: Yeah, it's confusing because have two classes called `Client` which have a method called `:connect`

In this test:

- `StorageAsync::Client#connect`: This is called twice, each per attempt
- `AsyncRedis::Client#connect`: This is called once, no matter how many attempts

This test asserts `AsyncRedis::Client#connect` is called once, that means we are reusing the same instance instead of the multiple instances we were creating before. That's the bug we fix in this PR.

